PR #13, opened 2019-07-17
Add CORS section

### LINKED ISSUE #12: Write CORS section
Translate https://github.com/solid/solid-spec/blob/103b1e027356bd525e4cad0138e8288f4881df39/recommendations-server.md#cors---cross-origin-resource-sharing into an implementable text.

RubenVerborgh:
- [x] Write background
- [x] Explain which parts of the CORS spec should be implemented
- [x] Write security considerations.

Closes #12.

### DIFF
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -12,4 +12,5 @@ Issue: Explain that this specification is not documentation;
 
 ## Definitions ## {#definitions}
 
-Issue: Write Definitions section.
+A <dfn export>data pod</dfn> is a place for storing documents,
+with mechanisms for controlling who can access what.
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -14,3 +14,6 @@ Issue: Explain that this specification is not documentation;
 
 A <dfn export>data pod</dfn> is a place for storing documents,
 with mechanisms for controlling who can access what.
+
+A <dfn export>Solid app</dfn> is an application
+that reads or writes data from one or more [=data pods=].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -41,6 +41,52 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
 
 ## Cross-Origin Resource Sharing ## {#cors}
 
+### Background and Need ### {#cors-need}
+<em>This section is non-normative.</em>
+
+[=Solid apps=] typically access data from multiple sources.
+However,
+Web browsers by default prevent apps that run on one origin
+from accessing data on other origins.
+This cross-origin protection is a security mechanism
+that ensures malicious websites cannot simply read
+your profile or banking details from other websites.
+However, this reasonable default poses a problem
+even for benevolent Solid apps,
+which might have good reasons to access data from different places.
+For instance,
+a Solid app at `https://app.example/`
+would be prevented from accessing data on
+`https://alice-data-pod.example/` or `https://bob-data-pod.example/`,
+even when Alice and Bob have given the user of the app
+their permission to see some of their data.
+
+For cases where the other origins
+have their own access protection mechanism—
+[like within Solid](#wac)—
+the browser's built-in cross-origin protection
+is actually an obstacle rather than a feature.
+After all,
+[=data pods=] already ensure through access control
+that certain documents can only be accessed by specific people.
+Preventively blocking apps from different origins
+thus introduces an unnecessary barrier.
+
+Fortunately,
+Web servers can indicate to the browser
+that certain documents do not require cross-origin protection.
+This mechanism to selectively disable that protection
+is called *Cross-Origin Resource Sharing* or *CORS* [[FETCH]].
+By responding to browser requests
+with a specific combination of HTTP headers,
+servers can indicate which actions are allowed for a given resource.
+For a Solid data pod,
+the goal is to allow *all* actions on the CORS level,
+such that the deeper [access control layer](#wac)
+can exert full control over the app's allowed permissions.
+The next section describes how to achieve this
+through the right HTTP header configuration.
+
 Issue: Write Cross-Origin Resource Sharing section.
 
 A Solid data pod MUST conform to the CORS specification [[!FETCH]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -90,8 +90,8 @@ through the right HTTP header configuration.
 ### Required server-side implementation ### {#cors-server}
 
 A [=data pod=] MUST implement the CORS protocol [[!FETCH]]
-such that no request to it or response from it
-is blocked by the browser.
+such that, to the extent possible,
+no request to it or response from it is blocked by the browser.
 If the data pod wishes to block access to a resource,
 this MUST NOT happen via CORS
 but MUST instead be communicated to the Solid app in the browser
@@ -103,6 +103,32 @@ Note: Since the CORS protocol is part of a Living Standard,
 it should still be considered in evolution.
 Changes to the CORS protocol might thus happen at any point,
 which might necessitate changes to server implementations
-to prevent undesired blocking.
+for continued prevention of undesired blocking.
 A [proposal](https://github.com/whatwg/fetch/issues/878) to mitigate this
 has been suggested.
+
+Concretely,
+whenever a data pod receives an HTTP request
+containing an `Origin` header [[!RFC6454]],
+the server MUST respond with the appropriate `Access-Control-*` headers
+as specified in the CORS protocol [[!FETCH]].
+A data pod MUST also support the HTTP `OPTIONS` method [[!RFC7231]]
+such that it can respond appropriately to CORS preflight requests.
+
+Careful attention is warranted,
+especially because of the many edge cases.
+For instance,
+data pods SHOULD enumerate all response headers under `Access-Control-Expose-Headers`
+because there are cases (credentials mode set to `include`)
+where using `*` as a value does not work.
+As another example,
+data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
+This is recommended because,
+despite `Accept` being safelisted,
+values longer than 128 characters are blocked by default,
+and the number of available RDF formats that Solid apps might use
+could easily lead to longer value lengths.
+
+Note: Under the current CORS protocol,
+the correct HTTP response headers depend on the specifics of the request
+and thus cannot be pre-generated.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -134,9 +134,8 @@ through HTTP status codes such as
 `401`, `403`, or `404` [[!RFC7231]].
 
 Note: Since the CORS protocol is part of a Living Standard,
-it should still be considered in evolution.
-Changes to the CORS protocol might thus happen at any point,
-which might necessitate changes to server implementations
+it might be changed at any point,
+which might necessitate changes to data pod implementations
 for continued prevention of undesired blocking.
 A [proposal](https://github.com/whatwg/fetch/issues/878) to mitigate this
 has been suggested.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -87,6 +87,22 @@ can exert full control over the app's allowed permissions.
 The next section describes how to achieve this
 through the right HTTP header configuration.
 
-Issue: Write Cross-Origin Resource Sharing section.
+### Required server-side implementation ### {#cors-server}
 
-A Solid data pod MUST conform to the CORS specification [[!FETCH]].
+A [=data pod=] MUST implement the CORS protocol [[!FETCH]]
+such that no request to it or response from it
+is blocked by the browser.
+If the data pod wishes to block access to a resource,
+this MUST NOT happen via CORS
+but MUST instead be communicated to the Solid app in the browser
+through HTTP status codes such as `401`, `403`, or `404` [[!RFC7231]].
+Furthermore,
+all response headers MUST be exposed to the Solid app.
+
+Note: Since the CORS protocol is part of a Living Standard,
+it should still be considered in evolution.
+Changes to the CORS protocol might thus happen at any point,
+which might necessitate changes to server implementations
+to prevent undesired blocking.
+A [proposal](https://github.com/whatwg/fetch/issues/878) to mitigate this
+has been suggested.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -121,15 +121,16 @@ through the right HTTP header configuration.
 
 A [=data pod=] MUST implement the CORS protocol [[!FETCH]]
 such that, to the extent possible,
-the browser does not block
-any request sent to it or response received from it.
+the browser allows Solid apps
+to send any request and combination of request headers
+to the data pod,
+and the Solid app can read any response and response headers
+received from the data pod.
 If the data pod wishes to block access to a resource,
 this MUST NOT happen via CORS
 but MUST instead be communicated to the Solid app in the browser
 through HTTP status codes such as
 `401`, `403`, or `404` [[!RFC7231]].
-Furthermore,
-all response headers MUST be exposed to the Solid app.
 
 Note: Since the CORS protocol is part of a Living Standard,
 it should still be considered in evolution.
@@ -141,28 +142,28 @@ has been suggested.
 
 Concretely,
 whenever a data pod receives an HTTP request
-containing an `Origin` header [[!RFC6454]],
+containing a valid `Origin` header [[!RFC6454]],
 the server MUST respond with the appropriate `Access-Control-*` headers
 as specified in the CORS protocol [[!FETCH]].
+In particular,
+the data pod MUST set the `Access-Control-Allow-Origin` header
+to the valid `Origin` value from the request
+and list `Origin` in the `Vary` header value.
+The data pod MUST make all used response headers readable for the Solid app
+through `Access-Control-Expose-Headers`
+(with the possible exception of the `Access-Control-*` headers themselves).
 A data pod MUST also support the HTTP `OPTIONS` method [[!RFC7231]]
 such that it can respond appropriately to CORS preflight requests.
 
 Careful attention is warranted,
 especially because of the many edge cases.
 For instance,
-data pods SHOULD enumerate all response headers under `Access-Control-Expose-Headers`
-because there are cases (credentials mode set to `include`)
-where using `*` as a value does not work.
-As another example,
-data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
-This is recommended because,
-while CORS allows some headers such as `Accept` by default,
-they are blocked when their value exceeds 128 characters
-(which is not uncommon for RDF consumers such as Solid apps).
-
-Note:
-The correct HTTP response headers depend on the specifics of the request
-and thus cannot be pre-generated.
-For example,
-`Access-Control-Allow-Origin: *` cannot be used with a credentials mode of `include`
-and should instead reflect a valid `Origin` sent by the client.
+data pods SHOULD explicitly enumerate
+all used response headers under `Access-Control-Expose-Headers`
+rather than resorting to `*`,
+which does not cover all cases (such as credentials mode set to `include`).
+Data pods SHOULD also explicitly list `Accept` under `Access-Control-Allow-Headers`,
+because values longer than 128 characters
+(not uncommon for RDF-based Solid apps)
+would otherwise be blocked,
+despite shorter `Accept` headers being allowed without explicit mention.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -127,9 +127,7 @@ If the data pod wishes to block access to a resource,
 this MUST NOT happen via CORS
 but MUST instead be communicated to the Solid app in the browser
 through HTTP status codes such as
-`401 Unauthorized`,
-`403 Forbidden`,
-or `404 Not Found` [[!RFC7231]].
+`401`, `403`, or `404` [[!RFC7231]].
 Furthermore,
 all response headers MUST be exposed to the Solid app.
 
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -87,15 +87,20 @@ can exert full control over the app's allowed permissions.
 The next section describes how to achieve this
 through the right HTTP header configuration.
 
+
 ### Required server-side implementation ### {#cors-server}
 
 A [=data pod=] MUST implement the CORS protocol [[!FETCH]]
 such that, to the extent possible,
-no request to it or response from it is blocked by the browser.
+the browser does not block
+any request sent to it or response received from it.
 If the data pod wishes to block access to a resource,
 this MUST NOT happen via CORS
 but MUST instead be communicated to the Solid app in the browser
-through HTTP status codes such as `401`, `403`, or `404` [[!RFC7231]].
+through HTTP status codes such as
+`401 Unauthorized`,
+`403 Forbidden`,
+or `404 Not Found` [[!RFC7231]].
 Furthermore,
 all response headers MUST be exposed to the Solid app.
 
@@ -124,11 +129,13 @@ where using `*` as a value does not work.
 As another example,
 data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
 This is recommended because,
-despite `Accept` being safelisted,
-values longer than 128 characters are blocked by default,
-and the number of available RDF formats that Solid apps might use
-could easily lead to longer value lengths.
+while CORS allows some headers such as `Accept` by default,
+they are blocked when their value exceeds 128 characters
+(which is not uncommon for RDF consumers such as Solid apps).
 
-Note: Under the current CORS protocol,
-the correct HTTP response headers depend on the specifics of the request
+Note:
+The correct HTTP response headers depend on the specifics of the request
 and thus cannot be pre-generated.
+For example,
+`Access-Control-Allow-Origin: *` cannot be used with a credentials mode of `include`
+and should instead reflect a valid `Origin` sent by the client.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -97,7 +97,8 @@ the browser's built-in cross-origin protection
 is actually an obstacle rather than a feature.
 After all,
 [=data pods=] already ensure through access control
-that certain documents can only be accessed by specific people.
+that certain documents can only be accessed
+by specific people or applications.
 Preventively blocking apps from different origins
 thus introduces an unnecessary barrier.
 
--- a/main/security.bs
+++ b/main/security.bs
@@ -1,6 +1,12 @@
 Security Considerations {#security}
 ===================================
 
+Some of the normative references with this specification
+point to documents with a _Living Standard_ status,
+meaning their contents can still change over time.
+It is advised to monitor these documents,
+as such changes might have security implications.
+
 A data pod MUST NOT assume that
 HTTP request headers sent by a client are valid,
 and MUST reject or sanitize invalid header values
--- a/main/security.bs
+++ b/main/security.bs
@@ -3,28 +3,40 @@ Security Considerations {#security}
 
 Issue: Write Security Considerations section.
 
+A data pod MUST NOT assume that HTTP request headers
+sent by a client are valid.
+For example,
+values for `Host` and `Origin`
+MUST NOT be assumed to be free of possibly malicious sequences
+such as `/..` or others.
+
+A data pod MUST NOT assume that the user agent is a regular Web browser,
+even when requests contain familiar values
+in headers such as `User-Agent` or `Origin`.
+Such an assumption could incorrectly lead to conclusions
+about the security model of the application making the request,
+whereas the request might actually come
+from a malicious actor that is not bound to the same security constraints.
+
 Solid data pods [disable all cross-origin protections](#cors-server) in browsers
 because resource access is governed explicitly by [Web Access Control](#wac).
 As such,
 data pods MUST NOT rely on cross-origin protection
 for shielding access to resources.
 While this ensures that unauthorized resource access will not occur,
-additional security measures MAY be needed
+additional security measures MAY be taken
 to prevent metadata in error responses from leaking.
 For instance,
 a malicious app could probe multiple servers
-to check whether the response status code is `401` or `403`,
+to check whether the response status code is
+`401 Unauthorized` or `403 Forbidden`,
 or could try to access an error page
 from an intranet server within the user agent's private network
 to extract company names or other data.
 To mitigate this,
-when a request from an untrusted the `Origin` arrives,
-the data pod MAY set the status code of error responses to `404`
+when a request from an untrusted `Origin` arrives,
+the data pod MAY set the status code of error responses to `404 Not Found`
 and/or anonymize or censor their contents.
-
-A data pod MUST NOT assume that the user agent is a regular Web browser,
-even when requests contain familiar values
-in headers such as `User-Agent` or `Origin`.
 
 
 ## Privacy Considerations ## {#privacy}
--- a/main/security.bs
+++ b/main/security.bs
@@ -30,10 +30,11 @@ from a non-browser actor unaffected by browser security constraints.
 Solid data pods [disable all cross-origin protections](#cors-server) in browsers
 because resource access is governed explicitly by [Web Access Control](#wac).
 As such,
-data pods MUST NOT rely on cross-origin protection
-for shielding access to resources.
-While this ensures that unauthorized resource access will not occur,
-additional security measures MAY be taken
+data pods MUST NOT rely on browser-based cross-origin protection mechanisms
+for determining the authentication status or representation of a resource.
+In particular,
+they MUST ignore HTTP cookies from untrusted origins.
+Additional security measures MAY be taken
 to prevent metadata in error responses from leaking.
 For instance,
 a malicious app could probe multiple servers
--- a/main/security.bs
+++ b/main/security.bs
@@ -28,14 +28,13 @@ additional security measures MAY be taken
 to prevent metadata in error responses from leaking.
 For instance,
 a malicious app could probe multiple servers
-to check whether the response status code is
-`401 Unauthorized` or `403 Forbidden`,
+to check whether the response status code is `401` or `403`,
 or could try to access an error page
 from an intranet server within the user agent's private network
 to extract company names or other data.
 To mitigate this,
 when a request from an untrusted `Origin` arrives,
-the data pod MAY set the status code of error responses to `404 Not Found`
+the data pod MAY set the status code of error responses to `404`
 and/or anonymize or censor their contents.
 
 
--- a/main/security.bs
+++ b/main/security.bs
@@ -1,22 +1,25 @@
 Security Considerations {#security}
 ===================================
 
-Issue: Write Security Considerations section.
-
-A data pod MUST NOT assume that HTTP request headers
-sent by a client are valid.
+A data pod MUST NOT assume that
+HTTP request headers sent by a client are valid,
+and MUST reject or sanitize invalid header values
+before processing them
+or incorporating them in messages sent to others.
 For example,
 values for `Host` and `Origin`
 MUST NOT be assumed to be free of possibly malicious sequences
-such as `/..` or others.
+such as `/..` or others,
+and invalid `Origin` values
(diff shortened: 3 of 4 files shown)


dmitrizagidulin:
👍 (nicely done so far!)

main/security.bs
```diff
@@ -3,6 +3,33 @@ Security Considerations {#security}
...
+to prevent metadata in error responses from leaking.
+For instance,
+a malicious app could probe multiple servers
+to check whether the response status code is `401` or `403`,
+or could try to access an error page
+from an intranet server within the user agent's private network
+to extract company names or other data.
+To mitigate this,
+when a request from an untrusted the `Origin` arrives,
+the data pod MAY set the status code of error responses to `404`
```
dmitrizagidulin: This is well put, and really important. I'd love to revisit this topic, actually (in another PR).

RubenVerborgh: 👍 https://github.com/solid/specification/issues/14

dmitrizagidulin: Awesome!

acoburn:
👍 
I know this is really minor, but I especially appreciate the short line length in this PR. Short lines will mean that, when there are multiple PRs focused on similar or adjacent sections of the text, there will be considerably less rebasing required.

main/resource-access.bs
```diff
@@ -41,6 +41,94 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+
+### Required server-side implementation ### {#cors-server}
+
+A [=data pod=] MUST implement the CORS protocol [[!FETCH]]
+such that, to the extent possible,
+no request to it or response from it is blocked by the browser.
+If the data pod wishes to block access to a resource,
+this MUST NOT happen via CORS
+but MUST instead be communicated to the Solid app in the browser
+through HTTP status codes such as `401`, `403`, or `404` [[!RFC7231]].
```
pmcb55: Suggest (just for readability) changing:
`401`, `402`, or `403`
...to...
`401 Unauthorized`, `402 Payment Required`, or `403 Forbidden`

main/resource-access.bs
```diff
@@ -41,6 +41,94 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+As another example,
+data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
+This is recommended because,
+despite `Accept` being safelisted,
+values longer than 128 characters are blocked by default,
+and the number of available RDF formats that Solid apps might use
+could easily lead to longer value lengths.
+
+Note: Under the current CORS protocol,
+the correct HTTP response headers depend on the specifics of the request
```
pmcb55: An example here of a response header value depending on the specifics of the request would be great...?

main/resource-access.bs
```diff
@@ -41,6 +41,94 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+where using `*` as a value does not work.
+As another example,
+data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
+This is recommended because,
+despite `Accept` being safelisted,
+values longer than 128 characters are blocked by default,
+and the number of available RDF formats that Solid apps might use
+could easily lead to longer value lengths.
+
+Note: Under the current CORS protocol,
```
pmcb55: Generally, I dislike the use of 'currently' without also providing a time reference (e.g. '...currently (July 2019)...', or maybe in this case providing a specific version number of the CORS protocol (if there is one)...?

main/security.bs
```diff
@@ -3,6 +3,33 @@ Security Considerations {#security}
...
+and/or anonymize or censor their contents.
+
+A data pod MUST NOT assume that the user agent is a regular Web browser,
```
csarven: Perhaps it doesn't necessarily need to be mentioned in the document, but for now, can you clarify (in the comments) why a pod might make a decision based on the requesting agent? By default, it is preferable to not assume anything, so is this MUST NOT here to just to prevent any potential issues?

main/resource-access.bs
```diff
@@ -41,6 +41,94 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+Careful attention is warranted,
+especially because of the many edge cases.
+For instance,
+data pods SHOULD enumerate all response headers under `Access-Control-Expose-Headers`
+because there are cases (credentials mode set to `include`)
+where using `*` as a value does not work.
+As another example,
+data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
+This is recommended because,
+despite `Accept` being safelisted,
```
pmcb55: What do you mean by 'Accept being safelisted' here (as this is a SHOULD, and not a MUST)? Do you just mean 'This is recommended because, even if `Accept` is explicitly listed, values longer than 128 characters are _still_ blocked by default...' ?

main/resource-access.bs
```diff
@@ -41,6 +41,94 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+values longer than 128 characters are blocked by default,
+and the number of available RDF formats that Solid apps might use
+could easily lead to longer value lengths.
```
pmcb55: So just in summary, I don't see here how we get around this current CORS limitation of 128 characters for RDF formats for Solid apps (and especially so with profile values too). What can/should this spec say about this specific problem to allow Solid apps work reliably? E.g. should it say something like 'Due to this current (July 2019) 128 character limitation of CORS, Solid apps MUST ensure their Accepts headers remain within this limit'...?

main/security.bs
```diff
@@ -3,6 +3,33 @@ Security Considerations {#security}
...
+even when requests contain familiar values
+in headers such as `User-Agent` or `Origin`.
+
+
 ## Privacy Considerations ## {#privacy}
 
 Issue: Write Privacy Considerations section.
+
+In order to prevent leakage of non-resource data,
+error responses SHOULD NOT contain identifiable information.
```
csarven: When I'm reading something like this, I'd normally like to briefly know more, so would adding ", for example, .." make sense?

Perhaps preferably, it should refer to a reusable subsection on "Identifiable Information" under the "Privacy Consideration" section.

Also, why not a MUST NOT?

main/security.bs
```diff
@@ -3,6 +3,33 @@ Security Considerations {#security}
...
+additional security measures MAY be needed
+to prevent metadata in error responses from leaking.
+For instance,
+a malicious app could probe multiple servers
+to check whether the response status code is `401` or `403`,
+or could try to access an error page
+from an intranet server within the user agent's private network
+to extract company names or other data.
+To mitigate this,
+when a request from an untrusted the `Origin` arrives,
```
pmcb55: Remove 'the'.

main/security.bs
```diff
@@ -3,6 +3,33 @@ Security Considerations {#security}
...
+because resource access is governed explicitly by [Web Access Control](#wac).
+As such,
+data pods MUST NOT rely on cross-origin protection
+for shielding access to resources.
+While this ensures that unauthorized resource access will not occur,
+additional security measures MAY be needed
+to prevent metadata in error responses from leaking.
+For instance,
+a malicious app could probe multiple servers
+to check whether the response status code is `401` or `403`,
```
pmcb55: Suggest `401 Unauthorized` or `403 Forbidden`

main/security.bs
pmcb55: Suggest `404 Not Found`

main/resource-access.bs
RubenVerborgh: No versions sadly (and this is hugely problematic for the Web), only a date, which I will add.

main/security.bs
RubenVerborgh: It is for security reasons. In some discussions in the past, people were assuming that the presence of an `Origin` header meant that the client had the security model of a browser. Will try to describe that.

main/resource-access.bs
RubenVerborgh: Will expand. CORS-safelisted means that the `Accept` header is implicitly part of the allowed headers; but (and this [recently changed](https://github.com/whatwg/fetch/issues/862)) only if its value is not longer than 128 chars. So to always unblock, list `Accept` explicitly (because the 128 can change to a lower number at any point without notice).

main/resource-access.bs
```diff
@@ -41,6 +41,94 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+especially because of the many edge cases.
+For instance,
+data pods SHOULD enumerate all response headers under `Access-Control-Expose-Headers`
+because there are cases (credentials mode set to `include`)
+where using `*` as a value does not work.
+As another example,
+data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
+This is recommended because,
+despite `Accept` being safelisted,
+values longer than 128 characters are blocked by default,
```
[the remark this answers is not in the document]
RubenVerborgh: Yes, explicit CORS protocol limit.

main/security.bs
RubenVerborgh: > Perhaps preferably, it should refer to a reusable subsection on "Identifiable Information" under the "Privacy Considerations" section.

👍

> Also, why not a MUST NOT?

Because it might be hard to determine what identifiable information is (an extra space in the source code that only occurs with v37.466 which they run at place X), and to entirely get rid of it. And maybe people don't mind.

csarven: True. Perhaps once we determine if and what is normatively and/or non-normatively mentioned as identifiable information, it can clarify this point. The normative items need a test, and so MUST NOT works here. Otherwise, SHOULD NOT is fine as a catchall. I think?

RubenVerborgh:
Suggestions applied, thanks!

main/introduction.bs
```diff
@@ -12,4 +12,8 @@ Issue: Explain that this specification is not documentation;
 
 ## Definitions ## {#definitions}
 
-Issue: Write Definitions section.
+A <dfn export>data pod</dfn> is a place for storing documents,
```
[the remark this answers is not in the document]
RubenVerborgh: Agreed that such things might be good. Created an issue for tracking this: https://github.com/solid/specification/issues/16

I just wrote two quick definitions because I needed them for cross-referencing purposes. (The "data pod" term will now be linked up.)

main/introduction.bs
```diff
@@ -12,4 +12,8 @@ Issue: Explain that this specification is not documentation;
...
+with mechanisms for controlling who can access what.
+
+A <dfn export>Solid app</dfn> is an application
+that reads or writes data from one or more [=data pods=].
```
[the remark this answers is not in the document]
RubenVerborgh: Indeed; tracking that in https://github.com/solid/specification/issues/16.

RubenVerborgh:
Could we please have more reviews, @kjetilk/@michielbdejong/@csarven/@pmcb55/… 🙂 thanks in advance!

RubenVerborgh:
Could we please have more reviews, @kjetilk / @michielbdejong / @csarven / @pmcb55 / … 🙂 thanks in advance!

RubenVerborgh:
@kjetilk Definitions are a separate commit to this PR; but we have #16 to refine them. I just expect that #16 will take several weeks, and I needed _some_ definition of those two terms to link back to.

Are you okay with addressing the definitions in #16, or should I replace them with dummy placeholders for the purpose of this PR? (I need something to be there, or the document will not compile.)

kjetilk:
> @kjetilk Definitions are a separate commit to this PR; but we have #16 to refine them. I just expect that #16 will take several weeks, and I needed _some_ definition of those two terms to link back to.
> 
> Are you okay with addressing the definitions in #16, or should I replace them with dummy placeholders for the purpose of this PR? (I need something to be there, or the document will not compile.)

Ah, OK, then yes, I retract that comment, indeed, we can do that in #16

main/resource-access.bs
```diff
@@ -70,7 +70,101 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+
+A [=data pod=] MUST implement the CORS protocol [[!FETCH]]
+such that, to the extent possible,
+the browser does not block
+any request sent to it or response received from it.
+If the data pod wishes to block access to a resource,
+this MUST NOT happen via CORS
+but MUST instead be communicated to the Solid app in the browser
+through HTTP status codes such as
+`401 Unauthorized`,
```
kjetilk: Here, we are using "Unauthenticated". I don't know if we want to reflect that in the spec already.

main/resource-access.bs
```diff
@@ -70,7 +70,101 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+`403 Forbidden`,
```
kjetilk: Here, we distinguish between `403 User Unauthorized` and `403 Origin Unauthorized`.

main/resource-access.bs
RubenVerborgh: Grrrm, it's @pmcb55 who asked me to expand the status codes (used to be just `401`) with status texts, for which I am using the official IETF definitions. So the status texts are not relevant at all here, and just a clarification. We can go back to just numbers, but I would not go into different status texts here. (Orthogonality and all.)

main/resource-access.bs
RubenVerborgh: Idem; not the place. Could either leave as-is, or only mention the numeric code.

main/resource-access.bs
kjetilk: Yeah, that'd be a good compromise for now. We did elaborate on this extensively in NSS5 though, and I think it should be in the spec down the road.

RubenVerborgh: Sorry @pmcb55, status text have to go again!

main/security.bs
```diff
@@ -3,10 +3,48 @@ Security Considerations {#security}
...
+because resource access is governed explicitly by [Web Access Control](#wac).
+As such,
+data pods MUST NOT rely on cross-origin protection
+for shielding access to resources.
+While this ensures that unauthorized resource access will not occur,
+additional security measures MAY be taken
+to prevent metadata in error responses from leaking.
+For instance,
+a malicious app could probe multiple servers
+to check whether the response status code is `401` or `403`,
```
pmcb55: I don't see the update to reflect `401 Unauthorized` or `403 Forbidden`...?

RubenVerborgh: They had been added, but have now been reverted as per https://github.com/solid/specification/pull/13#discussion_r308779850

main/resource-access.bs
csarven: Good point @kjetilk . Agree with @RubenVerborgh on not getting into diffs. Things like 401 status text can be stated once (before use) and the remainder of the spec can stick to only using codes. Besides, the text surrounding the status code gives (or should) sufficient context.

main/resource-access.bs
```diff
@@ -70,7 +70,99 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+such that, to the extent possible,
+the browser does not block
+any request sent to it or response received from it.
+If the data pod wishes to block access to a resource,
+this MUST NOT happen via CORS
+but MUST instead be communicated to the Solid app in the browser
+through HTTP status codes such as
+`401`, `403`, or `404` [[!RFC7231]].
+Furthermore,
+all response headers MUST be exposed to the Solid app.
```
csarven: all with the exception of identifiable information and possibly something else?

main/resource-access.bs
```diff
@@ -70,7 +70,99 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+
+Note: Since the CORS protocol is part of a Living Standard,
+it should still be considered in evolution.
+Changes to the CORS protocol might thus happen at any point,
+which might necessitate changes to server implementations
+for continued prevention of undesired blocking.
+A [proposal](https://github.com/whatwg/fetch/issues/878) to mitigate this
+has been suggested.
+
```
csarven: This Note is perhaps not significant enough for this document.

main/resource-access.bs
```diff
@@ -70,7 +70,99 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+containing an `Origin` header [[!RFC6454]],
+the server MUST respond with the appropriate `Access-Control-*` headers
+as specified in the CORS protocol [[!FETCH]].
+A data pod MUST also support the HTTP `OPTIONS` method [[!RFC7231]]
+such that it can respond appropriately to CORS preflight requests.
+
+Careful attention is warranted,
+especially because of the many edge cases.
+For instance,
+data pods SHOULD enumerate all response headers under `Access-Control-Expose-Headers`
```
csarven: "Careful attention is warranted,
especially because of the many edge cases.
For instance," can be removed?

main/resource-access.bs
```diff
@@ -70,7 +70,99 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+because there are cases (credentials mode set to `include`)
+where using `*` as a value does not work.
+As another example,
```
csarven: "As another example," can be removed?

main/resource-access.bs
```diff
@@ -70,7 +70,99 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
+This is recommended because,
+while CORS allows some headers such as `Accept` by default,
+they are blocked when their value exceeds 128 characters
+(which is not uncommon for RDF consumers such as Solid apps).
+
```
csarven: Would switching the order of these sentences flow better?

main/resource-access.bs
RubenVerborgh: I don't think there are any exceptions. That is, at the moment identifiable information is transferred, the server has established the identity of the user.

If there were exceptions, then these could be bent by having a command-line client _pretend_ to be a browser (or forward a request on behalf of a browser), and that way stealing the information anyway.

The crucial notion is that we do _not_ rely on the browser's same-origin protection mechanisms for security, but rather on the authentication and authorization mechanism that Solid implements. So anything the server sends has been guaranteed to be safe by another mechanism.

Do you have concrete examples of headers? Or would you suggest to make this a SHOULD?

main/resource-access.bs
RubenVerborgh: It depends. We don't have to link to the issue as far as I'm concerned, but it seems crucial to me that people know that there is no such thing as opting out of same-origin protections, that servers will periodically need to be updated because CORS is a living spec.

The link to the issue was to indicate that, perhaps in the future, such periodical updates might not be necessary anymore.

I feel strongly about keeping in the first part, less strongly about the second part. What do you think?

main/resource-access.bs
RubenVerborgh: I believe @pmcb55 asked for examples at an earlier stage, actually. But this example is an important one, as it has actually bitten us before and we didn't understand. (`Accept` is whitelisted, why do we have to list it? Oh, the CORS spec changed and the whitelist is no longer a whitelist.)

main/resource-access.bs
RubenVerborgh: Hmm, I think it makes it less clear that they are two distinct cases.

Or do you want to remove the entire case? (Again something that has bitten us, hence why I am specifying it.)

main/resource-access.bs
RubenVerborgh: You mean the `include` and `Accept` examples? Can do. Awaiting your comment on the examples to proceed.

main/security.bs
```diff
@@ -3,10 +3,48 @@ Security Considerations {#security}
...
+from a malicious actor that is not bound to the same security constraints.
+
+Solid data pods [disable all cross-origin protections](#cors-server) in browsers
+because resource access is governed explicitly by [Web Access Control](#wac).
+As such,
+data pods MUST NOT rely on cross-origin protection
+for shielding access to resources.
+While this ensures that unauthorized resource access will not occur,
+additional security measures MAY be taken
+to prevent metadata in error responses from leaking.
```
csarven: I think this block would be a good lead for this section.

main/security.bs
```diff
@@ -3,10 +3,48 @@ Security Considerations {#security}
...
+For instance,
+a malicious app could probe multiple servers
+to check whether the response status code is `401` or `403`,
+or could try to access an error page
+from an intranet server within the user agent's private network
+to extract company names or other data.
+To mitigate this,
```
csarven: Can this be omitted?

RubenVerborgh: What specific part and for what reason?

main/resource-access.bs
csarven: I was making an editorial suggestion to remove the text prior to the requirement. "Data pods SHOULD.." stays.

main/resource-access.bs
csarven: No, again, editorial suggestion. "Data pods SHOULD.." stays.

main/resource-access.bs
RubenVerborgh: Mmm, then we lose the connection between the two sentences. Then they seem like independent advice.

main/resource-access.bs
RubenVerborgh: Same then; think it sacrifices clarity.

main/resource-access.bs
csarven: How about "While CORS allows [..] data pods SHOULD [..]"? Omit "This is recommended because,".

main/resource-access.bs
csarven: I'd like to understand how to qualify what's in the set of "all". What breaks if not "all"?

main/resource-access.bs
csarven: Good. Let's keep both for now.

main/security.bs
csarven: L29-35. I guess this is a bit stylistic. I felt that the recommendation following (L36-37) was clear and sufficient. I don't have a strong opinion about this.

main/resource-access.bs
RubenVerborgh: In general, HTTP headers convey useful or important information to the client. `Content-Type` for instance, but also more rare headers like `Link`, or even custom ones like `WAC-Allow`. What they all have in common is that, one way or another, the server wants to tell the client something about the enclosed representation.

And in regular Web apps, where the script requesting the representation is hosted on the same origin as the representation, the script can see all headers.

However, when origins are different—as is the case with Solid—access to several headers is blocked because, if the server relies on cross-origin protection as an authentication mechanism, such headers are sensitive.

Yet Solid servers do not rely on such an authentication mechanism.

Hence, we are requesting seeing all headers like a regular Web app would.

So what breaks if not all: in general, Solid apps would have fewer capabilities than regular Web apps (despite resources being authenticated).

main/security.bs
RubenVerborgh: Hmm, I think it's important that implementers see the security risk. Otherwise, it won't be obvious in what cases the MAY of 36–37 applies to them or why they might want to do that.

main/resource-access.bs
RubenVerborgh: But then we lose the connection with the previous sentence. I think we should consider the two sentences together:

> As another example, data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
This is recommended because,
while CORS allows some headers such as `Accept` by default,
they are blocked when their value exceeds 128 characters
(which is not uncommon for RDF consumers such as Solid apps).

I guess I could do:

> As another example, data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers` because
CORS only allows the `Accept` header by default
if its value remains below 128 characters.
However, longer `Accept` headers are common with RDF consumers such as Solid apps.

But not sure the connection is as clear then.

main/resource-access.bs
csarven: I understand that. I had a different interpretation of the text because I didn't catch the emphasis/context of "expose".  Perhaps we can lift that a bit by mentioning `ACAH`?

RubenVerborgh: 👌

main/resource-access.bs
```diff
@@ -70,7 +70,99 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+such that it can respond appropriately to CORS preflight requests.
+
+Careful attention is warranted,
+especially because of the many edge cases.
+For instance,
+data pods SHOULD enumerate all response headers under `Access-Control-Expose-Headers`
+because there are cases (credentials mode set to `include`)
+where using `*` as a value does not work.
+As another example,
+data pods SHOULD always explicitly list `Accept` under `Access-Control-Allow-Headers`.
```
csarven: L132 says:
>"all response headers MUST be exposed to the Solid app"

(pending https://github.com/solid/specification/pull/13/files#r309930384 )

Doesn't that actually cover the listing of `Accept` and `Access-Control-Allow-Headers`? Hence, is this recommendation at least redundant (possibly conflicting)?

main/resource-access.bs
csarven: The information in this Note is useful for Security Considerations. What do you think about moving it there?

main/resource-access.bs
RubenVerborgh: No. Exposing the _response_ headers happens through `Access-Control-Expose-Headers`. `Accept` is not a response header and should thus not be _exposed_; it should be _allowed_ in requests through `Access-Control-Allow-Headers`.

main/resource-access.bs
RubenVerborgh: It's not a security issue, rather the reverse. CORS will arbitrarily add _more_ security in the future, blocking functionality but not security.

csarven: The point was to say something along the lines of "CORS is a LS, so take care with anything pertaining to security" - we can't warn against unknown items at this time, so it is a general warning.

main/resource-access.bs
csarven: Oops. I mixed that up (01:20 CEST is definitely not an ideal time for reviews)

main/security.bs
RubenVerborgh: It is; the two preceding paragraphs are independent of CORS.

RubenVerborgh:
@csarven Thanks for the review, made several editorial changes in cd7af5b14aad16e5504bd3964a96f442d5995aa5 based on your feedback.

main/resource-access.bs
```diff
@@ -70,7 +70,100 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+their permission to see some of their data.
+
+For cases where the other origins
+have their own access protection mechanism—
+[like within Solid](#wac)—
+the browser's built-in cross-origin protection
+is actually an obstacle rather than a feature.
+After all,
+[=data pods=] already ensure through access control
+that certain documents can only be accessed by specific people.
```
justinwb: minor nit - consider making it "specific people or applications", since i could delegate authority for an autonomous application or bot to access resources as well.

csarven: Alternatively, "agents (people or software)"

main/resource-access.bs
```diff
@@ -70,7 +70,100 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+as specified in the CORS protocol [[!FETCH]].
+In particular,
+the data pod MUST set the `Access-Control-Allow-Origin` header
+to the valid `Origin` value from the request
+and list `Origin` in the `Vary` header value.
+The data pod MUST make all used response headers readable for the Solid app
+through `Access-Control-Expose-Headers`
+(with the possible exception of the `Access-Control-*` headers themselves).
+A data pod MUST also support the HTTP `OPTIONS` method [[!RFC7231]]
+such that it can respond appropriately to CORS preflight requests.
```
csarven: I don't mind this but is this need to be repeated in addition to the CORS spec implementation?

csarven:
@RubenVerborgh This is looking all around great so fine to merge (pending remaining minor edits). We'll surely revisit as we go.

There is one thing that we should consider whether to expand on or not (perhaps as a Note): While credentials and Origin shouldn't be relied upon, there are potholes that need to be navigated around. Should we insist on ACAO * and no credentials? And/or if ACAO echos the Origin, clients should request with Cache-Control no-store (minding the caveats that come with it)?

main/resource-access.bs
RubenVerborgh: > CORS is a LS, so take care with anything pertaining to security

So the "CORS is a LS" is in there.
"take care with anything pertaining to security" is, purposely, not: CORS is only going to get stricter with regard to security, not loser. So there are no security problems, and hence I wouldn't move this to the security section either. There are only "reverse" security problems: when the standard makes things more strict so you can't see things you should be able to.

main/resource-access.bs
RubenVerborgh: Made it "specific people or applications", but added note to https://github.com/solid/specification/issues/16 so we define "agents" as a term that can be reused without the parenthetical.

main/resource-access.bs
RubenVerborgh: It's a direct consequence of another comment of yours (https://github.com/solid/specification/pull/13#discussion_r309930384) where you asked to mention `Access-Control-Expose-Headers`.

So now there is:
- a first paragraph explaining it on a high level ("use CORS in this way")
- a second paragraph giving the most important bits of advice ("certainly use these")
- a third paragraph explaining the necessity for rigor ("when using those headers, pay attention to these")

Strictly speaking, only the first paragraph is needed. However, the third mentions many things that have bitten us about things explained in the second. And the second by itself is also useful to know where to start.

So I'd keep them in.

RubenVerborgh:
> Should we insist on ACAO * and no credentials?

We're already insisting on _not_ using `*`. Because right now, it breaks for credentials, but who knows what CORS will break in the future because of its Living Standard status. I don't think `*` is a good workaround for caching, a problem that has its causes elsewhere.

> And/or if ACAO echos the Origin, clients should request with Cache-Control no-store (minding the caveats that come with it)?

I know you've looked into this, but I would first need to see a reproducible example to make sure that this is indeed a fully functional workaround (and if it is, whether we cannot tweak it into a better solution).

> This is looking all around great so fine to merge

Thanks for the reviews, all. @csarven If you could top off with an approval, would be nice.

michielbdejong:
looks great!

> explicitly enumerate
all used response headers under `Access-Control-Expose-Headers`
rather than resorting to `*`,
which does not cover all cases

is there a link about that somewhere?

> (such as credentials mode set to `include`).

Is there a link somewhere explaining how `Access-Control-Expose-Headers` affects `.withCredentials`? I think that was the only part i didn't understand.

Also, I don't understand why an app would send credentials other than the `Authorization` header, because sending for instance cookies to a cross-origin API sounds like a pretty bad plan in general?

RubenVerborgh:
Thanks @michielbdejong!

> is there a link about that somewhere?

Yes; these are all part of the normative FETCH reference preceding them.

> Is there a link somewhere explaining how `Access-Control-Expose-Headers` affects `.withCredentials`?

That would also be the FETCH reference. I've checked it there, and it does not seem affected. (There is no `*` mode for that header.)

> Also, I don't understand why an app would send credentials other than the `Authorization` header, because sending for instance cookies to a cross-origin API sounds like a pretty bad plan in general?

On the one hand, this is a matter of spec orthogonality. Just in case there are cookies (for _anything_, not limited to auth), they should be supported. The CORS section does not mandate if and how authentication happens; could be with cookies, could be without. Also, they are just an example; the bottom line is that _no_ request/response should be blocked at the browser level.

On the other hand, WebID-OIDC needs cookies for a crucial step in the authentication process, so we do have a concrete case for it already.

main/resource-access.bs
csarven: It is risky for us to assume anything about a mutating spec's security implications.

A general warning about potential security implications of LSs would suffice.

kjetilk:
Four approvals, none against. Obviously, it needs more work, but I assume we're *go* on merging this?

main/resource-access.bs
RubenVerborgh: Good idea, added through 3912ad14aefac4a761160dcd74f612c40c97a1fd.

RubenVerborgh:
Thanks to @csarven, @dmitrizagidulin, @justinwb, @kjetilk, @michielbdejong, @pmcb55 for feedback!

> Four approvals, none against. Obviously, it needs more work, but I assume we're go on merging this?

Merging seems currently blocked by https://github.com/solid/process/pull/95; I will inquire about the status.

michielbdejong:
> in case there are cookies (for anything, not limited to auth), they should be supported

not sure I agree, because then i think they would be shared *across* apps if the user happens to have them open in the same browser session? For instance, if a UI like the data browser or a control panel on the pod's domain sets a cookie, then a third-party web app would do an XHR request and be allowed to do its request with that cookie from the on-pod UI?

RubenVerborgh:
> For instance, if a UI like the data browser or a control panel on the pod's domain sets a cookie, then a third-party web app would do an XHR request and be allowed to do its request with that cookie from the on-pod UI?

That is correct.

And our assertion in this document is that this MUST NOT be prevented _by the browser_. So we purposely do not rely not on browser security mechanisms.

However, the server is free to block cookies from third-party origins itself.

>  think in this case cookies would stick to the pod domain and not to the app domain, and thus could leak between apps.

Cookies will never "leak"; `withCredentials` is about whether or not cookies are _included_ with the request. This option is _not_ about who can see them; that is always only the pod server.

michielbdejong:
i see how it's consistent with the "don't rely on cors" narrative, and i also see how the hole may still be stopped at the next layer, but i still think it's at least unconventional and maybe even bad practice.

i think it would look like a bug to any security auditor if you allow third-party web apps to send http requests with the cookie of an on-pod UI.

could you at least add a note to implementers saying that allowing `.withCredentials` means that if the server uses cookies for some on-pod UI then the server code needs an extra if-statement in front of the session handling middleware that removes the cookie from the request based on some regex of the http headers etc

michielbdejong:
btw i do agree about the not-relying-on-browsers for the rest of it, just not for the cookies thing that's mentioned right at the end of the text.

RubenVerborgh:
> > we purposely do not rely not on browser security mechanisms.
> 
> I think that's at least unconventional and possibly also bad practice.

Unconventional, yes, but by design (and without it, no Solid in browsers).
Bad practice, no, because we put in place other authorization mechanisms.

> we should make sure we instruct server implementers about how to replace the security measures they are removing there.

Yes, agreed. Note that there already is a statement saying that data pods MUST NOT rely on cross-origin protections (such as those against cookies), but it is indeed better to say this explicitly.

> could you at least add a note to implementers saying that allowing .withCredentials means that if the server uses cookies for some on-pod UI then the server code needs an extra if-statement in front of the session handling middleware that removes the cookie from the request based on some regex of the http headers etc

Good point. Since that solution is perhaps too implementation-specific, I added that untrusted cookies MUST be ignored in 0f4d5dc9aa3086394370e7ba0ca79770c576a3bd. Thanks!

main/resource-access.bs
```diff
@@ -70,7 +70,101 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+a Solid app at `https://app.example/`
+would be prevented from accessing data on
+`https://alice-data-pod.example/` or `https://bob-data-pod.example/`,
+even when Alice and Bob have given the user of the app
+their permission to see some of their data.
+
+For cases where the other origins
+have their own access protection mechanism—
+[like within Solid](#wac)—
+the browser's built-in cross-origin protection
```
TallTed:
```
mechanism—
[like within Solid](#wac)—
the
```
->
```
mechanism &mdash;
[like within Solid](#wac) &mdash;
the
```

(If edited to one line, there should be a space both preceding and following the `&mdash;` entity.)

main/resource-access.bs
```diff
@@ -70,7 +70,101 @@ A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
...
+and the Solid app can read any response and response headers
+received from the data pod.
+If the data pod wishes to block access to a resource,
+this MUST NOT happen via CORS
+but MUST instead be communicated to the Solid app in the browser
+through HTTP status codes such as
+`401`, `403`, or `404` [[!RFC7231]].
+
+Note: Since the CORS protocol is part of a Living Standard,
+it should still be considered in evolution.
```
TallTed: `considered in evolution.`
-> `considered to be evolving.`

main/resource-access.bs
RubenVerborgh: m-dashes don't take surrounding spaces (The Chicago Manual of Style).

TallTed: Accessible style guide references on the web have changed since the last time I searched this out.  Apparently, AP recommends the surrounding spaces, while CMOS recommends against.  Lesser-known and private style guides go both ways.  Do we have a definitive statement somewhere that Solid docs will follow CMOS?

PR #17, opened 2019-07-23
Explain migration is still in progress

michielbdejong:


Mitzi-Laszlo:
Looks good to me!

RubenVerborgh:
👌 great idea.
Should also think about what chat channel to use in the future, perhaps
encourage people to file issues if they spot unclarities.

michielbdejong:
Thanks!

Issue #14, opened 2019-07-18
Discuss returning 404 for privacy reasons


RubenVerborgh:
For instance, the HTTP spec says:

> An origin server that wishes to "hide" the current existence of a
   forbidden target resource MAY instead respond with a status code of
   404 (Not Found).
—https://tools.ietf.org/html/rfc7231#section-6.5.3

csarven:
Some thoughts:

I like the 404 here.

Would it be useful to make sure that the 404 response MUST be accompanied with an expiration time for caches?

If so, related concern: if implementations only include freshness information for the purpose of hiding, then the evilApp could infer that there is a resource being hidden even while they can't access it. So, then MUST all 404s be always accompanied with an expiration time so as to not leak that information?

Note that if 404 with SHOULD or MAY for expiration time, it increases the uncertainty of a resource existing for the evilApp to infer.

humont:
Hi, new to this community but have some thoughts here:

My impression is that the active use case for 404 in Solid is to check whether a username is taken - which can be done at root level (and return a 404).

Any requests in deeper directories could return a 403 Forbidden if the ACL's don't allow public read. The 403 can also be made default regardless of whether or not the resource/path is valid.

This would prevent inference of resource existence, but avoids using 404 when a resource may in actual fact exist (obfuscation).

RubenVerborgh:
Hi @humont,

> The 403 can also be made default regardless of whether or not the resource/path is valid.

Good point, that does not seem to be disallowed by [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.5.3):

> > The client MAY repeat the request with new or different
   credentials.  However, a request might be forbidden for reasons
   unrelated to the credentials.

The RFC's suggestion to use a 404 is only a MAY:

> > An origin server that wishes to "hide" the current existence of a
   forbidden target resource MAY instead respond with a status code of
   404 (Not Found).

However, the question is _why_ we would prefer a 403 over a 404. You write:

> avoids using 404 when a resource may in actual fact exist (obfuscation).

But such obfuscation is explicitly allowed by the spec (see above); and a constant 403 is no less an obfuscation. So at the moment, I am still leaning toward 404 for privacy reasons, if the pod user so desires. On the other hand, a 403 makes it easier, since no such preference should be stated then, and it hints at authenticating (whereas a 404 does not).

> My impression is that the active use case for 404 in Solid is to check whether a username is taken

[...]

PR #26, opened 2019-08-06
Add HTTP section

RubenVerborgh:
Issues such as https://github.com/inrupt/wac-ldp/issues/114 indicate that we want to explicitly refer to the HTTP specification.

I am following the intent of the current draft specs and implementations, with the exception of:
- suggesting HTTP/2 (only in addition to HTTP/1.1)
- suggesting instead of requiring range requests (unclear whether https://github.com/solid/node-solid-server/issues/442 was a suggestion or requirement)

Feedback most welcome.

### DIFF
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -28,23 +28,42 @@ all `http` URIs MUST redirect to their `https` counterparts
 using a response with a `301` status code and a `Location` header.
 
 A data pod MUST implement the server part
-of HTTP/1.1 Conditional Requests [[!RFC7232]]
+of <cite>HTTP/1.1 Conditional Requests</cite> [[!RFC7232]]
 to ensure that updates requested by clients
 will only be applied if given preconditions are met.
 It SHOULD additionally implement the server part
-of HTTP/1.1 Caching [[!RFC7234]]
-to reduce bandwidth.
+of <cite>HTTP/1.1 Caching</cite> [[!RFC7234]]
+to reduce bandwidth consumption.
 A data pod MAY implement the server part
-of HTTP/1.1 Range Requests [[!RFC7233]]
+of <cite>HTTP/1.1 Range Requests</cite> [[!RFC7233]]
 on large representations
 to achieve further bandwidth reduction.
 
 A data pod MUST implement the server part
-of HTTP/1.1 Authentication [[!RFC7235]].
+of <cite>HTTP/1.1 Authentication</cite> [[!RFC7235]].
 When a client does not provide valid credentials
 when requesting a resource that requires it (see [[#webid]]),
 the data pod MUST send a response with a `401` status code
 (unless `404` is preferred for security reasons).
+
+### Required client-side implementation ### {#http-client}
+
+A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
+It MAY additionally be an HTTP/2 client [[!RFC7540]]
+to improve performance.
+
+A Solid client MAY implement the client parts of
+<cite>HTTP/1.1 Conditional Requests</cite> [[!RFC7232]]
+to only trigger updates when certain preconditions are met.
+It MAY implement
+<cite>HTTP/1.1 Caching</cite> [[!RFC7234]]
+and
+<cite>HTTP/1.1 Range Requests</cite> [[!RFC7233]]
+to reduce bandwidth consumption.
+
+A Solid client MUST implement the client part
+of <cite>HTTP/1.1 Authentication</cite> [[!RFC7235]]
+if it needs to access resources requiring authentication (see [[#webid]]).
 
 
 ## Linked Data Platform ## {#ldp}
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -5,28 +5,32 @@ Issue: Write introduction to the Authenticated Resource Access section.
 
 ## Hypertext Transfer Protocol ## {#http}
 
-Issue: Write Hypertext Transfer Protocol section.
+A [=data pod=] MUST be an HTTP/1.1 server [[!RFC7230]][[!RFC7231]].
+It SHOULD additionally implement HTTP/2 [[!RFC7540]]
+to improve performance,
+especially in cases where individual clients
+are expected to send high numbers of successive requests.
 
-Draft:
-A Solid data pod MUST implement HTTP/1.1 Message Syntax and Routing [[!RFC7230]]
+A data pod SHOULD use the `https` URI scheme
+in order to secure the connection between clients and servers.
+When both `http` and `https` are supported,
+all `http` URIs MUST redirect to their `https` counterparts
+using a response with a `301` status code and a `Location` header.
 
-Draft:
-A Solid data pod MUST implement HTTP/1.1 Semantics and Content [[!RFC7231]].
+A data pod MUST implement HTTP/1.1 Conditional Requests [[!RFC7232]]
+to ensure that updates requested by clients
+will only be applied if given preconditions are met.
+It SHOULD additionally implement HTTP/1.1 Caching [[!RFC7234]]
+to reduce bandwidth.
+A data pod MAY implement HTTP/1.1 Range Requests [[!RFC7233]]
+on large representations
+to achieve further bandwidth reduction.
 
-Draft:
-A Solid data pod MUST implement HTTP/1.1 Authentication [[!RFC7235]].
-
-Draft:
-A Solid data pod SHOULD implement HTTP/1.1 Conditional Requests [[!RFC7232]].
-
-Draft:
-A Solid data pod SHOULD implement HTTP/1.1 Range Requests [[!RFC7233]].
-
-Draft:
-A Solid data pod SHOULD implement HTTP/1.1 Caching [[!RFC7234]].
-
-Draft:
-A Solid data pod MAY implement HTTP/2 [[!RFC7540]].
+A data pod MUST implement HTTP/1.1 Authentication [[!RFC7235]].
+When a client does not provide valid credentials
+when requesting a resource that requires it (see [[#webid]]),
+the data pod MUST send a response with a `401` status code
+(unless `404` is preferred for security reasons).
 
 
 ## Linked Data Platform ## {#ldp}
@@ -40,7 +44,7 @@ A Solid data pod MUST conform to the LDP specification [[!LDP]].
 ## WebID ## {#webid}
 
 Issue: Explain inline that agents accessing non-public Solid resources
-  need a WebID, and that this is a URL
+  need to authenticate with a WebID, which is a URL
   pointing to a document with an RDF representation.
 
 
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -5,8 +5,10 @@ Issue: Write introduction to the Authenticated Resource Access section.
 
 ## Hypertext Transfer Protocol ## {#http}
 
+### Required server-side implementation ### {#http-server}
+
 A [=data pod=] MUST be an HTTP/1.1 server [[!RFC7230]][[!RFC7231]].
-It SHOULD additionally implement HTTP/2 [[!RFC7540]]
+It SHOULD additionally be an HTTP/2 server [[!RFC7540]]
 to improve performance,
 especially in cases where individual clients
 are expected to send high numbers of successive requests.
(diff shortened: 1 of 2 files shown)


main/resource-access.bs
```diff
@@ -5,28 +5,32 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
-A Solid data pod MAY implement HTTP/2 [[!RFC7540]].
+A [=data pod=] MUST be an HTTP/1.1 server [[!RFC7230]][[!RFC7231]].
+It SHOULD additionally implement HTTP/2 [[!RFC7540]]
+to improve performance,
+especially in cases where individual clients
+are expected to send high numbers of successive requests.
+
+A data pod SHOULD use the `https` URI scheme
+in order to secure the connection between clients and servers.
+When both `http` and `https` are supported,
```
dmitrizagidulin: I wonder if a SHOULD here is not strong enough? How do we feel about requiring `https` at least for key API endpoints?

RubenVerborgh: See here: https://github.com/solid/specification/pull/26/files#diff-8d72b2ee1631b8eec737030e9b4a3c9c

MUST would also mean development/testing mode, and behind a reverse proxy.

main/resource-access.bs
```diff
@@ -5,28 +5,32 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
+A data pod MUST implement HTTP/1.1 Conditional Requests [[!RFC7232]]
+to ensure that updates requested by clients
+will only be applied if given preconditions are met.
+It SHOULD additionally implement HTTP/1.1 Caching [[!RFC7234]]
+to reduce bandwidth.
+A data pod MAY implement HTTP/1.1 Range Requests [[!RFC7233]]
+on large representations
+to achieve further bandwidth reduction.
+
+A data pod MUST implement HTTP/1.1 Authentication [[!RFC7235]].
```
dmitrizagidulin: I wonder if it's worth being more specific here, re RFC7234 HTTP Authentication. Both in terms of - it's such a general purpose RFC, so it might make sense to specify which parts of it we care about. I'm also concerned that people might mis-interpret the mention of it as an endorsement of Basic or Digest auth schemes?

RubenVerborgh: My intention was to tackle these in the specific authentication sections (WebID-TLS and WebID-OIDC).

It's not RFC2617, so basic and digest are not included.

main/resource-access.bs
dmitrizagidulin: (I think you meant to include a line number in that link?)

I agree, making things easier for dev and testing is an important goal. But for what it's worth, the current implementations (node-solid-server) _require_ TLS (at least behind a self-signed cert). As in, they won't run any sort of WebID-related API endpoints without it.

main/resource-access.bs
RubenVerborgh: But we might want to not make proxy auth mandatory though.

main/resource-access.bs
RubenVerborgh: (It should point to all changes in main/security.bs)

> But for what it's worth, the current implementations (node-solid-server) _require_ TLS (at least behind a self-signed cert).

They do not actually; I am running NSS in HTTP mode behind a reverse proxy that terminates HTTPS.

main/resource-access.bs
dmitrizagidulin: Got it, makes sense.
I think my only hesitation, then, is just that, as an implementer reading that spec, it's not super clear as to what it would _mean_ for the pod to implement RFC7235? As in, which headers, what responses to expect, etc?

dmitrizagidulin:
(Big thumbs up for this PR in general, I think this is an important section.)

main/resource-access.bs
RubenVerborgh: Yes, I should probably state that it should conform to the server part of that spec (and for other normative HTTP refs too). This then comes down hete to understanding one response header and being able to send one request header and one status code.

RubenVerborgh: Clarified in ed5c3fbe98332ea9ab7b63569871c789108a7c57. The explicit mention of `401` leads to `WWW-Authenticate` being a MUST there.

main/resource-access.bs
```diff
@@ -5,28 +5,65 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
+
+A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
+It MAY additionally be an HTTP/2 client [[!RFC7540]]
+to improve performance.
+
+A Solid client MAY implement the client parts of
+<cite>HTTP/1.1 Conditional Requests</cite> [[!RFC7232]]
+to only trigger updates when certain preconditions are met.
+It MAY implement
+<cite>HTTP/1.1 Caching</cite> [[!RFC7234]]
```
kjetilk: BTW, I just started a [caching panel](https://github.com/solid/process/blob/master/panels.md#caching-panel). I think we will have to do a lot more than `MAY`s in this area, but the correct route is rather to implement the optional features ourselves, and then specify stronger requirements when we see what is needed.

RubenVerborgh: 👍 excellent. I'm definitely interested in that one as well.

main/resource-access.bs
```diff
@@ -5,28 +5,65 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
+
+### Required server-side implementation ### {#http-server}
+
+A [=data pod=] MUST be an HTTP/1.1 server [[!RFC7230]][[!RFC7231]].
+It SHOULD additionally be an HTTP/2 server [[!RFC7540]]
+to improve performance,
+especially in cases where individual clients
+are expected to send high numbers of successive requests.
+
+A data pod SHOULD use the `https` URI scheme
```
kjetilk: This formulation seems to be a departure from the ["it's not the 's'"](https://www.w3.org/DesignIssues/Security-NotTheS.html). I think it should emphasize what the s means, i.e. the use of TLS. I think the important part of the discussion is in the security section, which you have below.

main/resource-access.bs
```diff
@@ -5,28 +5,65 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
+in order to secure the connection between clients and servers.
+When both `http` and `https` are supported,
+all `http` URIs MUST redirect to their `https` counterparts
+using a response with a `301` status code and a `Location` header.
```
kjetilk: Have we agreed on that as an upgrade mechanism? Not e.g. [RFC2817](https://tools.ietf.org/html/rfc2817)

main/resource-access.bs
```diff
@@ -5,28 +5,65 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
+
+A data pod MUST implement the server part
+of <cite>HTTP/1.1 Conditional Requests</cite> [[!RFC7232]]
+to ensure that updates requested by clients
+will only be applied if given preconditions are met.
+It SHOULD additionally implement the server part
+of <cite>HTTP/1.1 Caching</cite> [[!RFC7234]]
+to reduce bandwidth consumption.
```
kjetilk: I think there is a long discussion about what caching can and cannot do for us. Reducing bandwidth consumption is just one and I think probably of relatively small importance. I don't know if we should add a list of things now, or defer that to a Caching panel, but I think I'd rather say "to help clients ensure that apps have sufficient responsiveness" or something like that. If it should be concise, it should be about the end goal.

main/resource-access.bs
RubenVerborgh: I think so, yes. The 301 is an important semantic signal too. Also there's HTTP/2. Do you see any reason for using RFC2817 instead?

main/resource-access.bs
RubenVerborgh: Hmm, bandwidth might be a goal in itself, especially on mobile devices.
I can turn it into "to improve performance"?

main/resource-access.bs
RubenVerborgh: Wording is from RFC7230; do a grep for TLS there, and the most prominent mentions are in "https URI scheme" sections.

Happy to additionally mention TLS, just didn't want to be redundant. Do you have a proposal for phrasing?

kjetilk: What I was trying to get at is that using the https by itself doesn't secure anything, the server admin could in principle have been duped into installing a middle-man's certificate. 

How about 

"A data pod SHOULD be secured with TLS, and therefore use the `https` URI scheme."

main/resource-access.bs
kjetilk: No, I'm actually not very knowledgeable about it, I only know that RFC2817 was made for exactly this case and IIRC @timbl has been talking about it.

Also, it doesn't seem very settled what the semantics of `301` in relation to RDF, even though it has been [discussed for a long time](https://blog.ldodds.com/2007/03/07/the-semantics-of-301-moved-permanently/). It was [raised in the TAG](https://www.w3.org/2007/awwsw/tracker/issues/1) but doesn't seem to have a conclusion.

I suppose *we* could say somewhere that `301` means `owl:sameAs` :-)

main/resource-access.bs
kjetilk: > Hmm, bandwidth might be a goal in itself, especially on mobile devices.

Yes, it can.

> I can turn it into "to improve performance"?

Yeah, that is probably better.

main/resource-access.bs
RubenVerborgh: OK, we can reconsider in the future if needed.

main/resource-access.bs
```diff
@@ -5,28 +5,65 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
-Draft:
-A Solid data pod SHOULD implement HTTP/1.1 Caching [[!RFC7234]].
-
-Draft:
-A Solid data pod MAY implement HTTP/2 [[!RFC7540]].
+### Background and Need ### {#http-need}
+<em>This section is non-normative.</em>
+
+Solid clients and servers need to exchange data securely over the Internet,
+and they do so using the Web protocol HTTP.
```
TallTed: `protocol HTTP.`
-> 'protocol, HTTP.`

RubenVerborgh: I intended it without actually. Like, "the Web protocol WebSockets", "the Web protocol CORS", "the Web protocol Memento". Not "_the_ Web protocol, namely HTTP".

TallTed: Ah. IME, what you meant would more commonly be presented (and understood) as "the WebSockets Web protocol", "the CORS Web protocol", etc.

main/resource-access.bs
csarven: >When both `http` and `https` are supported

If only `http` is supported, the rest doesn't apply.

Hence, at least other upgrade mechanisms like RFC2817 can still be used.

main/security.bs
```diff
@@ -3,6 +3,14 @@ Security Considerations {#security}
 
 Issue: Write Security Considerations section.
 
+Data pods SHOULD use TLS connections
+to protect the contents of requests and responses
+from eavesdropping and modification by third parties.
+Regular TCP connections MAY be used
```
justinwb: TLS connections are also regular TCP connections, so I think we should instead say something like "Unsecured connections that do not use TLS"

main/resource-access.bs
csarven: I find "the HTTP protocol" easier to read. "the HTTP Web protocol" is also okay to use in this case because of "Internet" mention, but anything after this need not mention "Web" - I think the context is obvious enough, and if need to distinguish somewhere for some reason, that can be done on a case by case basis.

RubenVerborgh: Indeed; the Internet/Web tuple was what I was going for.

main/resource-access.bs
```diff
@@ -5,28 +5,65 @@ Issue: Write introduction to the Authenticated Resource Access section.
...
+to only trigger updates when certain preconditions are met.
+It MAY implement
+<cite>HTTP/1.1 Caching</cite> [[!RFC7234]]
+and
+<cite>HTTP/1.1 Range Requests</cite> [[!RFC7233]]
+to improve performance.
+
+A Solid client MUST implement the client part
+of <cite>HTTP/1.1 Authentication</cite> [[!RFC7235]]
+if it needs to access resources requiring authentication (see [[#webid]]).
```
csarven: Server implementation says:
> "unless `404` is preferred for security reasons"

Acknowledging the open issue at https://github.com/solid/specification/issues/14 and that may change things here.. 

To be consistent in the mean time, the client should also say something about 404 . Perhaps along the lines of:

>The client MAY repeat the request with new or different credentials.

( https://github.com/solid/specification/issues/14#issuecomment-516808130 )

main/resource-access.bs
TallTed: (Of course, "the HTTP protocol", and even "the HTTP Web protocol", is redundant, as it's saying "the HyperText Transfer Protocol [Web] protocol"... unlike the other examples.)

RubenVerborgh: "the HTTP Web standard". You're welcome 😉

PR #89, opened 2019-10-04
Update LICENSE.md

TallTed:
* `- present` is not valid in a copyright statement.

* the year(s) should be followed by the name of the entity claiming the copyright.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright 2019 - present
+Copyright 2019 {replace with name of entity holding copyright}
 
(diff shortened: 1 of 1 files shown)


csarven:
Thanks Ted!

It is not clear in this PR but I've updated the LICENSE with your recommendations: https://github.com/solid/specification/commit/86862caa64df39431902705fe8730391c390c128

We will probably have to use a specific range (eg. 2019 - {year}) going forward as there will be revisions. I've specified "W3C Solid Community Group" as the copyright holder given that this work is carried out under that group.

csarven:
I'd like to note here that the draft or final report for the W3C Solid Community Group needs to eventually use the boilerplate text mentioned at https://www.w3.org/community/reports/reqs/ (under Copyright).

PR #113, opened 2019-11-06
Prep authentication spec skeleton.

dmitrizagidulin:
Sets up skeleton for the initial draft of the solid authentication spec.

### DIFF
--- a/webid-oidc/index.bs
+++ b/webid-oidc/index.bs
@@ -1,23 +1,46 @@
 <pre class='metadata'>
 Title: WebID-OIDC
+Boilerplate: issues-index no
 Shortname: webid-oidc
 Level: 1
 Status: w3c/ED
 Group: Solid Community Group
 URL: https://solid.github.io/specification/webid-oidc/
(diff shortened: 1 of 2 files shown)


webid-oidc/index.bs
```diff
@@ -1,23 +1,46 @@
...
-Editor: Your Name, Your Company http://example.com/your-company, your-email@example.com, http://example.com/your-personal-website
+Repository: https://github.com/solid/specification
+Markup Shorthands: markdown yes
+Max ToC Depth: 2
+Editor: Dmitri Zagidulin
+Editor: Justin Bingham
 Abstract:
-  This document introduces a mechanism for authenticating agents with a WebID
-  through the TLS protocol.
+  This document introduces a mechanism for user and application authentication
```
RubenVerborgh: Nitpick, but is it just people and apps? Clients, intermediaries, bots…?

dmitrizagidulin: Hehe, no, you're right, I figured we'd dive into that as it gets going. But lemme see if I can rephrase now..

dmitrizagidulin: Rephrased.

PR #114, opened 2019-11-06
Rename WebID-OIDC to Solid-OIDC.

dmitrizagidulin:
Renames authentication spec from WebID-OIDC to Solid-OIDC, to address https://github.com/solid/authentication-panel/issues/29.
(Note that this is a PR to another PR, #113.)

PR #140, opened 2020-01-20
clarify WebID-TLS

elf-pavlik:
During today's authentication call we discussed WebID-TLS with @dmitrizagidulin and @bblfish. We thought that mentioning it in non-normative section will suffice.
I also think that solid project doesn't need to take any responsibility for WebID-TLS spec itself since it doesn't seem to have any more a special role in solid ecosystem.

I keep in mind #114 but for now we can just use *WebID-OIDC* and rename everything once we resolve that naming issue.


csarven:
Can you share the implementation reports indicating where WebID-TLS "doesn't seem to have any more a special role in solid ecosystem"?

I would've expected some reasonable documentation, data, minutes etc backing up the case for downplaying WebID-TLS to the extent you propose. I don't see it in this repo or in the authn-panel repo. Perhaps I missed? This is not to imply that that is not a preferable direction to take in the spec, but considering where Solid is coming from - essentially why we had WebID-TLS to begin with - I think it'd be appropriate to capture the whys, give interested parties a chance to respond... before making the call.

bblfish:
Part of what the WebID group could produce is a report on what the problems with the protocol
are, which could be used to inform further work. It would probably be a good idea to first work this out.

One issue has been for example client side renegotiation in HTTP2.0.
But there seems to be progress on that front.
https://greenbytes.de/tech/webdav/draft-ietf-httpbis-http2-secondary-certs-05.html

elf-pavlik:
Please see:

[Removed references to TLS as primary auth method #171](https://github.com/solid/solid-spec/pull/171)
[Removed references to TLS that imply it is the primary form of authenication #26](https://github.com/solid/webid-oidc-spec/pull/26)

I think solid  spec might go with MAY and just reference https://www.w3.org/2005/Incubator/webid/spec/tls/
Main problem I can see with that comes from lack of active work on WebID-TLS. I don't know if anyone actively participating in solid would want to adopt it as work item. Given amount of work required to agree on one common mechanism combining authentication and authorization (at least part where user can delegate subset of one's own access to specific client application) I think we better focus on that and don't even try to take responsibility of addressing issues related to WebID-TLS.

If we decide to go with MAY I think we should document somewhere, maybe something like charter in process repo, who would want to take responsibility of resolving issues related to WebID-TLS and set clear scope  for those responsibilities.

Myself I rather focus on MUST mechanism and until we have this one stable don't divide effort on any of the possible MAYs.

elf-pavlik:
> Myself I rather focus on MUST mechanism and until we have this one stable don't divide effort on any of the possible MAYs.

I would like to ask if anyone plans to prioritize any WebID-TLS related work over all the other work needed. I really think that what this PR proposes can help us not to get distracted by WebID-TLS, by simply acknowledging its historical role in Solid and warning that one can't rely on broad support for WebID-TLS.

justinwb:
@elf-pavlik does this pull request have the support of the other members of the [authentication panel](https://github.com/solid/authentication-panel)? If so, and we could see endorsement from panel members, that would be helpful.

ewingson:
+1 from here, although having TLS-key as alternative login-method could go with MAY seems to be good...

justinwb:
@elf-pavlik and others - there seems to be general agreement on the substance of this pull from editors, authentication panel members, and prior threads. There's a large refactoring of this Solid ecosystem document in progress to extract the core solid client/server protocol into it's own document. Authentication modes will be included, and we anticipate that the substance of this pull will be covered when that is done. We can leave this pull open in the meantime, but it's likely that different text will be needed to address the substance of your submission in that new format.

csarven:
Reflected in https://github.com/solid/specification/commit/43a0cc739fb537988a602326b6fe4493a9f35d10 . Thanks!

elf-pavlik:
I still see webid-tls directory in the root.

csarven:
That wasn't particularly in use. Removed now. Nighty night.

PR #156, opened 2020-03-23
Initial auxiliary resources draft

justinwb:
Included in this pull is a [candidate proposal](https://github.com/solid/process#reviewing-proposals) from the [data interoperability panel](https://github.com/solid/data-interoperability-panel/tree/master/resource-metadata) for the handling of resource metadata in a Solid server. Specifically, this proposal introduces a mechanism for linking to metadata about resources in the Solid Ecosystem, and enumerates a set of reserved metadata types.

Remaining issues have been identified inline. Most noticeable will be the need to identify the contents of identified metadata types. The panel felt it was appropriate to raise those discussions to the editorial level first, as the various metadata types cross panel boundaries.

It should be noted that a significant amount of review and iteration was conducted over the course of several months, which was captured in solid/data-interoperability-panel#32.

### DIFF
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -159,6 +159,7 @@ to the Solid resource it is directly associated with, via HTTP `Link` headers.
 In these instances, the link relation `rel=describes` MUST be used.
 
 Issue: Is MUST too strong, as opposed to encouraging via SHOULD instead?
+[Related Issue](https://github.com/solid/data-interoperability-panel/issues/37)
 
 <div class="example">
   <p>
@@ -215,6 +216,7 @@ used, and may be added to the reserved set in the future.
 
 Issue: Agree on specific link relation URIs to use for shape, server managed,
 configuration.
+[Related Issue](https://github.com/solid/specification/issues/172)
 
 #### Web Access Control #### {#ar-wac}
 
@@ -224,7 +226,7 @@ type by Solid servers.
 The ACL auxiliary resource directly associated with a given resource is
 discovered by the client via the `rel="acl"` `Link` relation in a `Link` header.
 
-Issue: Considering moving some of this information to [[#wac]]
+Note: Consider moving some of this information to [[#wac]]
 
 A given Solid resource MUST NOT be directly associated with more than one ACL
 auxiliary resource. A given ACL auxiliary resource MUST NOT be directly
@@ -248,7 +250,7 @@ against authorization statements therein.
 
 #### Resource Description #### {#ar-description}
 
-Issue: Consider where there are any common parameters that would be
+Note: Consider where there are any common parameters that would be
 ubiquitous across resource descriptions that should be defined as part of the
 specification.
 
@@ -264,12 +266,14 @@ in a `Link` header.
 
 Issue: Consider whether a given Solid resource should be allowed to have
 multiple resource description auxiliary resources.
+[Related Issue](https://github.com/solid/specification/issues/173)
 
 A given Solid resource MUST NOT be directly associated with more than one
 Descriptive auxiliary resource.
 
 Issue: Determine what the default permissions should be on resource description
 auxiliary resources, or whether we should have them at all.
+[Related Issue](https://github.com/solid/specification/issues/174)
 
 To create or modify a Descriptive auxiliary resource, a given
 [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
@@ -303,10 +307,14 @@ is discovered by the client via the `rel=http://www.w3.org/ns/solid/terms#shape`
 a Shape validation auxiliary resource is discovered by the client via the
 `rel=describes` `Link` relation in a `Link` header.
 
-Issue: Consider moving some of this information to the Shape Validation section
+Note: Consider moving some of this information to the Shape Validation section
 
 A given Solid resource MUST NOT be directly associated with more than one
 Shape Validation auxiliary resource.
+
+Issue: Determine what the default permissions should be on shape validation
+auxiliary resources, or whether we should have them at all.
+[Related Issue](https://github.com/solid/specification/issues/174)
 
 To create or modify a Shape validation auxiliary resource, an
 [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
@@ -316,10 +324,7 @@ privileges per the
 [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
 on the resource directly associated with it.
 
-Issue: Determine what the default permissions should be on shape validation
-auxiliary resources, or whether we should have them at all.
-
-To read a Shape validation auxiliary resource, an
+To discover or read a Shape validation auxiliary resource, an
 [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
 MUST have
 [acl:Read](https://github.com/solid/web-access-control-spec#aclcontrol)
@@ -395,11 +400,20 @@ Configuration auxiliary resource.
 
 Issue: Determine what the default permissions should be on configuration
 auxiliary resources, or whether we should have them at all.
+[Related Issue](https://github.com/solid/specification/issues/174)
 
-To discover, read, create, or modify a Configuration auxiliary resource, an
+To create or modify a Configuration auxiliary resource, an
 [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
 MUST have
-[acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol)
+[acl:Write](https://github.com/solid/web-access-control-spec#aclcontrol)
+privileges per the
+[ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+on the resource directly associated with it.
+
+To discover or read a Configuration auxiliary resource, an
+[acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
+MUST have
+[acl:Read](https://github.com/solid/web-access-control-spec#aclcontrol)
 privileges per the
 [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
 on the resource directly associated with it.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -77,73 +77,105 @@ A Solid data pod MUST conform to the LDP specification [[!LDP]].
 
 ## Auxiliary Resources ## {#rm}
 
-### Background and Need ### {#rm-need}
+### Background and Need ### {#ar-need}
 <em>This section is non-normative.</em>
 
-This section introduces a mechanism for linking Auxiliary Resources with
-regular resources in the Solid Ecosystem using HTTP Link headers. The auxiliary
-resources may serve as
-supplementary descriptions or, when supported by a Solid server implementation,
-may influence the behavior of the resources. A given resource might link to zero
-or more such related auxiliary representations. It is not meant to replace the
-ability for the regular resource to self-describe.
+An auxiliary resource may provide supplementary information about a given
+Solid resource, or affect how that resource and others associated with it
+are processed, served, or interpreted. Different auxiliary resource types
+provide different capabilities. This section introduces a mechanism for linking
+auxiliary resources with regular Solid resources.
 
-Examples of this mechanism in use include:
+Auxiliary resources are needed to influence the configuration, processing, or
+interpretation of Solid resources without changing the composition of the
+resources themselves. To do so would be undesirable in many cases, and not
+possible in others. Auxiliary resources are not meant to replace the ability
+of a Solid resource to self-describe.
+
+Examples of auxiliary resources in use include:
 
 <ul>
-  <li>A binary JPEG image with a Link header, pointing to a description that has
-  an RDF representation.</li>
-  <li>A container with a Link header, pointing to an ACL resource that governs
-  access controls for the contained resources.</li>
-  <li>A resource whose shape is constrained by a ShEx schema includes a Link
-  header that points to an auxiliary resource defining that constraint.</li>
-  <li>A resource with an associated set of configurable parameters includes a
-  Link header that points to a resource where those configurable parameters
-  reside.</li>
+  <li>A binary JPEG image linked to an auxiliary resource that includes
+  information describing that binary JPEG.</li>
+  <li>A container linked to an auxiliary resource that includes access control
+  statements for that container and the resources that belong to it.</li>
+  <li>A resource representation whose shape is constrained by a given ShEx
+  schema that links to an auxiliary resource defining that schema.</li>
+  <li>A resource with an associated set of configurable parameters links to an
+  auxiliary resource where those configurable parameters reside.</li>
 </ul>
 
-NOTE: Provide a better explanation of why Auxiliary Resources are needed.
+### Required Server-side Implementation ### {#ar-server}
 
-### Auxiliary Resource Discovery ###  {#rm-discovery}
 
-To discover the auxiliary resources directly associated with a given target
-resource, a client MUST issue a `HEAD` or `GET` request to the target resource
-URL and inspect the `Link` headers in the response, or make an HTTP `GET`
-request on the target resource URL to retrieve an RDF representation, and
-inspect the encoded RDF graph for relations of one or more auxiliary types.
-These
-may be carried out in either order, but if the first fails to result in
-discovering the auxiliary resource or the described resource, the second MUST be
-tried.
-
-Issue: It would seem that requiring that auxiliary resources MUST looked up
-through *both*
-Link headers and the RDF graph is redundant and invites synchronization
-problems. Suggest that the primary mechanism is through Link headers, and the
-client MAY or SHOULD inspect the RDF graph for any additional auxiliary
-resources.
+For any defined auxiliary resource available for a given Solid resource, all
+representations of that resource MUST include an HTTP `Link` header pointing to
+the location of each auxiliary resource.
 
 The `rel={relation-type}` [[!RFC8288]] will define the relationship to the
-target URL in the `Link` header. Refer to [[!RFC8288]] for allowed [link
-relation types](https://httpwg.org/specs/rfc8288.html#link-relation-types).
+target URL in the HTTP `Link` header. URIs are encouraged to indicate Link
+relation types.
 
-For any defined auxiliary type available for a given resource, all
-representations of that resource MUST include a Link header pointing to the
-location of each auxiliary resource. An auxiliary resource discovered through a
-Link header
-for a given resource is considered to be *directly associated* with that
-resource. Access to different types of auxiliary resources require varying
-levels of
+An auxiliary resource linked with a given Solid resource through an HTTP `Link`
+header is considered to be *directly associated* with that resource. It is up
+to the server to determine how that association affects processing based on the
+auxiliary resource type.
+
+A given Solid resource MAY link to zero or more auxiliary resources. A given
+Solid resource MAY Link to auxiliary resources on a different server under a
+different authority, per the configuration of the Solid server on which that
+resource resides.
+
+Access to different types of auxiliary resources require varying levels of
 authorization, which MUST be specified as part of the definition for a given
-auxiliary type.
+auxiliary resource type.
 
-<p class="example">
-As defined by [[!WAC]]), a client can use this mechanism to discover the
-location of an ACL resource:<br/>
-`Link: <https://server.example/acls/resource.acl>; rel="acl"`
-</p>
+An auxiliary resource that resides on a Solid server MUST adhere to the same
+interaction model used by other regular Solid resources, except where specified
+in the definition of that auxiliary resource type.
 
-The following table lists [[#rm-reserved]] and the associated link relation URIs
+### Required Client-side Implementation ### {#ar-client}
+
+#### Discovery of Auxiliary Resources #### {#ar-discovery}
+
+To discover the auxiliary resources directly associated with a given Solid
+resource, a Solid client MUST issue a `HEAD` or `GET` request to the target
+resource URL and inspect the `Link` headers in the response.
+
+<div class="example">
+  <p>
+    As defined by [[!WAC]]), a client would discover the location of an auxiliary
+    ACL resource through the following HTTP `Link` header:
+  </p>
+  <div class="code">
+    `Link: <https://server.example/acls/resource.acl>; rel="acl"`
+  </div>
+</div>
+
+#### Discovery of Annotated Solid Resources #### {#ar-annotated}
+
+Certain auxiliary resource types MAY require the auxiliary resource to link back
+to the Solid resource it is directly associated with, via HTTP `Link` headers.
+In these instances, the link relation `rel=describes` MUST be used.
+
+Issue: Is MUST too strong, as opposed to encouraging via SHOULD instead?
+
+<div class="example">
+  <p>
+    An auxiliary resource `<https://server.example/auxiliary/resource.meta>` is
+    directly associated with `<https://server.example/resource.ttl>`. The auxiliary
+    type for `resource.meta` requires linking back to the annotated resource, which
+    is `resource.ttl`, so `resource.meta` must have the following included in its
+    `Link` headers:
+  </p>
+  <div class="code">
+    `Link: <https://server.example/resource.ttl>; rel="describes"`
+  </div>
+</div>
+
+### Reserved Auxiliary Resource Types ### {#ar-reserved}
+
+The following table lists [[#ar-reserved]] and the associated link relation URIs
 that are used for discovery. Other auxiliary types and relations may also be
 used, and may be added to the reserved set in the future.
 
@@ -152,102 +184,73 @@ used, and may be added to the reserved set in the future.
   <colgroup span="2"></colgroup>
   <thead>
     <tr>
-      <th>auxiliary Type</th>
+      <th>Auxiliary Type</th>
       <th>Link Relation</th>
     </tr>
   </thead>
   <tbody>
     <tr>
-      <td>[[#rm-wac]]</td>
+      <td>[[#ar-wac]]</td>
       <td>```"acl"``` or ```http://www.w3.org/ns/solid/terms#acl```</td>
     </tr>
     <tr>
-      <td>[[#rm-description]]</td>
+      <td>[[#ar-description]]</td>
       <td>```"describedby"``` or
 ```https://www.w3.org/ns/iana/link-relations/relation#describedby```</td>
     </tr>
     <tr>
-      <td>[[#rm-shape]]</td>
+      <td>[[#ar-shape]]</td>
       <td>```http://www.w3.org/ns/solid/terms#shape```</td>
     </tr>
     <tr>
-      <td>[[#rm-managed]]</td>
+      <td>[[#ar-managed]]</td>
       <td>```http://www.w3.org/ns/solid/terms#servermanaged```</td>
     </tr>
     <tr>
-      <td>[[#rm-configuration]]</td>
+      <td>[[#ar-configuration]]</td>
       <td>```http://www.w3.org/ns/solid/terms#configuration```</td>
     </tr>
   </tbody>
 </table>
 
-#### Discovery of Annotated Resource #### {#rm-annotated}
+Issue: Agree on specific link relation URIs to use for shape, server managed,
+configuration.
 
-Certain auxiliary resource types require the Solid server to link back to the
-annotated resource that the auxiliary resource is directly associated with, via
-Link
-headers. In these instances, the link relation ```rel=describes``` MUST be used.
+#### Web Access Control #### {#ar-wac}
 
-<p class="example">
-An auxiliary resource `<https://server.example/auxiliary/resource.meta>` is
-directly associated with `<https://server.example/resource.ttl>`. The auxiliary
-type for `resource.meta` requires linking back to the annotated resource, which
-is `resource.ttl`. `resource.meta` must have the following included in its
-`Link` headers:<br/>
-`Link: <https://server.example/resource.ttl>; rel="describes"`
-</p>
-
-### Auxiliary Resource Characteristics ### {#rm-characteristics}
-
-A given resource MAY Link to auxiliary resources on a different server under a
-different
-authority, per the configuration of the Solid server on which that resource
-resides.
-
-Auxiliary resources that reside on a Solid server MUST adhere to the same
-interaction model used by other standard Solid resources, except where specified
-in the definition of the associated auxiliary type in [[#rm-reserved]].
-
-An auxiliary resource MUST be deleted by the Solid server when the resource it
-is
-directly associated with is also deleted and the Solid server is authoritative
-for both resources.
-
-### Reserved Auxiliary Resource Types ### {#rm-reserved}
-
-#### Web Access Control #### {#rm-wac}
-
-ACL resources as defined by [[!WAC]] MUST be supported as an auxiliary
+ACL resources as defined by [[#wac]] MUST be supported as an auxiliary
 type by Solid servers.
 
 The ACL auxiliary resource directly associated with a given resource is
 discovered by the client via the `rel="acl"` `Link` relation in a `Link` header.
 
+Issue: Considering moving some of this information to [[#wac]]
+
 A given Solid resource MUST NOT be directly associated with more than one ACL
 auxiliary resource. A given ACL auxiliary resource MUST NOT be directly
-associated
-with more than one Solid resource.
+associated with more than one Solid resource.
 
 To discover, read, create, or modify an ACL auxiliary resource, an
 [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
 MUST have
 [acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol)
-privileges per the [ACL inheritance
-
-algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+privileges per the
+[ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
 on the resource directly associated with it.
 
-An ACL auxiliary resource MUST NOT be deleted unless the resource directly
-associated with it is deleted.
+An ACL auxiliary resource MUST be deleted by the Solid server when the resource
+it is directly associated with is also deleted and the Solid server is
+authoritative for both resources.
 
 A Solid server SHOULD sanity check ACL auxiliary resources upon creation or
 update to restrict invalid changes, such as by performing shape validation
 against authorization statements therein.
 
-#### Resource Description #### {#rm-description}
+#### Resource Description #### {#ar-description}
 
-Issue: Identify any auxiliary resource description parameters that MUST be
-recognized by all implementations.
+Issue: Consider where there are any common parameters that would be
+ubiquitous across resource descriptions that should be defined as part of the
+specification.
 
 Resource description is a general mechanism to provide descriptive metadata for
 a given resource. It MUST be supported as an auxiliary type by Solid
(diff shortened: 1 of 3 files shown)


csarven:
Can we make sure to link and track each criteria in the proposed document to a corresponding issue (or remarks, questions raised within) or vice-versa?

justinwb:
> Can we make sure to link and track each criteria in the proposed document to a corresponding issue (or remarks, questions raised within) or vice-versa?

@csarven are you talking about historical linkage, or creating new issues for any issues identified inline in the document? By citing solid/data-interoperability-panel#32 we capture just about all of the historic context. I can see reason already to make new issues for those raised inline in the submission which I can create today (unless I'm misinterpreting!)

csarven:
In order to manage the issues in the solid/specification repository, we should track all criteria in the proposed document to what's raised in this repository. See for example some of the commits in https://github.com/solid/specification/commits/editors-draft and how it references back to the issues and rough consensus that was reached in this repository. It helps to know the coverage and how to handle the open issues and their states in the project board going forward. Note that I'm not saying that all criteria in the proposed document has to have a corresponding issue in this repository - I know that not everything is covered here but was instead covered in the panel... and in that case, it should link to the issues / rough consensus that's documented in solid/data-interoperability-panel.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+#### Web Access Control #### {#rm-wac}
+
+ACL resources as defined by [[!WAC]] MUST be supported as a resource metadata
+type by Solid servers.
+
+The ACL metadata resource directly associated with a given resource is
+discovered by the client via the `rel="acl"` `Link` relation in a `Link` header.
+
+A given Solid resource MUST NOT be directly associated with more than one ACL
```
RubenVerborgh: Remove "directly" (inheritance)?

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+metadata resource. A given ACL metadata resource MUST NOT be directly associated
```
RubenVerborgh: inheritance?

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+with more than one Solid resource.
+
+To discover, read, create, or modify an ACL metadata resource, an
+[acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
+MUST have
+[acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol)
+privileges per the [ACL inheritance
+algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+on the resource directly associated with it.
```
RubenVerborgh: Plus mention that no other permissions affect the ACL resource.

Or maybe don't mention anything at all here, and just defer to ACL?

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+An ACL metadata resource MUST NOT be deleted unless the resource directly
+associated with it is deleted.
```
RubenVerborgh: Why not? Inheritance.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+A Solid server SHOULD sanity check ACL metadata resources upon creation or
+update to restrict invalid changes, such as by performing shape validation
+against authorization statements therein.
```
RubenVerborgh: Myeah, I think a lot of the above should move to the ACL document.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
 Draft:
 A Solid data pod MUST conform to the LDP specification [[!LDP]].
 
+## Resource Metadata ## {#rm}
```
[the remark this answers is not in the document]
RubenVerborgh: Also, reading the below, it seems this should really be `Metadata document`, given that we are only talking about metadata that resides in external documents, not metadata like content-type etc.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+associated with it is deleted.
+
+A Solid server SHOULD sanity check ACL metadata resources upon creation or
+update to restrict invalid changes, such as by performing shape validation
+against authorization statements therein.
+
+#### Resource Description #### {#rm-description}
+
+Issue: Identify any resource description metadata parameters that MUST be
+recognized by all implementations.
```
RubenVerborgh: client or server? (plus move to relevant parts)

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+servers.
+
+The Descriptive metadata resource directly associated with a given resource is
+discovered by the client via the `"rel=describedby"` `Link` relation in a `Link`
+header. Conversely, the resource being described by a Descriptive metadata
+resource is discovered by the client via the `rel="describes"` `Link` relation
+in a `Link` header.
+
+A given Solid resource MUST NOT be directly associated with more than one
+Descriptive metadata resource.
```
RubenVerborgh: ? why not

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+To create or modify a Descriptive metadata resource, a given
+[acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
+MUST have
+[acl:Write](https://github.com/solid/web-access-control-spec#aclcontrol)
+privileges per the [ACL inheritance
+algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+on the resource directly associated with it.
```
RubenVerborgh: Don't like; I think we should just let the normal ACL mechanism do its job and have it on the separate resource.

ACL should likely remain the only exception (because there's a very good reason).

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+[acl:Read](https://github.com/solid/web-access-control-spec#aclcontrol)
+privileges per the [ACL inheritance
+algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+on the resource directly associated with it.
+
+#### Shape Validation #### {#rm-shape}
+
+Issue: Detail the composition of a shape metadata resource, including the shape
+language, shape url, and any additional parameters to be used in shape
+validation by the server implementation.
```
RubenVerborgh: How would it be any different of regular resource validation, given that a) metadata resources are regular resources and b) regular resources also need validation?

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+Shape Validation ensures that any data changes in a given resource conform to an
+associated [[!SHACL]] or [[!SHEX]] data shape. It MUST be supported as a
+resource metadata type by Solid servers.
```
RubenVerborgh: Not sure why MUST

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+The Shape validation metadata resource directly associated with a given resource
+is discovered by the client via the `rel=http://www.w3.org/ns/solid/terms#shape`
+`Link` relation in a `Link` header. Conversely, the resource being described by
+a Shape validation metadata resource is discovered by the client via the
+`rel=describes` `Link` relation in a `Link` header.
+
+A given Solid resource MUST NOT be directly associated with more than one
+Descriptive metadata resource.
```
RubenVerborgh: duplicate?

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+To create or modify a Shape validation metadata resource, an
+[acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
+MUST have
+[acl:Write](https://github.com/solid/web-access-control-spec#aclcontrol)
+privileges per the [ACL inheritance
+algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+on the resource directly associated with it.
```
RubenVerborgh: Idem; let's not do this

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+To read a Shape validation metadata resource, an
+[acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
+MUST have
+[acl:Read](https://github.com/solid/web-access-control-spec#aclcontrol)
+privileges per the [ACL inheritance
+algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+on the resource directly associated with it.
+
+A Solid server SHOULD sanity check Shape validation metadata resources upon
+creation or update to restrict invalid changes.
```
RubenVerborgh: So shape validate the shape?

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+Managed metadata resource.
+
+To read a Server Managed metadata resource, an
+[acl:agent](https://github.com/solid/web-access-control-spec#describing-agents)
+MUST have [acl:Read](https://github.com/solid/web-access-control-spec#aclread)
+privileges per the [ACL inheritance
+algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm)
+on the resource directly associated with it. Modes of access beyond
+[acl:Read](https://github.com/solid/web-access-control-spec#aclread) MUST NOT be
+permitted on a Server Managed metadata resource.
```
RubenVerborgh: Idem

main/resource-access.bs
```diff
@@ -201,7 +486,8 @@ data pods SHOULD explicitly enumerate
 all used response headers under `Access-Control-Expose-Headers`
 rather than resorting to `*`,
 which does not cover all cases (such as credentials mode set to `include`).
-Data pods SHOULD also explicitly list `Accept` under `Access-Control-Allow-Headers`,
+Data pods SHOULD also explicitly list `Accept` under
+`Access-Control-Allow-Headers`,
```
RubenVerborgh: unrelated change

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+### Metadata Discovery ###  {#rm-discovery}
+
+To discover the metadata resources directly associated with a given target
+resource, a client MUST issue a `HEAD` or `GET` request to the target resource
+URL and inspect the `Link` headers in the response, or make an HTTP `GET`
+request on the target resource URL to retrieve an RDF representation, and
+inspect the encoded RDF graph for relations of one or more metadata types. These
+may be carried out in either order, but if the first fails to result in
+discovering the metadata resource or the described resource, the second MUST be
```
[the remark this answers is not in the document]
csarven: I've explained MUST, objective, cases with certain kinds... at https://github.com/solid/data-interoperability-panel/pull/32#issuecomment-599977210 . A client might not care about what exactly? If they're engaged in discovery, they certainly do care to follow through and check possible places. Of course they can halt the process at any point (as explained in panel PR comment) but the point is about capability.

"the" metadata would be a reference to reserved metadata types (as outlined re acl, shape, configuration, server managed, describedby).

Client "fails" to discover simply by not finding the desired relation. There is no error code for this kind of discovery, if that's what you're looking for.

Here the minimal expectation between the Link header relations and the RDF body relations would be that they shouldn't conflict eg. don't use different target URIs. A server can only control the headers but probably should stay out of the RDF body that's controlled by a client. Naturally that introduces different complications. See below.

---

Going forward:

I've explained the background on my proposal for documentation and then retracted the proposal in https://github.com/solid/data-interoperability-panel/issues/37#issuecomment-603110090 in response to @ericprud 's points. The limitations or perhaps the expectations it sets on server needs more attention in my opinion. If only the server can set the Link headers for metadata resource discovery, we need to be clear on the conditions in which each relation (per metadata type) is revealed by a server.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
 Draft:
 A Solid data pod MUST conform to the LDP specification [[!LDP]].
 
+## Resource Metadata ## {#rm}
+
+### Background and Need ### {#rm-need}
+<em>This section is non-normative.</em>
+
+Resource Metadata introduces a mechanism for linking to metadata about resources
```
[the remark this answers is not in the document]
justinwb: Updated in https://github.com/solid/specification/pull/156/commits/d85703f7c85fc6000ccb7592dd241cb3c74a4cfd

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+in the Solid Ecosystem using HTTP Link headers. The metadata may serve as
```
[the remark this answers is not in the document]
justinwb: We may need to better clarify the scope of this section, possibly starting with the title (as you rightly point out that metadata is a broad term). The aim is not to propose a replacement for the ability of the resource to self-describe (we make this point later in the paragraph). The aim is to provide a mechanism to associate resource(s) that are related to the resource in question, but not part of it. For example, access control is closely related to a resource, but the access control rules shouldn't be included as part of the resource itself. This does call into question whether we require additional sections to cover other parts of the metadata spectrum, or if it's good enough to simply be more specific in our scope?

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+ability for the directly associated resource to self-describe.
+
+Examples of this mechanism in use include:
+
+<ul>
+  <li>A binary JPEG image with a Link header, pointing to a description that has
+  an RDF representation.</li>
+  <li>A container with a Link header, pointing to an ACL resource that governs
+  access controls for the contained resources.</li>
+  <li>A resource whose shape is constrained by a ShEx schema includes a Link
```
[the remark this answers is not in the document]
justinwb: Our focus in this section was on shape association as a means to enforce of data constraints (i.e. the data in this resource has been validated against this shape). In that regard it is more like an access control rule, which would apply regardless of the representation. You raise a really good point that if the associated shape is used to inform the representation (i.e. https://profilenegotiation.github.io/I-D-Profile-Negotiation/I-D-Profile-Negotiation.html), it's a bit of a different story. Perhaps this is a good time to start discussing how associated shapes can be used in Profile-based Content Negotiation?

main/resource-access.bs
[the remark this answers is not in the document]
ericprud:
```suggestion
## Resource Metadata ## {#rm}
For the purposes of this document, consider three kinds of metadata; 1. metadata that has a significance in Solid (affects the processing of Solid resources), 2. metadata that specific servers maintain to reliably serve HTTP, wayback, or logging, or other services provided by the server, and 3. general metadata like titles of images or licenses of code. This document defines how clients can access 1 and 2.
```

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+A given resource MAY Link to metadata on a different server under a different
+authority, per the configuration of the Solid server on which that resource
+resides.
+
+Metadata resources that reside on a Solid server MUST adhere to the same
+interaction model used by other standard Solid resources, except where specified
+in the definition of the associated metadata type in [[#rm-reserved]].
+
+A metadata resource MUST be deleted by the Solid server when the resource it is
```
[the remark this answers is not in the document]
ericprud: IMO, Solid will be a lot simpler if ACLs are 1:1 onto resources. For example, grepping through the filesystem for something which happens to have a `acl:accessTo  <X>` will be problematic and invite synchronization difficulties. If we assume that, then it doesn't make sense to have an ACL with no resource.

We can explore shared ACLs if we have time and cycles, but I've never seen it used in any other system so I doubt there's a value which compensates for the complexity. (Also, it seems easy to add later on.)

main/resource-access.bs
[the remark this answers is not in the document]
justinwb: > Also, reading the below, it seems this should really be `Metadata document`, given that we are only talking about metadata that resides in external documents, not metadata like content-type etc.

I agree we should probably agree on what the right title for the section should be - though `Metadata Document` doesn't resonate for me. I think metadata like content-type would make sense to store in the server managed metadata type, so I think we still have coverage for that (though underspecified currently). Note that there's an inline issue to determine what should be included/required in the server-managed resources.

> This document defines how clients can access 1 and 2.

Generally happy with this suggested text - though I think that we also provide pathways for #3, by identifying the Resource Description type, which is meant for storage of that kind of associated metadata.

main/resource-access.bs
ericprud: My mental model is that there is exactly one ACL resource (or handle, if you will) that is allocated to a give resource. If it doesn't exist, the inheritance kicks in. So "associated with" means there is a dedicated URL to fulfill the role of ACL.

main/resource-access.bs
ericprud: Sure, but that's a tooling feature (copy ACLs from X)

main/resource-access.bs
ericprud: [ShExR.shex](https://github.com/shexSpec/shexTest/blob/master/doc/ShExR.shex) is pretty fussy. [SHACL.shacl](https://www.w3.org/TR/shacl/#shacl-shacl) had issues around performance but I think they backed off on the expressivity enough to make it reasonable.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+
+The `rel={relation-type}` [[!RFC8288]] will define the relationship to the
+target URL in the `Link` header. Refer to [[!RFC8288]] for allowed [link
+relation types](https://httpwg.org/specs/rfc8288.html#link-relation-types).
+
+For any defined metadata type available for a given resource, all
+representations of that resource MUST include a Link header pointing to the
+location of each metadata resource. Metadata discovered through a Link header
+for a given resource is considered to be *directly associated* with that
+resource. Access to different types of metadata require varying levels of
```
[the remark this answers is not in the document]
justinwb: ACL Inheritance is outside of the context of direct association. The scope of direct association is to say that a given metadata resource has a specifically defined association with a target resource. It is then up to the server to determine how it processes that association based on the metadata type. ACL inheritance is a perfect case of that.

For example: 

- Resource `C` is a container
- Resource `R` is a document that is a member of the Resource `C` container
- Metadata Resource `M` is an ACL resource that is directly associated with Resource `C`. Specifically, if you looked at the Link Headers for Resource `C`, you would see a rel=acl association for Resource `M`
- Resource `R` doesn't have an ACL resource that has been directly associated with it, which is what causes the server to inherit the ACL rules from Metadata Resource `M`. 
- The fact that this inheritance happens doesn't change the fact that Resource `C` is directly associated with Metadata Resource `M`, and Resource `R` is not. It is the specific absence of a directly associated ACL metadata resource for Resource `R` that triggers the inheritance.

Overall this approach has proven to be quite simple and we haven't found good reasons to make it any more complicated.

main/resource-access.bs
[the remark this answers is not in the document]
justinwb: Agree with @RubenVerborgh's [suggestion](https://github.com/solid/specification/pull/156/files/4c4dde9fe6352dd5802fd4b4b4d0fc98c7de7657#r396715703) that we make the exposure / discovery something that is specific to metadata types. I think we adjust this introductory text to enumerate the different ways of exposure and discovery (link headers, rdf graph), and then get specific for each type. Following that, we relax the MUST so that a client isn't required to look for every possible type of metadata in each path, each time.

> If only the server can set the Link headers for metadata resource discovery, we need to be clear on the conditions in which each relation (per metadata type) is revealed by a server.

@csarven The description of the different types already is doing a fair bit of this, but please take a look at comment anywhere you feel is lacking.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+tried.
+
+Issue: It would seem that requiring that metadata MUST looked up through *both*
+Link headers and the RDF graph is redundant and invites synchronization
+problems. Suggest that the primary mechanism is through Link headers, and the
+client MAY or SHOULD inspect the RDF graph for any additional metadata.
+
+The `rel={relation-type}` [[!RFC8288]] will define the relationship to the
+target URL in the `Link` header. Refer to [[!RFC8288]] for allowed [link
+relation types](https://httpwg.org/specs/rfc8288.html#link-relation-types).
```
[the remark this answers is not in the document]
justinwb: Agree - though we have the issue that IANA registered link relations currently don't have official URI mappings, which @elf-pavlik raised in solid/data-interoperability-panel#38. IMO - we decide on what URIs we want to use to represent the IANA registered string relations, and then consider marking the use of strings as not recommended.

main/resource-access.bs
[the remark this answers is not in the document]
justinwb: This is pretty straightforward when you you're dealing with directly associated metadata and things like inheritance and handled outside of it (which is our intention here). When metadata is 1:1 with a given resource, it should be expected that deleting the resource would also result in cleanup of metadata directly associated with it. 

FWIW - We had also asked @timbl's guidance on this, and his take was that the metadata specific to that resource (i.e. directly associated) should be removed at the same time.

main/resource-access.bs
justinwb: @ericprud has it right - i also explained in [this comment](https://github.com/solid/specification/pull/156/files/4c4dde9fe6352dd5802fd4b4b4d0fc98c7de7657#r397562513). Note that we do explicitly say in the individual metadata type sections that the relationships are 1:1. We could consider moving that up to the general section so its all stated togther.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+request on the target resource URL to retrieve an RDF representation, and
+inspect the encoded RDF graph for relations of one or more metadata types. These
+may be carried out in either order, but if the first fails to result in
+discovering the metadata resource or the described resource, the second MUST be
+tried.
+
+Issue: It would seem that requiring that metadata MUST looked up through *both*
+Link headers and the RDF graph is redundant and invites synchronization
+problems. Suggest that the primary mechanism is through Link headers, and the
+client MAY or SHOULD inspect the RDF graph for any additional metadata.
```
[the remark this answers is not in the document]
csarven: If I'm not mistaken, that Issue text is new; was added after the PR that got merged in data-interoperability-panel and before the PR here.

I've expressed implementation's capability at https://github.com/solid/data-interoperability-panel/pull/32#issuecomment-599977210 so the checks are not redundant without a reason. A client can stop after it checks one location (Link header or RDF body). There are cases where it may be necessary to check both, and that's obviously because it didn't discover the relation from the first check. There are also cases where only one needs to be checked and that depends on the target resource that the client is starting its discovery process from.

I feel that the background on why this was proposed is being overlooked: https://github.com/solid/data-interoperability-panel/issues/37#issuecomment-603110090 . I suppose we can come back to that...

If everything is left up to the server, it needs to advertise all of the "metadata" relations for each resource up front (and whether their targets have any content or not, but later expected to be updated by the client) at server-determined locations... because clearly the client has no way of adjusting that.

If the intention is to only link 1) Web documents and 2) have the serve control the relation, the Link header with rel is the only thing that's required for interop. There is no point in talking about discovery through RDF body unless there are some use cases that calls for it eg. requiring URIs with fragments, client controlling it. And in those cases, of course the Link header will not suffice; neither the relation can be expressed or the client have the means to express it.

The potential synchronisation issue is clear only when linking Web documents. If RDF body relations are allowed and subjects use URI fragments, synchronisation issue doesn't come up.

main/resource-access.bs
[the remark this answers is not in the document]
RubenVerborgh: 👍 Yes to the intent of the text, but phrasing is too complex.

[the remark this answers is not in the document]
justinwb: After some discussion (thanks @RubenVerborgh, @ericprud, @csarven) - we've decided that the name of this topic should more closely reflect what's being done, which is associated resources of various types with a given target resource. Have gone ahead and renamed from "Resource Metadata" to "Auxiliary Resources".

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+## Resource Metadata ## {#rm}
+
+### Background and Need ### {#rm-need}
+<em>This section is non-normative.</em>
+
+Resource Metadata introduces a mechanism for linking to metadata about resources
+in the Solid Ecosystem using HTTP Link headers. The metadata may serve as
+supplementary descriptions or, when supported by a Solid server implementation,
+may influence the behavior of the resources. A given resource might link to zero
+or more such related metadata representations. It is not meant to supplant the
```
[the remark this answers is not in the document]
justinwb: Changed to replace

main/resource-access.bs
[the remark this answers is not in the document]
csarven: Auxiliary resources such as access control, configuration and server-managed strictly operate on the level of HTTP resources, whereas others such as shapes and descriptions can alternatively work at the level of RDF graphs. For the former, only Link rel is required, and the latter can alternatively make use of the RDF body.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+      <td>```"acl"``` or ```http://www.w3.org/ns/solid/terms#acl```</td>
+    </tr>
+    <tr>
+      <td>[[#rm-description]]</td>
+      <td>```"describedby"``` or
+```https://www.w3.org/ns/iana/link-relations/relation#describedby```</td>
+    </tr>
+    <tr>
+      <td>[[#rm-shape]]</td>
+      <td>```http://www.w3.org/ns/solid/terms#shape```</td>
```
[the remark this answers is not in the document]
justinwb: Did you have that in mind for only shape (I'm guessing via https://www.w3.org/TR/vocab-dcat-2/#Class:Distribution), or for others as well (like server managed, configuration, description)?

main/resource-access.bs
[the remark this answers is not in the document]
justinwb: Have cleared out the requirement for inspection of the RDF body for metadata in general discovery. Since the scope has been narrowed from overall resource metadata to auxiliary resources, we can leave it up to the a given auxiliary resources types to define whether it is necessary to check the RDF body as part of processing.

main/resource-access.bs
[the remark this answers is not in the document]
justinwb: Added issue inline for discussion.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+      <td>```http://www.w3.org/ns/solid/terms#configuration```</td>
+    </tr>
+  </tbody>
+</table>
+
+#### Discovery of Annotated Resource #### {#rm-annotated}
+
+Certain metadata resource types require the Solid server to link back to the
+annotated resource that the metadata is directly associated with, via Link
+headers. In these instances, the link relation ```rel=describes``` MUST be used.
```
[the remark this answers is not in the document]
justinwb: Added issue inline for discussion

main/resource-access.bs
justinwb: Adjusted to SHOULD, as I think it should at a minimum be encouraged.

main/resource-access.bs
justinwb: The definition of a server managed resources is that the permissions are restricted to read-only (modifiable only by the server). If the permissions aren't limited then it isn't server managed.

main/resource-access.bs
justinwb: Have changed the issue to better explain the point - "Issue: Consider where there are any common parameters that would be
ubiquitous across resource descriptions that should be defined as part of the specification."

main/resource-access.bs
justinwb: Added issue inline for tracking

main/resource-access.bs
justinwb: If we take this stance, then we are saying that effectively any client that wants to provide resource description needs control privileges, because the server won't be in a position to do it for them automatically. We should be doing as much as we can to ensure that basic operations can be done without control permissions, because otherwise we'll end up with basic clients that have much more privilege than they should. Issue created inline to track this.

main/resource-access.bs
justinwb: What this is saying is that we define a shape to validate this shape validation auxiliary resources. The auxiliary resource is not the shape itself, but a pointer to it, with some additional parameters.

justinwb: Have updated the issue a bit to be more explicit.

main/resource-access.bs
justinwb: Issue created inline for this.

main/resource-access.bs
justinwb: See prior comment - the auxiliary resource is not the shape itself, it is a pointer to the shape that should be used for validation along with other parameters. There is another issue inline above this to provide the shape that this resource should conform to.

main/resource-access.bs
```diff
@@ -75,6 +75,352 @@ Issue: Write Linked Data Platform section.
...
+An auxiliary resource that resides on a Solid server MUST adhere to the same
+interaction model used by other regular Solid resources, except where specified
+in the definition of that auxiliary resource type.
+
+### Required Client-side Implementation ### {#ar-client}
+
+#### Discovery of Auxiliary Resources #### {#ar-discovery}
+
+To discover the auxiliary resources directly associated with a given Solid
+resource, a Solid client MUST issue a `HEAD` or `GET` request to the target
```
dmitrizagidulin, line 142: How do we feel about an OPTIONS request as well?

RubenVerborgh: https://www.mnot.net/blog/2012/10/29/NO_OPTIONS

dmitrizagidulin: Ahhh got it, thank you. That's very helpful.

main/resource-access.bs
```diff
@@ -75,6 +75,352 @@ Issue: Write Linked Data Platform section.
...
+
+#### Shape Validation #### {#ar-shape}
+
+Shape Validation auxiliary resources as defined by (link to shape validation)
+SHOULD be supported as an auxiliary type by Solid servers.
+
+The Shape validation auxiliary resource directly associated with a given
+resource
+is discovered by the client via the `rel=http://www.w3.org/ns/solid/terms#shape`
+`Link` relation in a `Link` header. Conversely, the resource being described by
```
dmitrizagidulin, line 306: What's the history behind having two descovery mechanisms for shape? (shape and describes)?

dmitrizagidulin:
+1 from me, this is a _crucial_ and quite exciting addition to the spec.
(Also, I will definitely be bringing it up for discussion / consideration at the Secure Data Storage WG).

main/resource-access.bs
```diff
@@ -75,6 +75,352 @@ Issue: Write Linked Data Platform section.
...
+Note: Consider where there are any common parameters that would be
+ubiquitous across resource descriptions that should be defined as part of the
+specification.
+
+Resource description is a general mechanism to provide descriptive metadata for
+a given resource. It MUST be supported as an auxiliary type by Solid
+servers.
+
+The Descriptive auxiliary resource directly associated with a given resource is
+discovered by the client via the `"rel=describedby"` `Link` relation in a `Link`
```
acoburn, line 262: Minor syntactical adjustment

```suggestion
discovered by the client via the `rel="describedby"` `Link` relation in a `Link`
```

main/resource-access.bs
justinwb: We use this convention on all of the auxiliary resources that have a 1:1 relationship with the resource they have been associated with. Essentially its just a way to have a bi-directional association, specifically:

Resource A -> Shape Validation V
Shape Validation V -> Resource A

Note that the shape validation resource is not meant to actually link straight to the shape itself, but to an RDF document that will point to the shape(s), contain some other helpful metadata like type of shape(s), enforcement level - (warn, fail), etc.

Now that aux. resource section is wrapping I plan to start documenting that one, so there should be more detail soon.

main/resource-access.bs
```diff
@@ -75,6 +75,291 @@ Issue: Write Linked Data Platform section.
...
+#### Configuration #### {#rm-configuration}
+
+Issue: Identify any configuration metadata parameters that MUST be recognized by
+all implementations.
+
+Configuration metadata is used to store configurable parameters for a given
+resource. For example, whether to maintain
+[Mementos](https://tools.ietf.org/html/rfc7089) for a given resource, or how the
+data within a given resource should be rendered or indexed. It MUST be supported
+as a resource metadata type by Solid servers.
```
csarven: Perhaps: "For example, whether to maintain versioned resources, or how the data within a given resource should be rendered, indexed, or processed further." ?

csarven: "rendered" being?

csarven: Is the primary user of the contents of a configuration resource the server? Similar to above re "process", is there a statement missing about the conformance requirement eg. "Server MUSTSHOULDMAY process request payload when configuration resource is created or updated."

Is the information in configuration of interest to clients (put there by other applications)?

main/resource-access.bs
```diff
@@ -75,6 +75,352 @@ Issue: Write Linked Data Platform section.
...
+
+An auxiliary resource linked with a given Solid resource through an HTTP `Link`
+header is considered to be *directly associated* with that resource. It is up
+to the server to determine how that association affects processing based on the
+auxiliary resource type.
+
+A given Solid resource MAY link to zero or more auxiliary resources. A given
+Solid resource MAY Link to auxiliary resources on a different server under a
+different authority, per the configuration of the Solid server on which that
+resource resides.
```
csarven, line 127: This entails that an ACL resource can be on a different server.. and so raises security issues that should be addressed. Are there existing issues tackling some of them? If the following is intended for WAC to address, that's fine (or if this text is written with known answers to the questions below, that's fine too).. but I would suggest that what's expected from ACLs should be aligned as opposed to assuming that it will work out somehow.

* Will agent(s) controlling a resource on server-A be authorized to request write operations on the ACL on server-B? Ditto the application.

* What should happen if/when remote ACL becomes inaccessible (for whatever reason for any period)?

* What should happen if/when remove ACL is compromised?

main/resource-access.bs
```diff
@@ -75,6 +75,352 @@ Issue: Write Linked Data Platform section.
...
+is authoritative for both resources.
+
+Issue: Provide a shape to validate a shape validation auxiliary resource. May
+include the shape language, shape url, and any additional parameters to be used
+in shape validation by the server implementation.
+
+A Solid server SHOULD sanity check Shape validation auxiliary resources upon
+creation or update to restrict invalid changes.
+
+#### Server Managed #### {#ar-managed}
```
csarven, line 346: What are differences between server-managed containment triples and the triples in server-managed resources (auxiliary)? It appears to be that we have two different mechanisms in the spec that are conceptually similar, if not the same. I suggest to at least partly address: https://github.com/solid/specification/issues/177 .

main/resource-access.bs
justinwb: Configuration type has been removed from submission, now being tracked at #178

justinwb: Configuration type has been removed from submission, now being tracked at #178

main/resource-access.bs
justinwb: I think that we can actually remove this statement, and even be explicit that we do not permit it. If any use cases present themselves that would force us to relax it, we can reevaluate. I've gone back and looked at the genesis of this, and don't see a good reason beyond an attempt for added flexibility in the future.

main/resource-access.bs
csarven: >Note that the shape validation resource is not meant to actually link straight to the shape itself, but to an RDF document that will point to the shape(s), contain some other helpful metadata like type of shape(s), enforcement level - (warn, fail), etc.

Why isn't the shape resource self-describing ie. including both the shape and anything else that's relevant?

The name of the property signals "shape", the resource is called "shape resource".. but when the client gets there, they are directed to another resource which is the actual shape?

Too many meta layers for my taste.

We have rel=acl, describedby,.. targeting the most essential resource. How is the metadata (eg. enforcement level) about the shape resource (ie. the target of rel=shape) is of use to the client?

main/resource-access.bs
justinwb: Answered in #177

main/resource-access.bs
justinwb: > Why isn't the shape resource self-describing ie. including both the shape and anything else that's relevant?

The shape resource is public and doesn't know anything about the context in which you are applying it. For example, you may only want to warn when the validation fails. That a decision you make in how to apply the validation, not the author of the shape.

> The name of the property signals "shape", the resource is called "shape resource".. 

Agree that the property name is misleading - it might be best to change it to something else like "constraint" or "validation". Also there's nowhere in this document that the words "shape resource" are used?

> but when the client gets there, they are directed to another resource which is the actual shape?

The purpose of this auxiliary resource is to illustrate one or more validation constraints set on that resource. It's not limited only to the shape itself, but also other parameters related to how that validation should be evaluated / applied.

dmitrizagidulin:
Minor request - would it be possible to have an example `HEAD` request that includes all of the auxiliary resource types in the response? Just so that developers can see it at a glance?

justinwb:
> Minor request - would it be possible to have an example `HEAD` request that includes all of the auxiliary resource types in the response? Just so that developers can see it at a glance?

@dmitrizagidulin I agree this is useful. I wrote one, but then it felt like it would fit better in a corresponding primer document to the solid protocol spec. Thoughts?

main/resource-access.bs
justinwb: Per the consensus in #177, server-managed as a broad catch-all has been removed from the proposal, and will be replaced with more specific relations in separate submissions.

main/resource-access.bs
justinwb: This has been [updated](https://github.com/solid/specification/pull/156/commits/4b48d5fc42273789e07c20c853b5bf4592641aff) to constrain auxiliary resources to the same server / authority, with issue #176 noted inline to determine whether MUST NOT is appropriate.

dmitrizagidulin:
@justinwb oh, if you have one, please add it to this PR! We definitely need all the examples we can get our hands on, in the spec.

csarven:
I agree. It is useful to give a brief example in this context. We can always edit later.

justinwb:
> @justinwb oh, if you have one, please add it to this PR! We definitely need all the examples we can get our hands on, in the spec.

@dmitrizagidulin @csarven expanded the discovery examples to cover all types in [this commit](https://github.com/solid/specification/pull/156/commits/93e92098f4c7dabb4bb62bdb917504550e65a313)

dmitrizagidulin:
@justinwb Perfect, thanks!

main/resource-access.bs
csarven: >The purpose of this auxiliary resource is to illustrate one or more validation constraints set on that resource. It's not limited only to the shape itself, but also other parameters related to how that validation should be evaluated / applied.

That's fine but I don't see a reason why that's not all (including all constraints as well as any other related information) be stated in the auxiliary shape resource.

>A resource representation whose shape is constrained by a given ShEx schema that links to an auxiliary resource defining that schema. 

Sounds closer to what I'm saying than what you're saying but perhaps I'm misinterpreting.

In any case, I can't draw the underlying rationale from the text or arguments. I'm suspecting two reasons for having an intermediary (auxiliary) resource:

* the lifecycle of the shape (constraints) information shouldn't depend on the lifecycle of the primary resource, so the auxiliary resource hop is a way to let the shapes live on. On deleting primary resource, auxiliary resource can get deleted without touching the shapes (which happen to be elsewhere and possibly used by other resources).

* there are multiple approaches to shape validation (SHACL/ShEx) so instead of picking one to associate resources with, we have the intermediary auxiliary resource to point at them.

The first seems reasonable.

The second less so - after all, no particular reason why there can't be specific link relations point at shacl or shex resources. On that note, at least one of these things needs to be required in order to have minimum level of interop.

justinwb:
> I can't say I'm completely comfortable with the decision around auxiliary shape resource acting as a way to discover the actual shape validation (+ "metadata") resources but I'm okay to not block the PR for that either at this point - it is a design decision. I hope we can address the key concerns in: https://github.com/solid/specification/pull/156/files/93e92098f4c7dabb4bb62bdb917504550e65a313#r420914309 at some point. As I read it, how interop will happen around shapes is not clear cut. One client may use SHACL, the server may be okay with it, but another client may only understand ShEx.

Completely understand and agree. There is another section that @ericprud and i plan on contributing that will detail the specifics for shape validation. As part of authoring and reviewing that I'm confident we'll address those concerns.

Issue #54, opened 2019-09-20
WebAC link relation header


acoburn:
The Solid Web Access Control document describes a [mechanism for ACL resource location discovery](https://github.com/solid/web-access-control-spec#acl-resource-location-discovery). That involves inspecting a Link header with the `rel="acl"` parameter.

That link relation, however, is not registered with [IANA](https://www.iana.org/assignments/link-relations/link-relations.xhtml). Alternatively, many Linked Data specifications ([LDP](https://www.w3.org/TR/ldp/#ldpr-gen-pubclireqs), [LDN](https://www.w3.org/TR/ldn/#discovery), [OA](https://www.w3.org/TR/annotation-protocol/#discovery-of-annotation-containers)) make use of IRIs for use with these sorts of discovery protocols. In the second case, an IRI could be defined in the [ACL vocabulary](http://www.w3.org/ns/auth/acl) without needing to register the relation with IANA.

More discussion of this issue is at https://github.com/solid/web-access-control-spec/issues/21

timbl:
Yes, Let's register this with IANA.    We should definitely not change to use a rel=<URI>, as that would break Solid implementations.

csarven:
Request to register "acl" link relation type: https://github.com/protocol-registries/link-relations/issues/32

Issue #47, opened 2019-09-20
Which LDP container types does a Solid server need to support


acoburn:
The LDP specification defines three [container types](https://www.w3.org/TR/ldp/#ldpc): Basic, Direct and Indirect, none of which is required for LDP conformance. Most Solid use cases require support for Basic containment. Will the Solid specification require (`MUST`) support for Basic containment? Will the Solid specification take any position on the other container types?

pmcb55:
I'm surprised the current spec doesn't explicitly make LDP Basic Container support a MUST, but I certainly assume the new version of the spec will. I don't see a strong need to say anything beyond that at the moment (i.e. I don't see a need to say a server MAY implement Direct or Indirect), at least not until we find good use-cases where Direct containment would be really useful (which there very well could be!).

RubenVerborgh:
> I'd be surprised if the current spec doesn't already make LDP Basic Container support a MUST

The v0.x draft doesn't make any normative statements, so no 🙂

That's precisely the reason why we need to go through this: use more exact language so it is unambiguous what needs to be implemented and what not.

csarven:
LDP-BC probably covers majority of the applicable cases and the implementation experience to date.

Anecdotal evidence from the vicinity of Solid: I haven't seen any server implementation with LDP-DC and LDP-IC support or any Solid application making use of it.. or any major request to have them (minor: https://github.com/solid/node-solid-server/issues/399 ).

DC and IC impose a particular information model and management. We have considerations on Shapes to do that in a richer way. Less hidden magic by the server probably better. TSE doesn't need to forbid anything ie. takes no position: just let DC and BC support fallback to LDP.

I'd suggest that only `MUST` LDP-BC. If this is not a `MUST`, there is no strong interop among anything pertaining to writing and discovering a class of resources. (I'm assuming that it'd be odd to have DC and/or IC but not BC... or even just LDPC without specifics.) In fact, a number of other LDP-based features can also drop or don't say anything more than what LDP says.

TSE can revisit supporting DC and IC at a later date; starting with demand + implementation experience.

kjetilk:
This seems like an issue we can have a "rough consensus" process around, doesn't it?

csarven:
Proposal following F2F meeting of 2019-10-29:
LDP-BC should be the only required type of containment.

dmitrizagidulin:
+1 to this. (MUST + only LDP-BC)

RubenVerborgh:
Clarified during call of 2019-12-04

[...]

Issue #46, opened 2019-09-19
Should a Solid server limit the reuse of URLs?


acoburn:
In a non-normative section, the LDP specification states that [LDP servers should not re-use URIs](https://www.w3.org/TR/ldp/#h-ldp-webarch-uri-reuse). Does the Solid specification take a stance on this, one way or the other? What if a client deletes a well-known URL, such as `/profile/card` or `/public/inbox`?

RubenVerborgh:
I follow the text of the LDP spec there; the intention is that the same URI is never used to identify another resource. In practice, servers cannot know whether a resource that was deleted but then reinstated is the same or different.

Hence, I think this is non-enforceable.

csarven:
It is not that it is impossible but that such setup requires additional machinery applied consistently over time. Certainly immutable snapshots of resources can exist and they can be used to compare if the server and the URI owner is willing. As one one would expect, having anything along those lines - more importantly the promise - is complicated for all parties involved.

Limiting URI reuse would not only require a higher bar for servers but for clients as well. It entails that clients will also need to know how to handle the failure ie. noting that a particular URI use differs from a previous issue.

I would suggest that the Solid spec should not set this as a requirement. The LDP covers at a good level in that it is non-normative and just throws in "LDP servers should not re-use URIs". That sounds like making a best practice suggestion to me.

I would also add that when we have more clarity on how immutable resources are used in Solid, that may help to phrase non-normative text about how it may be purposed to handle URI reuse, collisions, etc.

RubenVerborgh:
Warning: pedantic comment ahead.

> It is not that it is impossible but that such setup requires additional machinery applied consistently over time. Certainly immutable snapshots of resources can exist and they can be used to compare if the server and the URI owner is willing.

The only thing we can achieve with that is enforcing the same _representation_, but not the same _resource_. Hence, we do not enforce that the same URI is never reused with another resource.

A resource (in the REST sense) is a conceptual relationship, whereas a representation is a value of that resource at a given point in time.

Suppose I define the resource https://pod.example/weather/today as today's whether. Then I would update representations such that:
- Monday: sun
- Tuesday: rain
- Wednesday: sun

Now suppose that, after Wednesday, I get tired of maintaining the weather, so I `DELETE` it. But on Friday I change my mind and I want to put:
- Friday: snow

A system that remembers representations would not allow me to do that, because it is a different representation from the last one.

However, the resource definition never changed.

We can also find examples in the opposite direction.

---

[...]

PR #157, opened 2020-03-26
Some criteria based on rough consensus

csarven:
Commits include some of the criteria from the issues based on rough consensus. Topics vary but some emphasis on "resource access".

The request for review is to confirm what's captured here is accurate ie. the raw criteria. Text needs to be revised and re-organised in the spec but that is intended to be addressed in a future PR.

Edit: I'd like to suggest that technical objections to the criteria should be taken up in the referenced issues in order to better preserve the rationale in context of how they came through.

### DIFF
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -18,6 +18,8 @@ with mechanisms for controlling who can access what.
 A <dfn export>Solid app</dfn> is an application
 that reads or writes data from one or more [=data pods=].
 
+A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set resources atomically and recursively. Based on consensus https://github.com/solid/specification/issues/126#issuecomment-569920473 .
+
 Issue: Introduce the structure of this document.
   [Cross-server interoperability](#resource-access)
   [Cross-app interoperability](#clients)
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -20,7 +20,7 @@ that reads or writes data from one or more [=data pods=].
 
 A <dfn export>read operation</dfn> entails that information can be revealed about the content of a resource, and categorically excludes operations to "write", as well as unintended exposure or side-effects within that. [[Source](https://github.com/solid/specification/issues/149#issue-568433265)]
 
-A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/126#issuecomment-569920473)]
+A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set of resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/126#issuecomment-569920473)]
 
 An <dfn export>append operation</dfn> entails an update in that information can be added but not removed, and categorically excludes operations to "read" and "replace", as well as unintended exposure or side-effects within that. An append operation affects a set resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/118#issuecomment-569648485)]
 
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -18,6 +18,8 @@ with mechanisms for controlling who can access what.
 A <dfn export>Solid app</dfn> is an application
 that reads or writes data from one or more [=data pods=].
 
+A <dfn export>read operation</dfn> entails that information can be revealed about the content of a resource, and categorically excludes operations to "write", as well as unintended exposure or side-effects within that. Based on conensus https://github.com/solid/specification/issues/149#issue-568433265 .
+
 A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set resources atomically and recursively. Based on consensus https://github.com/solid/specification/issues/126#issuecomment-569920473 .
 
 An <dfn export>append operation</dfn> entails an update in that information can be added but not removed, and categorically excludes operations to "read" and "replace", as well as unintended exposure or side-effects within that. An append operation affects a set resources atomically and recursively. Based on consensus https://github.com/solid/specification/issues/118#issuecomment-569648485 .
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -18,11 +18,11 @@ with mechanisms for controlling who can access what.
 A <dfn export>Solid app</dfn> is an application
 that reads or writes data from one or more [=data pods=].
 
-A <dfn export>read operation</dfn> entails that information can be revealed about the content of a resource, and categorically excludes operations to "write", as well as unintended exposure or side-effects within that. Based on conensus https://github.com/solid/specification/issues/149#issue-568433265 .
+A <dfn export>read operation</dfn> entails that information can be revealed about the content of a resource, and categorically excludes operations to "write", as well as unintended exposure or side-effects within that. [[Source](https://github.com/solid/specification/issues/149#issue-568433265)]
 
-A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set resources atomically and recursively. Based on consensus https://github.com/solid/specification/issues/126#issuecomment-569920473 .
+A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/126#issuecomment-569920473)]
 
-An <dfn export>append operation</dfn> entails an update in that information can be added but not removed, and categorically excludes operations to "read" and "replace", as well as unintended exposure or side-effects within that. An append operation affects a set resources atomically and recursively. Based on consensus https://github.com/solid/specification/issues/118#issuecomment-569648485 .
+An <dfn export>append operation</dfn> entails an update in that information can be added but not removed, and categorically excludes operations to "read" and "replace", as well as unintended exposure or side-effects within that. An append operation affects a set resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/118#issuecomment-569648485)]
 
 Issue: Introduce the structure of this document.
   [Cross-server interoperability](#resource-access)
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -18,11 +18,11 @@ with mechanisms for controlling who can access what.
 A <dfn export>Solid app</dfn> is an application
 that reads or writes data from one or more [=data pods=].
 
-A <dfn export>read operation</dfn> entails that information can be revealed about the content of a resource, and categorically excludes operations to "write", as well as unintended exposure or side-effects within that. [[Source](https://github.com/solid/specification/issues/149#issue-568433265)]
+A <dfn export>read operation</dfn> entails that information about a resource's existence or its description can be known. [[Source](https://github.com/solid/specification/issues/149#issue-568433265)]
 
-A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set of resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/126#issuecomment-569920473)]
+A <dfn export>write operation</dfn> entails that information about resources can be created or removed. [[Source](https://github.com/solid/specification/issues/126#issuecomment-569920473)]
 
-An <dfn export>append operation</dfn> entails an update in that information can be added but not removed, and categorically excludes operations to "read" and "replace", as well as unintended exposure or side-effects within that. An append operation affects a set resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/118#issuecomment-569648485)]
+An <dfn export>append operation</dfn> entails that information can be added but not removed. [[Source](https://github.com/solid/specification/issues/118#issuecomment-569648485)]
 
 Issue: Introduce the structure of this document.
   [Cross-server interoperability](#resource-access)
--- a/main/introduction.bs
+++ b/main/introduction.bs
@@ -20,6 +20,8 @@ that reads or writes data from one or more [=data pods=].
 
 A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set resources atomically and recursively. Based on consensus https://github.com/solid/specification/issues/126#issuecomment-569920473 .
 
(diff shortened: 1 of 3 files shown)


main/introduction.bs
```diff
@@ -18,6 +18,12 @@ with mechanisms for controlling who can access what.
 A <dfn export>Solid app</dfn> is an application
 that reads or writes data from one or more [=data pods=].
 
+A <dfn export>read operation</dfn> entails that information can be revealed about the content of a resource, and categorically excludes operations to "write", as well as unintended exposure or side-effects within that. [[Source](https://github.com/solid/specification/issues/149#issue-568433265)]
```
RubenVerborgh: (nit) Not just content but also existence, metadata, etc.?

main/introduction.bs
```diff
@@ -18,6 +18,12 @@ with mechanisms for controlling who can access what.
...
+
+A <dfn export>write operation</dfn> entails an update in that information can be created or removed, and categorically excludes operations to "read", as well as unintended exposure or side-effects within that. A write operation affects a set of resources atomically and recursively. [[Source](https://github.com/solid/specification/issues/126#issuecomment-569920473)]
```
RubenVerborgh: "set of" and "recursively" seem to be mutually exclusive. Either it is one resource, and child resources are affected recursively, or we define a set of resource (possibly using recursive means) and the operation affects all members in the set.

main/resource-access.bs
```diff
@@ -67,13 +69,35 @@ if it needs to access resources requiring authentication (see [[#webid]]).
 When it receives a response with a `403` or `404` status code,
 it MAY repeat the request with different credentials.
 
+A Solid client MUST use the `Content-Type` request header in `PUT`, `POST` and `PATCH` requests. [[Source](https://github.com/solid/specification/issues/70#issuecomment-547924171)]
```
RubenVerborgh: Ref to RFC as to _how_ it should be used?

main/resource-access.bs
```diff
@@ -67,13 +69,35 @@ if it needs to access resources requiring authentication (see [[#webid]]).
...
 
-## Linked Data Platform ## {#ldp}
+## Uniform Resource Identifier ## {#uri}
 
-Issue: Write Linked Data Platform section.
+### Required shared slash semantics ### {#uri-path-component}
 
-Draft:
-A Solid data pod MUST conform to the LDP specification [[!LDP]].
+Server and client share the slash semantics in the URI path component representing hierarchy [[!RFC3986]]. [[Source](https://github.com/solid/specification/issues/35#issuecomment-547949014)]
```
RubenVerborgh: …, that is: (explain exactly what this entails and perhaps also does not entail).

main/resource-access.bs
```diff
@@ -67,13 +69,35 @@ if it needs to access resources requiring authentication (see [[#webid]]).
...
+
+Server MUST NOT create a URI differing only in the trailing slash with that of an existing URI. [[Source](https://github.com/solid/specification/issues/107#issuecomment-567482817)]
```
RubenVerborgh: Servers don't create URIs; URIs always exist. What exactly do we mean? That a redirect is needed? That either `foo` or `foo/` should yield a 404?

main/resource-access.bs
```diff
@@ -67,13 +69,35 @@ if it needs to access resources requiring authentication (see [[#webid]]).
...
+
+Server MAY redirect a request with 301 from a non-existing URI to an existing URI with a matching path component ending with the same segment term. Behaviour pertaining to authorization MUST proceed this optional redirect. [[Source](https://github.com/solid/specification/issues/107#issuecomment-567454889)]
```
RubenVerborgh: Same: URIs always exist.

proceed => precede?

main/resource-access.bs
```diff
@@ -67,13 +69,35 @@ if it needs to access resources requiring authentication (see [[#webid]]).
...
+
+## Linked Data ## {#linked-data}
+
+### Containment ### {#resource-containment}
+
+Solid has the notion of containers to represent a collection of linked resources to help with resource discovery and lifecycle management.
+
+There is a 1-1 correspondence between containment triples and relative reference within the path name hierarchy. [[Source](https://github.com/solid/specification/issues/98#issuecomment-547506617)]. It follows that all resources are discoverable from a container and that it is not possible to create orphan resources. [[Source](https://github.com/solid/specification/issues/97#issuecomment-547459396)]
+
+The representation and behaviour of containers in Solid is corresponds to LDP Basic Container and MUST be supported. [[Source](https://github.com/solid/specification/issues/47#issuecomment-561675764)]
```
RubenVerborgh: Does this mean we cannot have other container types?

main/resource-access.bs
```diff
@@ -67,13 +69,35 @@ if it needs to access resources requiring authentication (see [[#webid]]).
...
+
+Server MUST create intermediate containers as well as corresponding containment triples derived from the URI path component of `PUT`, `POST`, and `PATCH` requests. [[Source](https://github.com/solid/specification/issues/68#issuecomment-561690124)]
```
RubenVerborgh: Mno, we cannot check if the server _creates_ containment triples. Those containment triples need to be present in representations, yes.

main/resource-access.bs
```diff
@@ -67,13 +69,35 @@ if it needs to access resources requiring authentication (see [[#webid]]).
...
+
+### Persistence ### {#uri-persistence}
+Servers should not re-use URIs, regardless of the mechanism by which resources are created. Certain specific cases exist where URIs may be reinstated when it identifies the same resource, but only when consistent with Web architecture's <cite><a href="https://www.w3.org/TR/webarch/#URI-persistence">URI persistence</a></cite> [[!WEBARCH]]. [[Source](https://github.com/solid/specification/issues/46#issuecomment-589619372)]
```
RubenVerborgh: 1) I think there is no technical way to enforce this. It's a non-computable problem.

2) If there is, then we should write down that technical way instead of the above explanation. 

Also SHOULD NOT in caps.

main/resource-access.bs
```diff
@@ -106,6 +130,9 @@ Issue: Write Web Access Control section.
 Draft:
 A Solid data pod MUST conform to the Web Access Control specification [[!WAC]].
 
+A resource can advertise an ACL document that is directly associated by using the HTTP `Link` header with a `rel` value of `acl`. [[Source](https://github.com/solid/specification/issues/31#issuecomment-548360553)]
```
RubenVerborgh: Add TODO: registration.

main/resource-access.bs
csarven: A URI doesn't "exist" until it is named, registered, or associated with a resource... but I get your point, will update. Yes, a redirect in the following line.

RubenVerborgh: > A URI doesn't "exist" until it is named, registered, or associated with a resource...

It does: https://csarven.ca/it-does
A URI is a string conforming to a certain pattern. The set of all URIs that will ever be minted exists, it just happens to be infinite.

So everything needs to be phrased in terms of resources being created.

csarven:
I'd like to suggest that technical objections to the criteria should be taken up in the referenced issues in order to better preserve the rationale in context of how they came through.

dmitrizagidulin:
Important formatting question: can we make sure we line-break at 80 chars, which is best practice for GitHub-hosted source code and specs? Having essentially "one line per paragraph" leads to way too many UI problems in reviewing.

main/introduction.bs
dmitrizagidulin: I'm not sure I understand what line 21 is saying.. Any chance to re-phrase it simpler?

main/resource-access.bs
csarven: Not sure how deep or philosophical (mathematical?) you want to go, but I hope we can hover on being pragmatic. Sure, I get it, certainly a set of all things exists on a plane but not sure how you want to put that knowledge into actual use. We are basing things off RFC, WebArch.. have I overlooked your exists or can't recall a DesignIssue?

URIs exist to identify a resource. Server advertises a name because it happened to associate a resource with a URI - I used "create" as a generally comprehensible terminology. I'm okay to change that though.

RubenVerborgh: No philosophy indeed, pragmatic but exact.

As I wrote above, it’s a resource that is being created on the server; that’s the state change that happens and that we should be describing across the entire document.

It’s important to be exact, as URIs can be minted before the corresponding resource is created. Common case in point: an ACL URL for a specific resource (while that resource currently inherits another ACL).

main/resource-access.bs
RubenVerborgh: Counter-example to one possible interpretation in which resources cannot be recreated after they have been deleted: a specific ACL (which is also a resource) might be deleted and be needed again at a later time.

main/resource-access.bs
csarven: The ACL example is precisely what I mean by naming. The URI starts to exists because it is named / written down. That's being pragmatic. There is an artifact, the string, spelled out, it is visible.. we can talk about it. On the other hand, what's the point in saying that something somehow exists but not actually name it? That'd be being philosophical. Like I said, the context of "create" was because - and I'll use you own words - "a resource is **created** on the server" and that it is identified with a URI.

I'll revise tomorrow.

RubenVerborgh: Got it. Suggestion:

"If two URIs differ only in the trailing slash, and the server has associated a resource with one of them, then the other URI MUST NOT correspond to another resource. Instead, the server MAY respond to requests for the latter URI with a 301 redirect to the former. Behaviour pertaining to authorization MUST proceed this optional redirect."

main/resource-access.bs
csarven: https://github.com/solid/specification/issues/46#issuecomment-589619372 was agreed as non-normative. It was derived from LDP's notable information. The general case doesn't need to be "computable". Note also "we don't specify how a server or a URI owner determines representation consistency (re AWWW's good practice)" in https://github.com/solid/specification/issues/46#issuecomment-602511328. The specific case(s) eg. 401, are intended to be revisited after ~FPWD. If you still object, can we capture that in the issue?

Where the practice is applicable the text is about reinstating resources and their continuity as per URI persistence. Inability to reinstate is not covered.

main/resource-access.bs
csarven: Do you mean IANA registration for "acl"? That and possibly the URI for "acl" is covered in the other issues eg. https://github.com/solid/specification/issues/54 . It will make its way in.

main/introduction.bs
csarven: Right. The context of "content" is information general, as opposed to strictly the HTTP message body. It'd include "existence" but let's mention to clarify any way.

main/resource-access.bs
RubenVerborgh: Non-normative is fine, but shall we then mark as such?

csarven: I was reserving uppercase keywords for normative but we can also lead the text with non-normative heading.

main/introduction.bs
csarven: What do you think of this revision: https://github.com/solid/specification/pull/157/commits/8c28e8edd6bb6e19f21200b870736cd344a095be

main/introduction.bs
csarven: What do you think of this revision: https://github.com/solid/specification/pull/157/commits/8c28e8edd6bb6e19f21200b870736cd344a095be

main/introduction.bs
csarven: What do you think of this revision: https://github.com/solid/specification/pull/157/commits/8c28e8edd6bb6e19f21200b870736cd344a095be

main/resource-access.bs
csarven: Ok https://github.com/solid/specification/commit/d0b199ae46a642100dcecaba264084defe5dcae8 . Those methods will probably have their own sections going in more detail.

main/resource-access.bs
csarven: https://github.com/solid/specification/commit/44a0c2d9408be26fe44a82bd0a0a7a25378ce838

main/resource-access.bs
csarven: No. The statement only mentions what's required. Optionals or recommendations against anything is not mentioned - normally they're allowed but would be out of spec behaviour. The linked source and https://github.com/solid/specification/issues/47#issuecomment-587421049 suggests to leave the descriptions for the other cases (if at all) for later.

main/resource-access.bs
csarven: Noted. That "create" reflected what's agreed, or at least wasn't intended the way you are framing it. In any case, I've revised to focus on the representation: https://github.com/solid/specification/pull/157/commits/13603c04cea9f2dd9d28b0f8fdb431bcfbb6c196

main/resource-access.bs
csarven: Revised: https://github.com/solid/specification/pull/157/commits/2abbeca6554a22529aae45b4eb925bf6729b8d3b . Entailment is further clarified in the other sections.

main/introduction.bs
RubenVerborgh: 👍

main/introduction.bs
RubenVerborgh: I see no issues; curious if we need to handle recursion elsewhere now (but not a blocker).

main/resource-access.bs
RubenVerborgh: OK; just noticed typo _is corresponds_.

csarven: Thanks. Caught that earlier https://github.com/solid/specification/pull/157/commits/53fbc185e1a7e6d1d151870c94973ed9e8d60496

main/introduction.bs
csarven: Figured generalised description may not suffice or particularly useful here. It may be best to describe recursion in context of something with the interdepencies.

main/introduction.bs
dmitrizagidulin: @csarven clear, thank you!

main/introduction.bs
```diff
@@ -18,6 +18,12 @@ with mechanisms for controlling who can access what.
 A <dfn export>Solid app</dfn> is an application
 that reads or writes data from one or more [=data pods=].
 
+A <dfn export>read operation</dfn> entails that information about a resource's existence or its description can be known. [[Source](https://github.com/solid/specification/issues/149#issue-568433265)]
```
justinwb, line 21: Is the intent for these citations to github issues / comments to persist indefinitely, or only during the drafting process?

main/introduction.bs
```diff
@@ -18,6 +18,12 @@ with mechanisms for controlling who can access what.
...
+
+A <dfn export>write operation</dfn> entails that information about resources can be created or removed. [[Source](https://github.com/solid/specification/issues/126#issuecomment-569920473)]
```
justinwb, line 23: Wondering whether we need to explicitly mention delete here

main/introduction.bs
csarven: Only drafting so that there is a record somewhere (commit history) so that it could be look up if needed. I'm open to the idea of having them persist after.

main/introduction.bs
csarven: Perhaps. I've created an issue to track this question: https://github.com/solid/specification/issues/165 and some background on why it is not currently there. Unless the existing definitions are unclear with respect to "delete", can we revisit in future? We can move the issue forward in the meantime.

main/introduction.bs
justinwb: Happy to have them in there during drafting so that addresses my short-term question. Agree it'd be good to keep track of them long-term, though I think we should discuss _where_ to do so when we get to that point.

main/introduction.bs
justinwb: Issue looks good - agree we can revisit in future.

Issue #128, opened 2019-12-04
Clarify the heuristics to determine the interaction model if none is specified


RubenVerborgh:
In case the client does not specify an LDP interaction model, the server will determine the interaction model in a heuristic way. This issue is an action item for making these heuristics concrete and writing them down, such that all servers implement them in the same way.

Related: #105, #121

kjetilk:
Since we don't have the actual clarification here, I'll return this to TODO.

kjetilk:
I think there are two parts to this, one is the specific heuristics, which is simple. The other thing is consistency, which is probably harder. The latter was more the intention of #121 , but it really is connected. 

My proposal for the heuristics:

1. Iff the request-URI [path component](https://tools.ietf.org/html/rfc3986#section-3.3) ends with the character `/` is an LDP-C. Since only LDP-BC is supported by Solid, it is permissible to interprete it as that. Note, however, that in Solid, LDP-C doesn't necessarily imply LDP-RS, since it may be represented by an RDF-bearing document.
1. If the request-URI path component does not end with `/`, then the resource is an LDP-RS iff it has a media type that implies that it state is fully represented in RDF, corresponding to an RDF graph.
1. Any other resources are assumed to be LDP-NR.

I'm thinking it is good to define it in terms of media type in the second point, so that the interaction model can always be inferred from headers (since #70), the alternative is to inspect the body to ensure that. We could probably enumerate the media types that are LDP-RS at this point.

I'm still of the opinion that the HTML+RDFa falls between LDP-NR and LDP-RS, (#69), but I'm willing to go with that if it makes stuff simpler.

TallTed:
> [@kjetilk] I'm still of the opinion that the HTML+RDFa falls between LDP-NR and LDP-RS

There *is* no between.  "Fully represented in RDF" (LDP-RS) and "not fully represented in RDF" (LDP-NR) add up to 100% of all possibilities (LDPR). HTML+RDFa docs are not fully represented in RDF, so they are LDP-NR.

Otherwise, I think I agree with your last.

acoburn:
Would it make sense to require the response headers to include an indication of how the server understood the request? That is, if the server understands the new resource to be an LDP-BC, then the server `MUST` respond with a corresponding Link header, indicating an LDP-BC was created (the same would apply to any other LDP interaction model).

kjetilk:
> > [@kjetilk] I'm still of the opinion that the HTML+RDFa falls between LDP-NR and LDP-RS
> 
[...]

Issue #108, opened 2019-11-04
Specify POST


csarven:
In context of LDP, `POST` is used to add resources to containers ie. `POST` to LDPC resource optionally using the `Slug` header and `Link` for interaction model(s).

What should be the `POST` behaviour to target resources that are not LDPCs?

```
POST /foo/bar
````

If the target resource exist, what should happen?

If the target resource doesn't exist, what should happen?

What are the effects of interaction model(s) (LDPR and/or LDPC) used in request's `Link` header?

How should requests to create a resource with the LDP-NR interaction model get handled?

What are the effects without an interaction model?

acoburn:
I am happy to adjust the Trellis behavior to whatever gets specified here, but I will describe the current behavior in the context of your questions.

> What should be the POST behaviour to target resources that are not LDPCs?
>> If the target resource exist, what should happen?

Respond with `405 Method Not Allowed`

> What should be the POST behaviour to target resources that are not LDPCs?
>> If the target resource doesn't exist, what should happen?

Respond with `404 Not Found`

> What are the effects of interaction model(s) (LDPR and/or LDPC) used in request's Link header?

For a successful request (i.e. `/foo/bar` is an LDP-C) the link header would define the interaction model for the newly created resource. A successful response will echo that interaction model back to a client in a Link header.

> How should requests to create a resource with the LDP-NR interaction model get handled?

The body of the request (the raw byte stream) is used to create the new resource at the location hinted at with a `Slug` header (or at a server-generated location in the absence of a `Slug` header). A successful request will respond with the following:

- A `Location` header pointing to the new resource's URL
- A `Link` header with the interaction model of the new resource (i.e. LDP-NR)
- A `Link` header pointing to the location of the corresponding LDP-RS description of this LDP-NR (every LDP-NR has an LDP-RS description)

> What are the effects without an interaction model?

The server uses implementation-specific heuristics to assign an interaction model. I.e. if the `Content-Type` request header is an RDF serialization that is parseable by the server as RDF, the new resource is assigned an LDP-RS interaction model; otherwise, the resource is assigned an LDP-NR interaction model.

csarven:
Raised by @ericprud ( https://gitter.im/solid/specification?at=5dcd102092a84f79fe6477bb ):

Should a `201` response payload include statements related to adding a new resource to a container eg. `POST /foo/bar/` may mention resources which were created (nested containers) and the relations between them, metadata, associated shapes..

kjetilk:
[...]

Issue #40, opened 2019-09-19
Specify semantics of resource creation via PUT


acoburn:
The `PUT` method can be used to replace or create resources. This is explicitly allowable under both [LDP](https://www.w3.org/TR/ldp/#h-ldpr-http_put) and [HTTP](https://tools.ietf.org/html/rfc7231#section-4.3.4).

What does the Solid specification say about `PUT` to _create_ resources (`MAY`, `SHOULD` or `MUST`)?

Specifically, if a resource is created via `PUT`, what is the expectation about how this operation interacts with LDP containment?

For example, if a client creates a resource via `PUT` at `/foo/bar/baz`, should an `ldp:contains` triple be added to the container `/foo/bar`?

What if `/foo/bar` does not exist? Should the server create the container automatically or should the server reject the operation?

What if `/foo/bar` is a non-container resource (`ldp:RDFSource` or `ldp:NonRDFSource`)?

csarven:
Just to eliminate any ambiguity, by "create resources", we consider all ldp Resources; (Non)RDFSources as well as Containers:

>What does the Solid specification say about PUT to create resources (MAY, SHOULD or MUST)?

The thing is that LDP doesn't recommend containment triples to be updated through PUT ( https://www.w3.org/TR/ldp/#ldpc-put-mbrprops ):

>5.2.4.1 LDP servers SHOULD NOT allow HTTP PUT to update a LDPC’s containment triples; if the server receives such a request, it SHOULD respond with a 409 (Conflict) status code.

The reason I'm mentioning is that, if PUT is used, Solid needs to make a decision on whether when a resource is created, it should update the containment triples right there and then, or not. Otherwise, it has to go against the stream (re "SHOULD NOT"). Putting aside the case where the SHOULD NOT recommendation is ignored, PUT without updating containment triples would entail that the created resources won't be discoverable through the container.

Aside: that doesn't imply that created resource is not discoverable through a property other than `ldp:contains`, most likely through membership properties.

It is kind of a one way street. I would suggest that to keep PUT as a MAY (as LDP), but add that server if PUT is used, server MUST update containment triples. Otherwise, containment triples are confined to POST. I'm not sure about any downsides to not updating the containment triples other than the challenge of determining the right container.

I suppose that non-overly LDP-centric clients can still operate in PUT mode without having to rely on POST or POST w/ Slug, if they are authorised and the server is willing any way.

[...]

PR #160, opened 2020-04-03
Add heuristic to determine resource type

csarven:
Criteria for servers to apply heuristics to determine a request's resource type. Based on consensus at https://github.com/solid/specification/issues/128#issuecomment-573033297

### DIFF
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -132,100 +132,28 @@ code.
 
 ### Resource type heuristics ### {#resource-type-heuristics}
 
-The semantics of slashes in URI paths are shared ([[#uri-slash-semantics]]).
+When creating new resources, servers can determine an effective request URI's
+type by examining the URI path ending ([[#uri-slash-semantics]]).
 
-Servers MUST apply the following heuristics to determine the request's
-resource type and to construct the effective request URI for the target
-resource of an HTTP `POST`
-[[Source](https://github.com/solid/specification/issues/128#issuecomment-573033297)].
+Clients who want to assign a URI for their resource, MUST use PUT and PATCH
+requests. Servers MAY allow clients to suggest the URI for a resource created
+through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
 
-The consistency of a request is determined by checking the optional `Link` and
-`Slug` headers against the target URI
-[[Source](https://github.com/solid/specification/issues/40#issuecomment-566995240)].
+Clients who want the server to determine the URI for their resource, MUST use
+the POST request.
 
-Servers supporting the `Slug` header MUST apply the shared slash semantics on
-the `slugtext` on requests that also include the `Link` header with
-`rel="type"`.
+Servers MUST allow creating new resources with a POST request to URI path
+ending `/`. Servers MUST create a resource with URI path ending `/{id}` in
+container `/`. Servers MUST create a container with URI path ending `/{id}/`
+in container `/` for requests including the HTTP `Link` header with
+`rel="type"` targeting a valid LDP container type. Servers MUST handle
+subsequent requests to the newly created container's URI as if it is a valid
+LDP container type by including it the HTTP response's `Link` header.
 
-Issue:
-Redefine `slugtext` eg. exclude recursive container creation?
-[Related issue](https://github.com/solid/specification/issues/128)
+Clients who want to create hierarchical containers MUST use PUT and PATCH
+requests.
 
-<table class="data" id="heuristics-resource-type">
-  <caption>Heuristics to determine a resource type</caption>
-  <thead>
-    <tr>
-      <th>`Slug`</th>
-      <th>`Link`</th>
-      <th>Result</th>
-    </tr>
-  </thead>
-  <tbody>
-    <tr>
-      <td>`foo/`</td>
-      <td>No `Link` header.</td>
-      <td>Create a container with `foo/` appended to the request-URI.</td>
-    </tr>
-    <tr>
-      <td><code>foo/</code></td>
-      <td>`Link` header with `rel="type"` targeting a valid LDP container
-      type.</td>
-      <td>Create a container with `foo/` appended to the request-URI.</td>
-    </tr>
-    <tr>
-      <td><code>foo/</code></td>
-      <td>`Link` header without a `rel="type"` targeting a valid LDP container
-      type.</td>
-      <td>Create a resource with `foo` appended to the request-URI, check for
-      consistency.</td>
-    </tr>
-    <tr>
-      <td>`foo`</td>
-      <td>No `Link` header.</td>
-      <td>Create a resource with `foo` appended to the request-URI, check for
-      consistency.</td>
-    </tr>
-    <tr>
-      <td>`foo`</td>
-      <td>`Link` header with `rel="type"` targeting a valid LDP container
-      type.</td>
-      <td>Create a container with `foo/` appended to the request-URI.</td>
-    </tr>
-    <tr>
-      <td>`foo`</td>
-      <td>`Link` header without a `rel="type"` targeting a valid LDP container
-      type.</td>
-      <td>Create a resource with `foo` appended to the request-URI, check for
-      consistency.</td>
-    </tr>
-    <tr>
-      <td><em>none</em></td>
-      <td>No `Link` header.</td>
-      <td>Create a resource with server determined identifier appended to the
-      request-URI, check for consistency.</td>
-    </tr>
-    <tr>
-      <td><em>none</em></td>
-      <td>`Link` header with `rel="type"` targeting a valid LDP container
-      type.</td>
-      <td>Create a container with server determined identifier appended to the
-      request-URI.</td>
-    </tr>
-    <tr>
-      <td><em>none</em></td>
-      <td>`Link` header without a `rel="type"` targeting a valid LDP container
-      type.</td>
-      <td>Create a resource with server determined identifier appended to the
-      request-URI, check for consistency.</td>
-    </tr>
-  </tbody>
-  <tfoot>
-    <tr>
-      <td colspan="3">
-      </td>
-    </tr>
-  </tfoot>
-</table>
+[[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)].
 
 
 ## Auxiliary Resources ## {#rm}
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -153,12 +153,12 @@ type
     </tr>
     <tr>
       <td><code>foo/</code></td>
-      <td>`Link` header with `ldp:Container` or a specialisation.</td>
+      <td>`Link` header with `rel="type"` targeting a valid LDP container type.</td>
       <td>Create a container with `foo/` appended to the request-URI.</td>
     </tr>
     <tr>
       <td><code>foo/</code></td>
-      <td>`Link` header other than container type.</td>
+      <td>`Link` header without a `rel="type"` targeting a valid LDP container type.</td>
       <td>Create a resource with `foo` appended to the request-URI, check for
       consistency.</td>
     </tr>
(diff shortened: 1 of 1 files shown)


main/resource-access.bs
```diff
@@ -130,6 +130,86 @@ Note:
...
+    </tr>
+    <tr>
+      <td><code>foo/</code></td>
+      <td>`Link` header with `ldp:Container` or a specialisation.</td>
+      <td>Create a container with `foo/` appended to the request-URI.</td>
+    </tr>
+    <tr>
+      <td><code>foo/</code></td>
+      <td>`Link` header other than container type.</td>
+      <td>Create a resource with `foo` appended to the request-URI, check for
```
RubenVerborgh, line 162: This exception I find very weird. I don't think we should derive anything from non-recognized `Link` headers. I.e., all `no` and `other` cases should be identical for me.

main/resource-access.bs
```diff
@@ -130,6 +130,86 @@ Note:
...
+      consistency.</td>
+    </tr>
+    <tr>
+      <td>`foo`</td>
+      <td>No `Link` header.</td>
+      <td>Create a resource with `foo` appended to the request-URI, check for
+      consistency.</td>
```
RubenVerborgh, line 169: Is "check for consistency" defined elsewhere?

csarven:
Good suggestions. Will this work for now: https://github.com/solid/specification/pull/160/commits/320752964c491160f06f3a923be26d6607fbb330 ?

main/resource-access.bs
csarven: Yes, it'll be defined elsewhere via other issues. Mentioned in table footer and refers to https://github.com/solid/specification/issues/40#issuecomment-566995240 .

RubenVerborgh:
Yeah. (Maybe "slashes" for grammar? But I don't mind.)

main/resource-access.bs
csarven: That's indeed what's intended. What's derived is that it is a non-Container resource. Slug will be ignored in this case. Would rephrasing help?

RubenVerborgh: My interpretation of what written seems to be correct (so writing is ok); but I don't understand why from an unknown header can be derived the resource is non-Container. I don't think we should derive that.

csarven: What's recognised is "ldp:Container or a specialisation" (ldp:BasicContainer - as that's the only one currently required) as the "container type". I used "other than container type" to express their absence. Open to rephrasing.

RubenVerborgh: Ah okay, so only `Link` header values with a `type` link relation are included then? Or do we really mean _all_ `Link` header values?

csarven: Only the existence of Link with rel=type LDPC or LDP-BC is checked.

Note the check in prior row says "Link header with ldp:Container or a specialisation." - We are really talking about rel=type here.

So we can just mention the actual relation.

RubenVerborgh: Yes, I would make it clear that it is a `rel=type` header value. Then I have no further objections.

We could perhaps call them "indicated types" or so, and then declare above the table how those indicated types are extracted from `Link`. No strong opinion.

csarven: I used "or specialisation" instead of ldp:BasicContainer figuring that if other container types are supported in the future, they will also follow the slash semantics. If server gets Link rel=type LDP-DC, it will create a container resource with trailing slash.

RubenVerborgh: All fine with me, my issue/misunderstanding was in the phrasing of "`Link` header other than container type."

csarven: How about https://github.com/solid/specification/pull/160/commits/e40d685fcb1ed70484ba3d1e093eecfa5563a0d3 for this draft?

main/resource-access.bs
RubenVerborgh: Perfect, thanks!

main/resource-access.bs
```diff
@@ -130,6 +130,88 @@ Note:
 Servers that wish to disable URI re-use may want to use the `410` status
 code.
 
+### Resource type heuristics ### {#resource-type-heuristics}
+
+Servers MUST apply the following heuristics to determine a request's resource
+type
+[[Source](https://github.com/solid/specification/issues/128#issuecomment-573033297)].
```
justinwb, line 137: Consider explaining under what circumstances the server needs to fall back to this heuristic approach. (i.e. when the client does not specify an interaction model)?

main/resource-access.bs
```diff
@@ -130,6 +130,88 @@ Note:
...
+
+<table id="heuristics-resource-type">
+  <caption>Heuristics to determine a resource type</caption>
+  <thead>
+    <tr>
+      <th>`Slug`</th>
```
dmitrizagidulin, line 143: Can we also include Resource Name (in addition to a Slug) in that column, for creating resources via PUT?

main/resource-access.bs
```diff
@@ -130,6 +130,88 @@ Note:
 Servers that wish to disable URI re-use may want to use the `410` status
 code.
 
+### Resource type heuristics ### {#resource-type-heuristics}
+
+Servers MUST apply the following heuristics to determine a request's resource
+type
+[[Source](https://github.com/solid/specification/issues/128#issuecomment-573033297)].
+
+<table id="heuristics-resource-type">
```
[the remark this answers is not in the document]
csarven: Much better, thanks! https://github.com/solid/specification/pull/160/commits/aadf3a3757dcd76dcf6b5855a6810c72471ca59f

csarven:
[exit rabbit-holes]

I've revisited this issue from the requirements. Some initial considerations:

* Based on slash semantics, a container's URI ends with `/`.

* LDP-BC is the only container interaction required from Solid servers.

* The difference between the initial "`Link` header setting some other LDP interaction model" and the current "`Link` header without a `rel="type"` targeting a valid LDP container type."

* The `Slug` header can introduce conformance variability. The degree of variance and what's acceptable needs to be weighed.

* To what extent shared slashes should be applied to `slugtext`, if at all, partial (eg. excluding nested containers), or in full?

--

Most relevant functional requirements (F1 is borrowed from LDP-UCR):

* F1.1: The system shall provide the ability to create resources within a container.
* F1.2: The system shall provide the ability to create containers for composing resources. 
* F1.3: The system shall provide the ability to create nested containers.
* F2.1: The system shall provide the ability to use client's preferred naming.
* F2.2: The system shall provide the ability to use server's preferred naming.

Non-functional requirement (along the lines of LDP's NF2.3):
* NF1: The system shall support consistent global naming for resource types.

In context of creating resources:

PUT and PATCH can be used to meet F1 and F2.1 but not F2.2.

POST can meet F1 and F2 with different conformance configurations:

* C1: Servers MUST allow creating new resources with a `POST` request to URI path ending `/` [F1].
* C2: Servers MUST create a resource with URI path ending `/{id}` in container `/` [F1.1].
* C3: Servers MUST create a container with URI path ending `/{id}/` in container `/`:
* C3.1: request including `Link` header with `rel="type"` targeting a valid LDP container type [F1.2].
* C3.2.1: request including `Slug` header with `slugtext` using the shared slash semantics [F1.2, F1.3, F2.1].
* C3.2.2: request including `Slug` header with undefined `slugtext` [F1.1, F2.1, F2.2].

<table>
  <caption>Functional Requirements and Conformance Rules</caption>
  <thead>
    <tr>
      <th></th>
      <th>F1.1</th>
      <th>F1.2</th>
      <th>F1.3</th>
      <th>F2.1</th>
      <th>F2.2</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>C1</th><td>✔</td><td>✔</td><td>✔</td><td></td><td></td></tr>
      <th>C2</th><td>✔</td><td></td><td></td><td></td><td></td></tr>
      <th>C3</th><td></td><td>✔</td><td></td><td></td><td></td></tr>
      <th>C3.1</th><td></td><td>✔</td><td></td><td></td><td></td></tr>
      <th>C3.2.1</th><td></td><td>✔</td><td>✔</td><td>✔</td><td></td></tr>
      <th>C3.2.2</th><td></td><td></td><td></td><td>?</td><td>?</td></tr>
    </tr>
  </tbody>
  <tfoot>
    <tr><td colspan="6">
      ✔: Passed
      ?: Cannot tell
    </td></tr>
  </tfoot>
</table>

Notes:

* C1, C2, C3 are based on slash semantics.

* C3.1 is compatible with LDP's interaction model for containers.

* C3.1 using MUST and C3.2.1 using MUST will require servers to do a consistency check.

* C3.2.1 using MUST for `Slug` and MUST for defined `slugtext` will meet F1.3 and F2.1 (but not F2.2)

* C3.2.1 using MAY for `Slug` and MUST for defined `slugtext` would allow a client suggesting the name of a container with `Slug: foo/` to create a container at one server meanwhile a non-container resource at another server. This conformance variance is generally best avoided but does not pose an issue for interoperability.

* C3.2.2 meets F2.1 and F2.2 in a way that server can decide to integrate the suggestion.

---

Status check:

* The original rough consensus in https://github.com/solid/specification/issues/128#issuecomment-573033297 was based on the idea that heuristics are based on all LDP interaction models. The Solid requirement to date is only loosely based on LDP's interaction model for containers, and more specifically, BasicContainer implementation is the only specialisation that's prescribed (re: hierarchical containment and slash semantics). Note for example the important transition from original rough consensus: "`Link` header setting some other LDP interaction model" to the updated text in PR: "`Link` header without a `rel="type"` targeting a valid LDP container type."

* As in the note about C3.2.1 (MAY Slug, MUST defined `slugtext`..), there is an internal inconsistency - variability - in the rough consensus table in that `Slug: foo/` without `Link` creates a container but where `Slug` is not observed or ignored, it will create a resource.

* Taking a step back, there are questions that need more firm answers. Primarily to what extent should the heuristics influenced by slash semantics, and the possibility of `Link` and `Slug` headers being present?

* We can simplify the heuristics or lean towards eliminating it. Some considerations:

  * by not prescribing C3.2.1 in order to meet F1.3 via POST. Or alternatively, C3.2.1 is limited to suggesting a container (as opposed to nested containers - which can potentially introduce security issues because of varying code paths in implementations).

  * if any level of compatibility with LDP interaction models is desired, C3.1 is needed and should at least observe a valid LDP container type (LDP-BC at this time).

  * Looking back at PUT and PATCH and that they can't meet F2.2, we need POST to meet F2.2. However, POST with C3.2.2 may eliminate that possibility and C3.2.1 will prevent F2.2.

  * We should put a pause on the possibility to create resources where effective request URI ends with `/{id}`. It entails that only POST to container `/` is of interest. So, taking on C3.1 (and not any possible LDP interaction model). Consistency check is not required.

My current preference is that we simplify the implementation as much as possible and I think the issues surrounding heuristics introduces a degree of complexity with of course some "benefits" that are not critical to meeting the functional requirements. Note https://github.com/solid/specification/issues/108#issuecomment-549448159 as an example where certain code paths are blocked but the requirements met by other conformance rules.

csarven:
Focusing on F2.2:

`POST` request is required to meet F2.2 for URI path ending `/`.

Note: `POST` request to URI path ending `/{id}` will not address F2.2 because client ends up specifying the name (which is actually F2.1). It is hypothetically possible but this is an awkward way of going at it for `POST` which doesn't fit well with the rest of the setup. It is also neither a common practice or used in existing implementations.

F1.3 will not be prescribed for `POST` as it requires C3.2.1 as MUST and that consequently conflicts with F2.2.

Proposal:
* C1 is used to meet F1.1 and F1.2.
* C2 is used to meet F1.1.
* C3 + C3.1 is used to meet F1.2.

Note: C3.2.2 can be prescribed as MAY but it will be used towards F2.1 instead.

If agreed, I can update the PR and follow up on the related issues. Open to go through alternatives if anyone would like.

justinwb:
> If agreed, I can update the PR and follow up on the related issues. Open to go through alternatives if anyone would like.

👍 from me to proceed with updating PR

csarven:
Updated PR https://github.com/solid/specification/pull/160/commits/c1aac426f859b5c7d5f920822abb56ea56220e32 based on https://github.com/solid/specification/pull/160#issuecomment-636822687 . Will continue to add related criteria in new PRs. Thanks all!

Issue #44, opened 2019-09-19
ldp:constrainedBy recommendations


acoburn:
The LDP specification introduces an IRI to be used to [advertise any constraints](https://www.w3.org/TR/ldp/#h-ldpr-gen-pubclireqs) on the ability of a client to create or update resources:

```http
Link: <https://example.com/SpecificErrorCondition>; rel="http://www.w3.org/ns/ldp#constrainedBy"
```

If Solid published a common vocabulary of error conditions, this feature could potentially be used to accomplish a high level of interoperability across servers. Is such a recommendation within the scope of the Solid specification?

csarven:
Generally agree on making use of `ldp:constrainedBy`.

Before discussing a vocabulary for error details - generally a good idea any way, and I think we have some options there - can we inspect the underlying mechanism and iron out some assumptions:

* If a constraint exists, should it be always advertised (currently MAY in LDP)?
* Who generates a constraint?
  * Server autogenerates based on known set of errors. Also factor in unknowns?
  * Can a client generate a constraint and associate it with a resource? eg. I only want notifications with CC0 license.  In/out of scope?
* Can a constraint be updated? Who can update?
* Which information model(s) / shape languages (eg. SHACL and/or ShEx..) may/must be used to communicate constraints?

csarven:
Related https://github.com/solid/specification/issues/28

justinwb:
@csarven 

> Generally agree on making use of `ldp:constrainedBy`.

Also agree that there's a home for it here 👍 

> Before discussing a vocabulary for error details - generally a good idea any way, and I think we have some options there - can we inspect the underlying mechanism and iron out some assumptions:
> 
> * If a constraint exists, should it be always advertised (currently MAY in LDP)?

I think this depends on the type of constraint - it cannot be generalized. For example - a wac acl is a constraint, but it is only advertised if the acl:agent has acl:Control privileges. Regardless, if the goal is to stick with LDP conventions, the statement in 4.2.1.6 of the [LDP spec](https://www.w3.org/TR/ldp) only requires as a MUST that `ldp:constrainedBy` be used in the response to a request that failed based on imposed constraints.

_LDP servers must publish any constraints on LDP clients’ ability to create or update LDPRs, by adding a Link header with an appropriate context URI, a link relation of http://www.w3.org/ns/ldp#constrainedBy, and a target URI identifying a set of constraints [RFC5988], to all responses to requests that fail due to violation of those constraints_

[...]

Issue #28, opened 2019-08-07
Status messages for client errors


kjetilk:
In NSS5, we implemented more detailed status messages to give clients more feedback. We should bring this along into the spec. 

We have 
- 401 Unauthenticated
- 403 User Unauthorized
- 403 Origin Unauthorized

This was not added in #12

csarven:
Remind me, what's the implementation based on? Anything to do with https://tools.ietf.org/html/rfc7807 or some other standard or existing convention?

RubenVerborgh:
Shall we first distill the problem we are addressing?

Then we can look into various solutions.
I’m not a big fan of the reason phrase, as this is no longer present in HTTP/2, so will get lost (https://tools.ietf.org/html/rfc7540#section-8.1.2.4).

I think we should instead add details to the body.

acoburn:
For machine readability, we could also make use of Link headers with [`rel="http://www.w3.org/ns/ldp#constrainedBy"`](https://www.w3.org/TR/ldp/#h-ldpr-gen-pubclireqs). That could point to an IRI that gives lots of additional information. 

I would also note that, for authN and authZ errors, providing lots of additional data can leak security-related information that really shouldn't be conveyed to a client.

RubenVerborgh:
+1 for the above as a third option.

Regarding security, agreed; this is what we started to convey in https://github.com/solid/specification/pull/13/files#diff-8d72b2ee1631b8eec737030e9b4a3c9cR37-R38 and https://github.com/solid/specification/pull/13/files#diff-8d72b2ee1631b8eec737030e9b4a3c9cR59-R60; we probably want to add more details.

Mitzi-Laszlo:
Which Panel is this conversation associated to?

RubenVerborgh:
None. Not all conversations are panel-bound.

kjetilk:
:+1: to detail errors in RDF

csarven:
We agree on detailed errors in RDF in either response body or link relation (constrainedBy). For interop, we need to have a data model.

https://github.com/solid/specification/issues/112 generalises this issue with server errors: https://github.com/solid/specification/issues/111 .

See also https://github.com/solid/data-interoperability-panel/issues/29 /29 as possible input for the data model.

csarven:
Since the ldp:constrainedBy relation is for client errors, then only the response body in RDF can be used for both client and server errors.

csarven:
[...]

PR #185, opened 2020-06-23
Add constraint discovery for client errors

csarven:
Issue https://github.com/solid/specification/issues/44

### DIFF
--- a/ED/protocol.html
+++ b/ED/protocol.html
@@ -131,7 +131,7 @@ content: "";
     <main>
       <article about="" typeof="schema:Article doap:Specification">
         <h1 property="schema:name">Solid Protocol</h1>
-        <h2>Version <span property="doap:revision">0.9.1</span> Editor’s Draft, 2022-10-09</h2>
+        <h2>Version <span property="doap:revision">0.9.1</span> Editor’s Draft, 2022-10-14</h2>
 
         <details open="">
           <summary>More details about this document</summary>
@@ -178,7 +178,7 @@ content: "";
 
           <dl id="document-modified">
             <dt>Modified</dt>
-            <dd><time content="2022-10-09T00:00:00Z" datatype="xsd:dateTime" datetime="2022-10-09T00:00:00Z" property="schema:dateModified">2022-10-09</time></dd>
+            <dd><time content="2022-10-14T00:00:00Z" datatype="xsd:dateTime" datetime="2022-10-14T00:00:00Z" property="schema:dateModified">2022-10-14</time></dd>
           </dl>
 
           <dl id="document-repository">
@@ -327,6 +327,7 @@ content: "";
                     <li><a href="#writing-resources"><span class="secno">5.3</span> <span class="content">Writing Resources</span></a></li>
                     <li><a href="#deleting-resources"><span class="secno">5.4</span> <span class="content">Deleting Resources</span></a></li>
                     <li><a href="#resource-representations"><span class="secno">5.5</span> <span class="content">Resource Representations</span></a></li>
+                    <li><a href="#constraints-problem-details"><span class="secno">5.6</span> <span class="content">Constraints and Problem Details</span></a></li>
                   </ol>
                 </li>
                 <li class="tocline">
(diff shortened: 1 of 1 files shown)


dmitrizagidulin:
What's the use case (necessity) for this? :)

csarven:
I think it would fall under the UC here: https://www.w3.org/TR/ldp-ucr/#dfn-uc1 -- access guidance, https://www.w3.org/TR/ldp-ucr/#dfn-nf1.1

main/resource-access.bs
```diff
@@ -156,6 +156,25 @@ requests.
 [[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)].
 
 
+## Constraints and Problem Details ## {#constraints-problem-details}
+
+Servers MUST publish any constraints on clients’ ability to create or update
```
dmitrizagidulin, line 161: Are authorization policies in a WAC acl file a constraint?

csarven: No, this excludes the details of authorization policies straight off an ACL. It'd be more like linking to the constraint that for example says that server does not allow client to update containment triples.

justinwb: Assuming that shape validation would fall into this constraint bucket, it might be helpful to use that as a concrete example?

csarven: Do you mean to use the URI of a Shape Validation resource as example? Possibly, but I think I prefer to use a URI that will exist for sure as example, like a URI from the spec. That doesn't rule out the possibility to use server-specific constraints possibly with SV. Just thought that it may be more clear.

I'm still interested in documentation that clarifies the differences between solid:shape and ldp:constrainedBy - beyond https://gitter.im/solid/data-interoperability-panel?at=5e398f03f6945f41ef5ad6a6 and what's currently in the draft spec.

Can we come back to the example after this PR? If an example is desired, we can refer to a requirement in the spec along the lines of:

```
Link: <https://solid.github.io/specification/#http-client>; rel="http://www.w3.org/ns/ldp#constrainedBy"
```

but actually have a URI for the statement and use that in the example:

>A Solid client MUST use the Content-Type HTTP header in PUT, POST and PATCH requests [RFC7231].

csarven: It may be worthwhile to distinguish between the abilities that all clients need to fulfil and the abilities that a (particular) client is expected to fulfil for against a particular resource. Both are important. If solid:shape is already advertising the requirement (or thereof), I'm not sure what the added benefit would be to duplicate that target URI for the ldp:constrainedBy relation. So, from that point, I'd lean on well-defined abilities that a client needs to fulfil to create or update. solid:shape is specific to data whereas ldp:constrainedBy can be for anything (including data).

csarven:
Changed milestone to ~CR for reasons.. https://gitter.im/solid/specification?at=5f291b1ce3160e46baa2a17b

acoburn (approved):

kjetilk:
Since you expressed reservations about this @csarven , and it doesn't seem to be a lot of interest, and it is patching an obsolete file, perhaps we should just close this and return to it once the uses have matured?

csarven:
As said, I'm not opposed to closing the PR or even dropping the feature altogether. However:

Incorporating implementation feedback *is* part of the review process. We could close the issue on the basis - in addition to the reservations - that there isn't implementation feedback but then again we have merged features into the specification with zero or one (public) implementation... or even with <del>poor reviews</del><ins>emoji reactions</ins>. When people have bandwidth or interest to pick up this topic, both the PR and related issues are waiting for them. IMO, this PR now patching an obsolete file is not important for the decision to keeping the PR open or closing it - it is a trivial revision.

---

I suggest to raise this PR/related issues in chats and add to the editors team meeting agenda.

---

Aside reminder https://github.com/solid/specification/issues/44#issuecomment-648048614:

>The target URI of the constrainedBy can be a URI that is part of the Solid spec defining the constraint. The test suite uses those constraints in which the implementation reports are also based on.

kjetilk:
Feel free to raise it in chats. I'm not sure we have the bandwidth to process it, that's all, and then the question is to what extent that it is open represent a distraction. I don't know really.

csarven:
I've re-re-reviewed this PR and related considerations, and then rewrote it. All things considered, it is appropriate to include this as non-normative content. We can revisit the requirement levels for constraints and problem details when the needs/demand is stronger. Alternatively, and depending on how things progress, content along these lines could move to a Best Practices document.

I'd like to draw some attention to continuing related work on structured error messages in RDF: https://github.com/solid/specification/issues/28 . Such work will provide specific/granular information to recipients of the message to help them or the users on their next steps.

VirginiaBalseiro (approved):

Issue #171, opened 2016-01-27
No way to find out if user is authorized to write to a container


rhiaro:
No way to find out if user is authorized to write to (/delete from) a container without actually trying to write (/delete) something.

csarven:
@bblfish Thanks for writing up the Client-Authentication Discovery for this issue. Still relevant today.

As @dmitrizagidulin mentioned, the client may not be able to read the ACL or even be able to discover the location of the ACL. This is reflected in the current state of the specification:

>To discover, read, create, or modify an ACL auxiliary resource, an acl:agent MUST have acl:Control privileges per the ACL inheritance algorithm on the resource directly associated with it.

Even if all clients were to read the ACL, it raises the question of whether the policy they are granted to read is a subset of the information in the ACL. I don't think this flexible or at least requires more from servers - changing the representation based on agent. Alternatively, it is unclear if the link relation should refer to a different ACL resource describing only their readable bits. This duplicates the information and whether the resource is ephemeral or not is unknown.

While Allow includes HTTP methods as value, it is mapped from ACL. The WAC-Allow approach is also based on ACL, and so I presume the cost to calculate per request to be about the same. For an approach based on Link header (using URIs from the ACL vocab) would be in the same category.

Do we have other approaches documented? What implementation experiences can we take into account to make a decision here?

acoburn:
Trellis does not currently implement support for the `WAC-Allow` header (though it would be easy to do so). It does support `Allow`, but the value of the `Allow` response header is a function of the LDP interaction model (e.g. you can POST to a container but not an RDFSource). It would be an interesting idea to combine those two concepts, though, and filter the list of Allow headers by the operations that are supported. The downside of that approach, though, is that it wouldn't distinguish between the current user and a "public" user, which is what WAC-Allow does.

OTOH, one significant operational downside of the WAC-Allow header (and why Trellis doesn't support it) is that ACL lookups are expensive and WAC-Allow requires two separate lookups for each request: one for the current user and one for an anonymous user. (TBH, I'm not entirely convinced of the necessity of the extra lookup for an anonymous user)

csarven:
Proposal/requirements based on the `WAC-Allow` HTTP header is merged via PR https://github.com/solid/specification/pull/210 . See further details/discussion on concerns of this header, as well as use cases for different permission groups. Implementation feedback is always useful/welcome. Please follow up with new issues/proposals.

Issue #145, opened 2019-06-23
Which permissions are required for deleting an ACL document?


RubenVerborgh:
When deleting an ACL document R in a container C, do we require:
- Write and Control permissions on R?
- Write permissions on C?
- Both?
- Something else?

See solid/solid-spec#195.

kjetilk:
I transferred this issue, since it is a dependency for #41 , in which the rough consensus is indeed as @michielbdejong 's notes: 

When deleting an ACL document R about a target T, we require Control on T.

Then, it is also understood that

When deleting an document R, we require Write on R. If R has a ACL Document, it will be deleted as a side-effect.

csarven:
The premises that I've stated earlier was tied to common understanding or rough consensus based on available data. As this issue re-questioned the actual required permissions, my conclusion here:

>Taking this as a whole, Write on resource or associated R is the only requirement to delete R.

was based on the minimum requirement by building up from the premises. I think it turned out to be Write. I have no objection to requiring Control on R in order to delete the associated ACL. It was just what was previously spec'd or understood wasn't clear.

---

>Also note that ACL documents aren't independently contained, so the resource R may be a container.

Are you saying that when an ACL is created (probably via PUT or PATCH), it will not be added to containment? If so, where did you derive that from?

The following are equivalent:

> When deleting an resource R, we require Write on parent container C.

>P: If R can occur in containment triples, then Write permission on C is required ( if based on solid/solid-spec#195 (comment) )

But note:

>If R can't occur in containment triples (see also #116 ) then Write permission on C is not required.

I think it makes sense to exclude resources like ACL in containment but I don't think we have fully worked that out (as to why or why not).

---

The following two statements are essentially describing the same thing (note that I've used 'R' as that was the initial term used in this issue):

>If R has an ACL Document A, it will be deleted as a side-effect.

>P: Deleting a resource removes associated R (not including inherited R).

This may however raise an issue in that other resources may depend on the same ACL. So, what's currently not specified is the shape of the ACL - IIRC, this is looked into in the interop panel - in that all authorization policies is for the same target resource. Or at the very least acknowledging that an ACL of a resource can only be referred by that resource (rel=acl). That will ensure that once a resource is deleted and its ACL with it (provided eg. Control on the resource, and Write on container it is in), then there is no issue about other resources suddenly losing their ACL. Perhaps this warrants a separate issue (unless one exists that escapes me) in that resources should have a dedicated ACL that's not shared by other resources.

[...]

Issue #132, opened 2019-12-05
Recursive deletion of members on deletion of container


fabiancook:
Opening this issue to start discussion on moving forward with recursive deletion of members within a containers. Related to https://github.com/solid/solid-spec/issues/172 

I propose that we allow this from a specification level, this will allow for agents to remove a container without first reading & then deleting all its members. 

An example of when this would be useful is seen when an agent wants to no longer utilise an application that had previously been given write access to a specific folder within the agents storage pod, the user may not have rights to all directories within that container, but its still their storage pod to do whatever they please with. If we allow recursive deletion, it means that the agent doesn't need to have read or write access to those member resources, but can still delete the entire folder.

kjetilk:
As part of #41 , we should continue discussion here.

The way that I see this, there are several analog functions we need to consider:

1. `DELETE` on container deletes all the resources in the container, but does not recurse, i.e. `rm *`
1. `DELETE` on container recurses into the container hierarchy and deletes resources if it has `acl:Write` on the resources in the container hierarchy. Corresponding to `rm -r`.
1. `DELETE` on container recurses into the container hierarchy, and deletes as above, but deletes also if it has `acl:Control`, since `acl:Control` means it could have changed the permission if it wanted to. This would roughly correspond to `rm -rf`.

To me, the attraction of recursive deletion boils down to two aspects:
1. It saves work, since the client doesn't have to recurse.
1. It can be done atomic

Out of these, I think the second is actually the most important. It can be hard to do at scale, but I see it as even harder for the client to do it. The possibility of atomic deletes is important.

However, recursive deletion is also possibly a very destructive operation. We should have a mechanism so that the user can ensure that it doesn't happen by accident. It could prompt email confirmation, reconfirmation of credentials, UIs that pop up with a message or something (I thought I had an issue open for that, but couldn't find it now).

csarven:
I generally prefer `rm -rf` as the default for container but as we've postponed the inner details of that for later.. I'll mention the following as an alternative mid-step:

After container delete (with `rm *`) the response message ( eg. similar to error messages https://github.com/solid/specification/issues/28 ) can indicate what happened or what else can be done eg. `rm -rf` with a particular request header/value. So, the decision on what's default for delete can be mitigated with a header.

[I think dealing with notifications for this use case may be a bit too heavy duty. In what application or operating system do we ever get "notified" about attempts to delete a container? I can understand the case for notifications when deleting an account (or WebID or anything of high significance) but not for general use.]

[...]

Issue #72, opened 2019-09-25
Deleted account tombstone


kjetilk:
To complete https://github.com/solid/solid/issues/14, we need to record what to do with incoming requests after deletion. At the very least, we should make sure that nobody gets the same URI so that they can impersonate the former account owner.

As noted there, we should respond with `410` if nothing else has been recorded, but even better is it to allow users to register their new home and do a `301`. At the very least, their new WebID should be saved.

I thought about using the old [`.htaccess`](https://httpd.apache.org/docs/current/howto/htaccess.html) format for this, that we just dump such a file in the former home dir of the user.

csarven:
This issue should crosscheck with https://github.com/solid/specification/issues/41 , https://github.com/solid/specification/issues/61

kevin-inrupt:
It might be good to have an override for this functionality to allow the WebID can be reused. If a Pod get corrupted, sometimes the easiest approach to fixing it is to delete the Pod and immediately create a new one using the same WebID. Maybe have a "Pod reset" option?

csarven:
A resource server doesn't have knowledge of accounts eg. a WebID URL is an RDF source like all other RDF sources. Whether a deleted account is associated with a WebID can be reusable or not is something should be up to an identity provider. We don't have an interface to help indicate that a deleted account's resources should return a particular status message or a provenance record. If that becomes available, when requesting to access a deleted WebID URL, implementations can potentially mirror identity provider's intention, promise, or what's possible eg. 4xx, 3xx, or something else. I suggest to leave the specific status code or marking an account as "tombstone" to implementations to their discretion - same decision made for all resources: https://github.com/solid/specification/issues/103 ie. 410 is optional (falling back on RFC 7231). In that view, by default WebIDs can reused provided that URI persistence is taken into account.

TallTed:
I think it's worth noting on this topic that existing social networking sites tend to have "sunset", "grace", and/or "purge" terms for terminated/"deleted" accounts.

For a period of time (typically, a year or so), actions trying to interact with the terminated account receive an appropriate error message, and no-one is allowed to assume that account's identifier (typically, a simple username). Some sites also preserve the terminating user's data for this defined period. Some sites allow the original user to revive the account and, if preserved, restore that data, during the same period.

At the end of such a grace period, any preserved data is generally purged.  Typically at the same time, but sometimes after an additional waiting period, the username is "freed", and another user can adopt it.

[...]

PR #187, opened 2020-06-29
DELETE support and semantics

csarven:
In response to the following (and possibly other) issues on how the delete operation works in context of Web Access Control:

* https://github.com/solid/specification/issues/37
* https://github.com/solid/specification/issues/39
* https://github.com/solid/specification/issues/41
* https://github.com/solid/specification/issues/46
* https://github.com/solid/specification/issues/72
* https://github.com/solid/specification/issues/145

### DIFF
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -85,6 +85,10 @@ Server MAY omit the `DELETE` method in the HTTP response header `Allow` in respo
 
 The server might perform additional actions, as described in the normative references like [[!RFC7231]]. For example, the server could remove membership triples referring to the deleted resource, perform additional cleanup tasks for resources it knows are no longer referenced or have not been accessed for some period of time, and so on.
 
+Subsequent `GET` requests to the deleted resource usually results in a `404` or `410` status code, although HTTP allows others.
+[[Source](https://github.com/solid/specification/issues/72)]
+[[Source](https://github.com/solid/specification/issues/46)]
+
 ### Required client-side implementation ### {#http-client}
 
 A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -57,6 +57,9 @@ Servers MUST support the HTTP `DELETE` method [[!RFC7231]].
 
 When using Web Access Control ([[#wac]]):
 
+Servers MUST NOT allow delete requests to a storage's root container and associated ACL resource and exclude the `DELETE` method in the HTTP response header `Allow` in response to safe method requests [[!RFC7231]].
+[[Source](https://github.com/solid/specification/issues/37#issuecomment-627281466)]
+
 ### Required client-side implementation ### {#http-client}
 
 A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -66,6 +66,12 @@ To delete a resource, an `acl:agent` MUST have `acl:Write` privilege per the ACL
 When a contained resource is deleted, the server MUST also remove the corresponding containment triple, which has the effect of removing the deleted resource from the containing container.
 [[Source](https://www.w3.org/TR/ldp#ldpc-del-contremovesconttriple)]
 
+When a contained resource is deleted, the server MUST also delete the associated resources (see the [[#rm]] section).
+
+To delete an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST have `acl:Control` privileges per the ACL inheritance algorithm on the resource directly associated with it.
+[[Source](https://github.com/solid/specification/issues/145)]
+[[Source](https://github.com/solid/specification/issues/41)]
+
 ### Required client-side implementation ### {#http-client}
 
 A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -79,6 +79,12 @@ A container can only be deleted if it contains no resources ([[#resource-contain
 
 Server MAY omit the `DELETE` method in the HTTP response header `Allow` in response to safe method requests to a container, if `acl:agent` does not have the `acl:Write` privilege to a resource under its URI path.
 
+[[Source](https://github.com/solid/specification/issues/41)]
+
+<em>This section is non-normative.</em>
+
+The server might perform additional actions, as described in the normative references like [[!RFC7231]]. For example, the server could remove membership triples referring to the deleted resource, perform additional cleanup tasks for resources it knows are no longer referenced or have not been accessed for some period of time, and so on.
+
 ### Required client-side implementation ### {#http-client}
 
 A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -57,36 +57,54 @@ Servers MUST support the HTTP `DELETE` method [[!RFC7231]].
 
 When using Web Access Control ([[#wac]]):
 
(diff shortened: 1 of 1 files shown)


main/resource-access.bs
```diff
@@ -50,6 +50,51 @@ A Solid server MUST reject `PUT`, `POST` and `PATCH` requests without the
...
+[[Source](https://www.w3.org/TR/ldp#ldpc-del-contremovesconttriple)]
+
+When a contained resource is deleted, the server MUST also delete the associated resources (see the [[#rm]] section).
+
+To delete an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST have `acl:Control` privileges per the ACL inheritance algorithm on the resource directly associated with it.
+[[Source](https://github.com/solid/specification/issues/145)]
+[[Source](https://github.com/solid/specification/issues/41)]
+
+A container can only be deleted if it contains no resources ([[#resource-containment]]). When a `DELETE` method request is made to a container, the server MUST delete the contained resources except containers. To recursively delete a container, server MUST accept client's `DELETE` request including the HTTP `Prefer` header with `return="representation"; include="http://www.w3.org/ns/ldp#PreferContainment"` (see [[!LDP]]'s Preferences on the Prefer Request Header). Server MUST respond with the `409` status code and response body including containment triples about unaffected resources of the delete request for agents with `acl:Read` privilege per the ACL inheritance algorithm.
```
acoburn, line 75: This is confusing. First, this sentence states that a container _can only be deleted if it contains no resources_. Then it describes recursive delete, which contradicts the first statement.

As a general comment, recursive delete becomes _very_ complicated in the case of complex ACLs, atomicity guarantees and the prospect of partial failure. If you plan to support recursive delete in the specification, I would suggest that there be a mechanism for reporting partial failures to clients (or, as would be my preference, don't support recursive delete at all)

acoburn: Also, the `Prefer: return=representation` part is strange. That hint (`return=representation`) is typically used to control the _response_ of the HTTP operation, not the internal server behavior.

pmcb55: Well, I assume the first sentence could still hold if you implement recursive delete via a depth-first search - i.e. the server starts deleting the leaf-nodes, and then only deletes child containers once they are empty, and keep popping back up the stack.
Partial failures do still make be nervous though - i.e. having some mechanism to report them back to the user, and making it very clear to client-app developers to **_not_** assume a recursive delete **_will just happen_** on the server (i.e. where they update their user interface to remove an entire folder structure). Instead they need to be aware that the delete operation might partially fail (and if it does fail for one resource, does it stop the recursive delete operation at that point, or does it continue deleting remaining leaf-nodes and remaining empty containers?). And so those developers need to know that they need to wait for a server response (or an async notification) telling them what actually happened on the server before updating their UI.

csarven: I see what you mean with the confusion. It was intended to cover the DELETE request to a container. Perhaps it is redundant in that the third sentence mentions recursion - with the assumption that includes the effective request URI (ie. the container itself), in addition to the contained resources. If we omit the first sentence, it is clear?

csarven: True re: `Prefer: return=representation` and typical use so far. I figured that Prefer header in context of DELETE would be sufficient. It still has the client preferring to see the resulting representation of the container. This is where the 409 also comes into play ie. if there is a partial fail, client can know (provided that they have Read).

There wasn't a particular way where a client can request its preference to delete a container recursively - at least in the issues that I've looked into. Happy to look at alternative approaches.

csarven: The rough consensus was that the base DELETE would only remove the non-container resources, and to revisit recursive deletion. Prefer may be one way to do this.

Does Trellis support recursive delete? How would Trellis respond to `DELETE /foo/` , where `/foo/` contains `/foo/bar/` and containing `/foo/bar/baz`?

acoburn: No, Trellis does not support recursive delete. Let me emphasize that recursive delete is _hugely_ problematic.

Imagine the case of a container with 100,000 child resources. What happens if the HTTP connection times out before the delete operation finishes? Or if a user clicks a "stop" button in the browser?

Right now, Trellis would delete the container and leave an orphan nodes at /foo/bar/. I have some code, however, that would case the request to fail with a 409 error, preventing clients from deleting non-empty containers.

csarven: I have considered the `Expect` header but decided against it. Would reusing LDP's Prefer pattern be completely wrong in context of DELETE?

csarven: I would expect servers would have a approximation or limit the time in which they have to generate a response. We don't currently have a way (not to forget that there was no corresponding use case) for a client to send an additional request to stop the ongoing delete.

We can't have orphaned nodes in Solid ( https://github.com/solid/specification/blob/88baa6eb7bbd937f524150109bb8925ab39f619a/main/resource-access.bs#L103-L108 ).

csarven: @pmcb55 Indeed eg `DELETE /foo/` would remove `/foo/` if it has no contained resources. If `/foo/` contains resources, `DELETE /foo/` would remove them instead. I don't particularly like how the operation is different depending on what the target resource contains, but I do see its appeal. In the PR's wording, as mentioned earlier, the first sentence probably goes better with the third sentence; recursive delete also removing the container itself (effective request URI).

csarven: While server accepts the request for delete recursively, it could still respond with a 409 or another 4xx. Alternatively, instead of MUST, we use MAY (irrespective of the mechanism used to indicate client's wish for recursive delete).

acoburn: > DELETE /foo/ would remove /foo/ if it has no contained resources. If /foo/ contains resources, DELETE /foo/ would remove them instead.

FWIW, I would be strongly 👎  on this.

Imagine the case where you have two clients writing to `/foo/`. The first checks `/foo/` and finds that it is empty, it then issues `DELETE /foo/` expecting that it has deleted `/foo/`, but in fact it has deleted `/foo/bar` and `/foo/baz` (which were created by the second client the intervening time).

That scenario would become extremely difficult for clients to reason about.

acoburn:
> I have considered the Expect header but decided against it.

The `Expect` header is used for `100-continue` requests only. Its original use as a "must-understand" extension mechanism has been removed from [RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.1.1)

```
      Note: The Expect header field was added after the original
      publication of HTTP/1.1 [RFC2068] as both the means to request an
      interim 100 (Continue) response and the general mechanism for
      indicating must-understand extensions.  However, the extension
      mechanism has not been used by clients and the must-understand
      requirements have not been implemented by many servers, rendering
      the extension mechanism useless.  This specification has removed
      the extension mechanism in order to simplify the definition and
      processing of 100-continue. 
```

main/resource-access.bs
acoburn: > Would reusing LDP's Prefer pattern be completely wrong in context of DELETE?

The use of `return=representation` is incorrect. The use of Prefer is not incorrect.

csarven: Nods.. 

Proposal: postpone if and how recursive deletion can work to another PR. For now, the base requirement needs to be *along these lines*:

When a `DELETE` method request is made to a container, the server MUST delete the container if it contains no resources. If the container contains resources, the server MUST respond with the `409` status code and response body describing the error.

When we need to, we should add recursive delete in a way without interfering with those requirements.

timbl: Happy to see a version with no MAY and no "prefer".  This spec has to be tight and unambiguous.  Any time the word MAY appeared is a red flag!  And using anything like "prefer" to actually have hard must semantics sounded a bad idea.

csarven: Updated in https://github.com/solid/specification/pull/187/commits/c11eab495839bdaab5f51e06bc153c1260d51611

csarven:
Closing this PR in favour of https://github.com/solid/specification/pull/188 (including the commits from here; feature/delete). Please continue with the reviews there.

Issue #117, opened 2019-11-14
405 for "other" methods?


kjetilk:
I just created a test with the following request:

```turtle
:test_bob_r_o_req a http:RequestMessage ;
    http:method "DAHU" ;
    http:requestURI </test-auth/alice_share_bob_r.txt> .
```

NSS returns `400` while I expected `405`.

I think it is good practice to return a `405` for methods you do not implement, since it tends to be trivial to do and affords the client with more details to adjust its interactions. The [RFC](https://tools.ietf.org/html/rfc7231#section-6.5.5) is more relaxed on this topic than I tend to be though. The question is if we should say something normative about it, and if so, should that normative statement be a `MUST`?

csarven:
For that test, `405` is precise whereas `400` would be vague. I find `405` to be more useful for clients and don't particularly see why a server would choose to return a `400` especially if the issue is strictly about the method not being available on a given resource. There is nothing wrong, faulty, unsound with the request itself. If there are other dimensions that's taken into account, `400` may be more meaningful but that'd be a different test.

kjetilk:
Good, so we agree! 

Not a big issue, I'll set it on the June milestone.

Issue #85, opened 2019-10-03
Specify PATCH


csarven:

dmitrizagidulin:
Reminder to discuss: What to do about the `MS-Author-Via: SPARQL` header? (see PR https://github.com/solid/node-solid-server/pull/1313 for example).
Does it need to be added to the spec (I bring it up here because it's PATC-related)? As a MUST or a MAY?

kjetilk:
Just wondering if my "RDF Merge payload into document" might make be less problematic with `PATCH`... I just think it really needs to be a feature, the intuition for something like that to exist is very strong in me... :-)

kjetilk:
Just wanting to post this somewhere as a note to self, since we need to specify this behaviour, which is mostly correct:

```
        # HTTP/1.1 415 Unsupported Media Type
        # Connection: close
        # Date: Fri, 15 Nov 2019 00:12:28 GMT
        # ETag: W/"2b-rKTuKHccg1SjlLhItRjeKZPSFY4"
        # Vary: Accept, Authorization, Origin
        # Allow: OPTIONS, HEAD, GET, PATCH, POST, PUT, DELETE
        # Content-Length: 43
        # Content-Type: text/plain; charset=utf-8
        # Access-Control-Allow-Credentials: true
        # Access-Control-Expose-Headers: Authorization, User, Location, Link, Vary, Last-Modified, ETag, Accept-Patch, Accept-Post, Updates-Via, Allow, WAC-Allow, Content-Length, WWW-Authenticate, MS-Author-Via
        # Client-Date: Fri, 15 Nov 2019 00:12:28 GMT
        # Client-Peer: 172.22.8.2:8446
        # Client-Response-Num: 1
        # Client-SSL-Cert-Issuer: /CN=*.robin.kjernsmo.net
        # Client-SSL-Cert-Subject: /CN=*.robin.kjernsmo.net
        # Client-SSL-Cipher: TLS_AES_256_GCM_SHA384
        # Client-SSL-Socket-Class: Net::SSL
        # Client-SSL-Warning: Peer certificate not verified
        # Link: <alice_share_bob.txt.acl>; rel="acl", <alice_share_bob.txt.meta>; rel="describedBy", <http://www.w3.org/ns/ldp#Resource>; rel="type"
        # MS-Author-Via: SPARQL
        # Set-Cookie: connect.sid=[...] Path=/; Expires=Sat, 16 Nov 2019 00:12:28 GMT; HttpOnly; Secure
[...]

Issue #68, opened 2019-09-24
Clarify recursive container creation on POST, PATCH and PUT


dmitrizagidulin:
As was mentioned in the [semantics of PUT discussion](https://github.com/solid/specification/issues/40#issuecomment-534596878), we need to decide on a consistent policy for recursive container creation for various resource types (Resources and Containers) and via the 3 major creation verbs (PUT, PATCH and POST).

(Also sometimes referred to as `mkdir -p` semantics)

Current behavior (in Node Solid Server):

* PATCH - creating a resource: Intermediate containers are automatically created.
* PATCH - creating a container: Yes. Technically this gets parsed as a PATCH to a container's `.meta`
* POST - creating a resource: Does not create intermediate container, throws a `404 Not Found`.
* POST - creating a container: same as above.
* PUT - creating a resource: Intermediate containers are automatically crated (`mkdir -p` style).
* PUT - creating a container: Not currently supported, but [hopefully will be](https://github.com/solid/specification/issues/40)

acoburn:
Some clarifying questions:

- By "Current behavior", I assume Node Solid Server is meant. That may not apply to other implementations
- Is it possible to create resources via PATCH? LDP does not explicitly take a position on this, though all examples show PATCH as a mechanism for updating an existing resource. [RFC 5789](https://tools.ietf.org/html/rfc5789#section-2) states that PATCH `MAY` be used to create resources. See also related issue: #42 
- With `POST`, are you referring to a case where the target URL of the request does not exist? Wouldn't a 404 be returned in that case?
- Or is the interaction that a client `POST`s to an extant container but that request causes the creation of a new container hierarchy (e.g. with a `Slug` header that includes slashes?) And what if an implementation rejects such requests?

I would note that a RESTful design would tend to avoid side affects: i.e. a `PUT` to `/foo/bar/baz` that simultaneously creates `/foo`, `/foo/bar` and `/foo/bar/baz` might be considered problematic.

dmitrizagidulin:
@acoburn Yeah, I just meant 'current NSS behavior' (I edited the comment to clarify).

> Is it possible to create resources via PATCH?

In nss, yes, and I believe a lot of Tim's Data Browser / mashlib code relies on this functionality. As in, this is not incidental, but definitely part of the spec.

> With POST, are you referring to a case where the target URL of the request does not exist?

Right, ok, I just double checked, and NSS currently does return a 404 / target url does not exist.

[...]

Issue #60, opened 2019-09-20
Implementation guidance on ETags


acoburn:
The LDP specification [requires the use of ETag headers](https://www.w3.org/TR/ldp/#ldpr-gen-etags) for `GET` and `HEAD` responses. There is some subtlety to how ETags work in the context of RDF, and some implementation guidance (e.g. in a non-normative section) might be useful.

For example, [RFC 7232, section 2.3](https://tools.ietf.org/html/rfc7232#section-2.3) is clear that different ETags should be produced for different representations of a resource: 

> An entity-tag is an opaque validator for differentiating between multiple representations of the same resource, regardless of whether those multiple representations are due to resource state changes over time, content negotiation resulting in multiple representations being valid at the same time, or both.

Furthermore, [RFC 7232, section 2.1](https://tools.ietf.org/html/rfc7232#section-2.1) describes the difference between weak and strong ETags. For RDF serializations where RDF semantics may be more important than byte-for-byte consistency, is it legitimate to generate a strong ETag for an LDP-RS even if the server does not guarantee the order of the triples?

Consider also the case where a server can produce `text/turtle`, `application/ld+json` and `application/n-triples`, each of which generates a different ETag value. Then consider also that the server supports `Prefer` headers as well as `Content-Encoding` negotiation, each of which could produce an additional dimension for ETag generation. In that case, suppose a client wishes to send a `PATCH` request along with an `If-Match` header as part of a conditional request. What value should be included in the `If-Match` header? And given the various permutations of ETags that could be generated, does a server need to check all such permutations before accepting the conditional request? (Does `If-Match` make sense in the context of `PATCH`?)

This question is simpler in the context of `PUT`, but what if a client retrieves an LDP-RS as JSON-LD using a custom profile (`Accept: application/ld+json; profile="https://...."`), modifies the RDF graph and then, via `PUT` replaces the resource as JSON-LD. What value should be used with `If-Match`? Does this change if the resource is retrieved as JSON-LD and replaced as Turtle?

There are clearly some nuances here, and it may be helpful to provide some guidance to implementers.

kjetilk:
Interesting topic!

@acoburn :
> For RDF serializations where RDF semantics may be more important than byte-for-byte consistency, is it legitimate to generate a strong ETag for an LDP-RS even if the server does not guarantee the order of the triples?

It does seem to be that it would be legitimate to do so, from the spec:
[...]

PR #188, opened 2020-07-08
Reading and Writing resources

csarven:
In response to the following (and possibly other) issues on how read, append, write and delete operations work in context of Web Access Control:

* https://github.com/solid/specification/issues/33
* https://github.com/solid/specification/issues/37
* https://github.com/solid/specification/issues/39
* https://github.com/solid/specification/issues/39
* https://github.com/solid/specification/issues/40
* https://github.com/solid/specification/issues/41
* https://github.com/solid/specification/issues/42
* https://github.com/solid/specification/issues/43
* https://github.com/solid/specification/issues/46
* https://github.com/solid/specification/issues/60
* https://github.com/solid/specification/issues/68
* https://github.com/solid/specification/issues/72
* https://github.com/solid/specification/issues/85
* https://github.com/solid/specification/issues/108
* https://github.com/solid/specification/issues/117
* https://github.com/solid/specification/issues/145
* https://github.com/solid/specification/pull/160

### DIFF
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -258,7 +258,7 @@ To delete an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST have
 directly associated with it.
 [[Source](https://github.com/solid/specification/issues/145)]
 
-When a `DELETE` method request is made to a container, the server MUST delete
+When a `DELETE` request is made to a container, the server MUST delete
 the container if it contains no resources. If the container contains
 resources, the server MUST respond with the `409` status code and response
 body describing the error.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -85,6 +85,10 @@ Server MAY omit the `DELETE` method in the HTTP response header `Allow` in respo
 
 The server might perform additional actions, as described in the normative references like [[!RFC7231]]. For example, the server could remove membership triples referring to the deleted resource, perform additional cleanup tasks for resources it knows are no longer referenced or have not been accessed for some period of time, and so on.
 
+Subsequent `GET` requests to the deleted resource usually results in a `404` or `410` status code, although HTTP allows others.
+[[Source](https://github.com/solid/specification/issues/72)]
+[[Source](https://github.com/solid/specification/issues/46)]
+
 ### Required client-side implementation ### {#http-client}
 
 A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -161,8 +161,8 @@ response of a `OPTIONS *` request.
 
 ### Writing resources ### {#write}
 
-Servers MUST support the HTTP `PUT`, `POST` and `PATCH` methods [[!RFC7231]]
-for clients to create and update resources.
+When a server supports the HTTP `PUT`, `POST` and `PATCH` methods [[!RFC7231]]
+this specification imposes the following requirements:
 [[Source](https://github.com/solid/specification/issues/39#issuecomment-538017667)]
 
 Servers MUST create intermediate containers and include corresponding
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -223,7 +223,8 @@ resource directly associated with it.
 
 ### Deleting Resources ### {#delete}
 
-Servers MUST support the HTTP `DELETE` method [[!RFC7231]].
+When a server supports the HTTP `DELETE` method [[!RFC7231]] this
+specification imposes the following requirements:
 [[Source](https://github.com/solid/specification/issues/39#issuecomment-538017667)]
 
 When a `DELETE` request targets storage's root container or its associated ACL
@@ -231,10 +232,6 @@ resource, the server must respond with the `405` status code. Server MUST
 exclude the `DELETE` method in the HTTP response header `Allow` in response to
 safe method requests [[!RFC7231]].
 [[Source](https://github.com/solid/specification/issues/37#issuecomment-627281466)]
-
-To delete a resource, an `acl:agent` MUST have `acl:Write` privilege per the
-ACL inheritance algorithm on the resource and the containing container.
-[[Source](https://github.com/solid/solid-spec/issues/195#issuecomment-559799154)]
 
 When a contained resource is deleted, the server MUST also remove the
 corresponding containment triple, which has the effect of removing the deleted
@@ -244,16 +241,22 @@ resource from the containing container.
 When a contained resource is deleted, the server MUST also delete the
 associated resources (see the [[#rm]] section).
 
-To delete an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST have
-`acl:Control` privileges per the ACL inheritance algorithm on the resource
-directly associated with it.
-[[Source](https://github.com/solid/specification/issues/145)]
-
 When a `DELETE` request is made to a container, the server MUST delete
 the container if it contains no resources. If the container contains
 resources, the server MUST respond with the `409` status code and response
 body describing the error.
 [[Source](https://github.com/solid/specification/pull/187/files/b7426e95a1613e08195a853a4d0a403b7030f494#r447130915)]
+
+When using Web Access Control ([[#wac]]):
+
+To delete a resource, an `acl:agent` MUST have `acl:Write` privilege per the
+ACL inheritance algorithm on the resource and the containing container.
+[[Source](https://github.com/solid/solid-spec/issues/195#issuecomment-559799154)]
+
+To delete an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST have
+`acl:Control` privileges per the ACL inheritance algorithm on the resource
+directly associated with it.
+[[Source](https://github.com/solid/specification/issues/145)]
 
 <em>This section is non-normative.</em>
 
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -272,7 +272,7 @@ triples referring to the deleted resource, perform additional cleanup tasks
 for resources it knows are no longer referenced or have not been accessed for
 some period of time, and so on.
 
-Subsequent `GET` requests to the deleted resource usually results in a `404`
+Subsequent `GET` requests to the deleted resource usually result in a `404`
 or `410` status code, although HTTP allows others.
 [[Source](https://github.com/solid/specification/issues/72)]
 [[Source](https://github.com/solid/specification/issues/46)]
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -183,10 +183,8 @@ subsequent requests to the newly created container's URI as if it is a valid
 LDP container type by including the HTTP response's `Link` header.
 [[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)]
 
-For resource representations having an algorithm to perform an append
-operation, servers MUST update resources for HTTP `POST` requests. Servers
-MUST perform RDF merge [[!RDF11-MT]] on RDF sources [[!RDF-CONCEPTS]].
-[[Source](https://github.com/solid/specification/issues/108#issuecomment-573655070)]
+When a `POST` method request targets a non-container resource, the server
+`MUST` respond with the `405` status code.
 
 When a `PUT` or `PATCH` method request targets an auxiliary resource, the
 server MUST create or update it. When a `POST` method request with the `Slug`
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -221,12 +221,6 @@ have `acl:Control` privileges per the ACL inheritance algorithm on the
 resource directly associated with it.
 [[Source](https://github.com/solid/specification/issues/42#issuecomment-616688848)]
 
-<em>This section is non-normative.</em>
-
-While it is technically possible to auxiliary resources using the HTTP `POST`
-method with the `Slug` header, it allows varying implementations due to its
-non-deterministic resource naming.
-
 ### Deleting Resources ### {#delete}
 
 Servers MUST support the HTTP `DELETE` method [[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -210,6 +210,10 @@ believes that the resource does not have a current representation.
 [[Source](https://github.com/solid/specification/issues/108#issuecomment-567272797)]
 [[Source](https://github.com/solid/specification/issues/40#issuecomment-566995240)]
 
+Servers MAY use the HTTP `ETag` header with a strong validator for RDF bearing
+representations in order to encourage clients to opt-in to using the
+`If-Match` header in their requests.
+
 When using Web Access Control ([[#wac]]):
 
 To create or update an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -177,6 +177,12 @@ container's containment triples; if the server receives such a request, it
 SHOULD respond with a `409` status code.
 [[Source](https://github.com/solid/specification/issues/40#issuecomment-573358652)]
 
+Clients MUST use the HTTP `If-None-Match` header with a value of `"*"` to
+prevent an unsafe method from updating an existing representation of the
+target resource when creation of a new resource is intended.
+[[Source](https://github.com/solid/specification/issues/108#issuecomment-567272797)]
+[[Source](https://github.com/solid/specification/issues/40#issuecomment-566995240)]
+
 ### Deleting Resources ### {#delete}
 
 Servers MUST support the HTTP `DELETE` method [[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -151,11 +151,9 @@ When using Web Access Control ([[#wac]]):
 
 Servers MUST indicate their support for HTTP Methods by responding to HTTP
 `GET`, `HEAD` and `OPTIONS` requests on the resource's URL with the HTTP
-Method tokens in the HTTP response header `Allow`.
-
-Servers MUST indicate supported media types in the `Accept-Patch` [[!RFC5789]]
-and the `Accept-Post` [[!LDP]] headers for matching HTTP methods in `Allow`
-header value.
+Method tokens in the HTTP response header `Allow`. Servers MUST indicate
+supported media types in the `Accept-Patch` [[!RFC5789]] and the `Accept-Post`
+[[!LDP]] headers for matching HTTP methods in `Allow` header value.
 
 Servers MAY include the HTTP `Accept-Patch` and `Accept-Post` headers in the
 response of a `OPTIONS *` request.
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -125,6 +125,8 @@ Note:
 Servers that wish to disable URI re-use may want to use the `410` status
 code.
 
+## Reading and Writing Resources ## {#read-write}
+
 ### Resource type heuristics ### {#resource-type-heuristics}
 
 When creating new resources, servers can determine an effective request URI's
@@ -138,8 +140,6 @@ Clients who want the server to determine the URI for their resource, MUST use
 the POST request.
 
 [[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)].
-
-## Reading and Writing Resources ## {#read-write}
 
 ### Writing resources ### {#write}
 
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -57,6 +57,9 @@ Servers MUST support the HTTP `DELETE` method [[!RFC7231]].
 
 When using Web Access Control ([[#wac]]):
 
+Servers MUST NOT allow delete requests to a storage's root container and associated ACL resource and exclude the `DELETE` method in the HTTP response header `Allow` in response to safe method requests [[!RFC7231]].
+[[Source](https://github.com/solid/specification/issues/37#issuecomment-627281466)]
+
 ### Required client-side implementation ### {#http-client}
 
 A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -183,8 +183,9 @@ subsequent requests to the newly created container's URI as if it is a valid
 LDP container type by including the HTTP response's `Link` header.
 [[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)]
 
-When a `POST` method request targets a non-container resource, the server
-`MUST` respond with the `405` status code.
+When a `POST` method request targets a non-container resource without an
+existing representation, the server `MUST` respond with the `404` status code.
+[[Source](https://github.com/solid/specification/issues/108#issuecomment-549448159)]
 
 When a `PUT` or `PATCH` method request targets an auxiliary resource, the
 server MUST create or update it. When a `POST` method request with the `Slug`
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -66,6 +66,12 @@ To delete a resource, an `acl:agent` MUST have `acl:Write` privilege per the ACL
 When a contained resource is deleted, the server MUST also remove the corresponding containment triple, which has the effect of removing the deleted resource from the containing container.
 [[Source](https://www.w3.org/TR/ldp#ldpc-del-contremovesconttriple)]
 
+When a contained resource is deleted, the server MUST also delete the associated resources (see the [[#rm]] section).
+
+To delete an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST have `acl:Control` privileges per the ACL inheritance algorithm on the resource directly associated with it.
+[[Source](https://github.com/solid/specification/issues/145)]
+[[Source](https://github.com/solid/specification/issues/41)]
+
 ### Required client-side implementation ### {#http-client}
 
 A Solid client MUST be an HTTP/1.1 client [[!RFC7230]][[!RFC7231]].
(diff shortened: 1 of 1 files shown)


main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
-Servers MUST allow creating new resources with a POST request to URI path
+[[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)].
+
+### Reading resources ### {#read}
+
+Servers MUST support the HTTP `GET`, `HEAD` and `OPTIONS` methods [[!RFC7231]]
+for clients to read resources or to determine communication options.
+[[Source](https://github.com/solid/specification/issues/39#issuecomment-538017667)]
+
+When using Web Access Control ([[#wac]]):
```
RubenVerborgh, line 150: Why is this relevant?

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+
+Servers MUST indicate their support for HTTP Methods by responding to HTTP
+`GET`, `HEAD` and `OPTIONS` requests on the resource's URL with the HTTP
+Method tokens in the HTTP response header `Allow`.
```
RubenVerborgh, line 154: On what methods?

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+
+Servers MUST indicate supported media types in the `Accept-Patch` [[!RFC5789]]
+and the `Accept-Post` [[!LDP]] headers for matching HTTP methods in `Allow`
```
RubenVerborgh, line 157: On what methods?

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+header value.
+
+Servers MAY include the HTTP `Accept-Patch` and `Accept-Post` headers in the
+response of a `OPTIONS *` request.
```
RubenVerborgh, line 161: Can likely be removed if my above 2 questions are answered.

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+
+[[Source](https://github.com/solid/specification/issues/85#issuecomment-575386251)]
+[[Source](https://github.com/solid/specification/issues/43)]
+
+### Writing resources ### {#write}
+
+Servers MUST support the HTTP `PUT`, `POST` and `PATCH` methods [[!RFC7231]]
+for clients to create and update resources.
```
RubenVerborgh, line 169: What if my server is read-only?
I don't think it is a MUST, but rather a MAY.

RubenVerborgh: See https://github.com/solid/specification/issues/39#issuecomment-533298070

RubenVerborgh: Unless of course `405` is an an acceptable implementation.

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+header targets an auxiliary resource, the server `MUST` respond with the `403`
+status code and response body describing the error.
+[[Source](https://github.com/solid/specification/issues/42#issuecomment-616688848)]
+
+Servers SHOULD NOT allow HTTP `POST`, `PUT` and `PATCH` to update a
+container's containment triples; if the server receives such a request, it
+SHOULD respond with a `409` status code.
+[[Source](https://github.com/solid/specification/issues/40#issuecomment-573358652)]
+
+Clients MUST use the HTTP `If-None-Match` header with a value of `"*"` to
```
RubenVerborgh, line 202: Servers refuse if they don't?

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+prevent an unsafe method from updating an existing representation of the
+target resource when creation of a new resource is intended.
+[[Source](https://github.com/solid/specification/issues/108#issuecomment-567272797)]
+[[Source](https://github.com/solid/specification/issues/40#issuecomment-566995240)]
+
+Clients MUST use the HTTP `If-Match` header and HTTP `ETags` to ensure it
```
RubenVerborgh, line 208: Servers refuse if they don't? (Seems like this is covered by 428 below, but maybe best to connect those.)

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+[[Source](https://github.com/solid/specification/issues/42#issuecomment-616688848)]
 
-Clients who want to create hierarchical containers MUST use PUT and PATCH
-requests.
+<em>This section is non-normative.</em>
 
-[[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)].
+While it is technically possible to auxiliary resources using the HTTP `POST`
+method with the `Slug` header, it allows varying implementations due to its
+non-deterministic resource naming.
```
RubenVerborgh, line 230: Don't understand the implications of "allows varying implementations"; can we directly say what we want to say?

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+
+### Deleting Resources ### {#delete}
+
+Servers MUST support the HTTP `DELETE` method [[!RFC7231]].
```
RubenVerborgh, line 234: Idem as above; read-only Solid servers are a thing.

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+[[Source](https://github.com/solid/specification/issues/39#issuecomment-538017667)]
+
+When using Web Access Control ([[#wac]]):
+
+Servers MUST NOT allow delete requests to a storage's root container and
```
RubenVerborgh, line 239:
```suggestion
Servers MUST NOT allow `DELETE` requests to a storage's root container and
```

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+
+When a contained resource is deleted, the server MUST also delete the
+associated resources (see the [[#rm]] section).
+
+To delete an ACL resource (see [[#ar-wac]]), an `acl:agent` MUST have
+`acl:Control` privileges per the ACL inheritance algorithm on the resource
+directly associated with it.
+[[Source](https://github.com/solid/specification/issues/145)]
+
+When a `DELETE` method request is made to a container, the server MUST delete
```
RubenVerborgh, line 261:
```suggestion
When a `DELETE` request is made to a container, the server MUST delete
```

main/resource-access.bs
```diff
@@ -142,19 +139,151 @@ through POST, using the HTTP `Slug` header as defined in [[!RFC5023]].
...
+
+<em>This section is non-normative.</em>
+
+The server might perform additional actions, as described in the normative
+references like [[!RFC7231]]. For example, the server could remove membership
+triples referring to the deleted resource, perform additional cleanup tasks
+for resources it knows are no longer referenced or have not been accessed for
+some period of time, and so on.
+
+Subsequent `GET` requests to the deleted resource usually results in a `404`
```
RubenVerborgh, line 275:
```suggestion
Subsequent `GET` requests to the deleted resource usually result in a `404`
```

main/resource-access.bs
csarven: Originally this sentence was in the same paragraph as the former. I moved it to make it more clear for review. But I agree that it is best to specify if it stays in its own paragraph. If I moved it to the previous paragraph, would that address your question?

main/resource-access.bs
csarven: Same as https://github.com/solid/specification/pull/188#discussion_r451808566

main/resource-access.bs
RubenVerborgh: Yes, it would.

main/resource-access.bs
RubenVerborgh: Agreed.

main/resource-access.bs
RubenVerborgh: (Especially the "When using Web Access Control ([[#wac]]):" interjection was difficult in that sense.)

main/resource-access.bs
csarven: I suppose it can be removed. @kjetil raised the point that this should be said so that Solid implementations shouldn't spend time on it.

main/resource-access.bs
csarven: https://github.com/solid/specification/pull/188/commits/5ec6ad6536d5fbb4361399f432c7b7d67a6ec665

main/resource-access.bs
RubenVerborgh: OK, I did miss that it was `OPTIONS *` (so STAR), but perhaps we can cover that in one go above.

main/resource-access.bs
csarven: https://github.com/solid/specification/pull/188/commits/5ec6ad6536d5fbb4361399f432c7b7d67a6ec665

main/resource-access.bs
csarven: Does https://github.com/solid/specification/pull/188/commits/5ec6ad6536d5fbb4361399f432c7b7d67a6ec665 also address this?

RubenVerborgh: No, I'm not sure how this relates to WAC; value of the `Allow` header(s) should be independent of current user, right?

csarven: I'm not sure what would determine the allowed methods if not derived from an access policy. Alternatively, we'd expect all methods; GET, HEAD, OPTIONS, PUT, POST, PATCH, DELETE for every URL?

(Putting aside the other issue on if a server can be read-only.)

RubenVerborgh: > I'm not sure what would determine the allowed methods

Server support (so read-only or not is not a side issue for me here).

main/resource-access.bs
csarven: Decided to remove this block because it doesn't need to be mentioned.

main/resource-access.bs
csarven: I said aside for the moment because I didn't want to write out two sets above. So either:
* GET, HEAD, OPTIONS
* GET, HEAD, OPTIONS, PUT, POST, PATCH, DELETE

I suppose it is also possible to have a server that only allows creating, updating, deleting:
* PUT, POST, PATCH, DELETE

Noting that if only general server support (=what the spec prescribes) is listed, then we need something like WAC-Allow in order to help the client for their next request. It still doesn't tell them which methods are allowed - so potentially making the request but no success.

Noting also that having the server in read-only mode entails that information is coming from somewhere but presumably not the authorization policy. Otherwise, might as well map the access modes to the methods and list them in Allow.

RubenVerborgh: > then we need something like WAC-Allow in order to help the client for their next request

Yeah (but preferably a better mechanism).

But that is my reading of https://tools.ietf.org/html/rfc7231#section-7.4.1, which states:

> The "Allow" header field lists the set of methods advertised as supported by the target resource.

Note that it is about the _resource_, not about the current user-agent.

Although below, it says:

> The actual set of allowed methods is defined by the origin server at the time of each request.

And then again:

> An empty Allow field value indicates that the resource allows no methods

So I still see my interpretation validated (but the evidence is not explicit).

> Noting also that having the server in read-only mode entails that information is coming from somewhere

It might literally be a temperature sensor (and IoT was actually the case I was thinking of), with a specific ACL for reads still.

The difference is that, with a read/write server, the ACL might change. So `POST` is _supported_, just not _permitted_ for a certain user.

csarven: Henry captured the considerations well here:
https://github.com/solid/solid/wiki/Client-Authentication-Discovery#allow-headers

I'm fine with read-only or write-only use cases.

I'm not sure what you mean exactly re "specific ACL for reads still". Just want to understand if there is a way beyond an implementation configuration. In that case, we can qualify the operations eg. "Servers that want to enable read operations:" and similarly for the other operations.

Not specifying where the methods are coming from in addition to what the RFC says is also an option.

RubenVerborgh: > I'm not sure what you mean exactly re "specific ACL for reads still".

That was in response to you writing

> Otherwise, might as well map the access modes to the methods and list them in Allow.

where I wanted to remark that even in a read-only server, not all users might have read permissions, so such a mapping wouldn’t necessarily exist. So read-only and still having auth aren’t mutually exclusive.

csarven: So would removing "When using Web Access Control" also cover the read-only and write-only cases? Essentially how the methods in Allow are determined becomes an implementation detail.

RubenVerborgh: Works for me!

csarven: https://github.com/solid/specification/pull/188/commits/51bbd733b2feaf2073ab6d8bf720332c67f9fb0c

main/resource-access.bs
csarven: https://github.com/solid/specification/pull/188/commits/4d5cb9bac37e8206d267bf3df39a21840383b657

main/resource-access.bs
csarven: https://github.com/solid/specification/pull/188/commits/6eb9436bbc43318dcab1628bad16b9e3888fa2a6

main/resource-access.bs
csarven: The spec currently requires servers to implement 7232 - rough consensus in https://github.com/solid/specification/issues/33 seems to keep it that way - and there are no additional requirements as far as I can tell from the issues.

main/resource-access.bs
csarven: Same as https://github.com/solid/specification/pull/188#discussion_r452194752 . FYI, this requirement was derived from LDP's requirement.

main/resource-access.bs
csarven: https://github.com/solid/specification/pull/188/commits/09cc77a20cab781177e18e4b970cd4e3426ee5d0 should ties this up.

main/resource-access.bs
csarven: https://github.com/solid/specification/pull/188/commits/09cc77a20cab781177e18e4b970cd4e3426ee5d0 should ties this up.

main/resource-access.bs
```diff
@@ -130,31 +125,150 @@ Note:
...
+[[Source](https://github.com/solid/specification/pull/160#issuecomment-636822687)].
+
+### Reading resources ### {#read}
+
+Servers MUST support the HTTP `GET`, `HEAD` and `OPTIONS` methods [[!RFC7231]]
+for clients to read resources or to determine communication options.
+[[Source](https://github.com/solid/specification/issues/39#issuecomment-538017667)]
+
+Servers MUST indicate their support for HTTP Methods by responding to HTTP
+`GET`, `HEAD` and `OPTIONS` requests on the resource's URL with the HTTP
```
acoburn, line 155: I am a little concerned about the inclusion of `OPTIONS` here. If `OPTIONS` requests need to include the appropriate `Allow` header, then consider the case where an RDF resource supports PATCH but a NonRDF resource doesn't. Because CORS needs to be able to send unauthenticated preflight requests to that URL, if `OPTIONS` responses distinguish between resource types to unauthenticated clients, then a malicious (unauthenticated) client can use this method to probe a server to find out information about what data is stored there. To include `OPTIONS` here, while also supporting CORS (which I assume is a given) would open up a server implementation to a variety of attacks and/or data leakages.

I would recommend simply removing `OPTIONS` from this list.

c.f. also https://www.mnot.net/blog/2012/10/29/NO_OPTIONS

main/resource-access.bs
```diff
@@ -130,31 +125,150 @@ Note:
...
+
+When a `PUT` or `PATCH` method request targets an auxiliary resource, the
+server MUST create or update it. When a `POST` method request with the `Slug`
+header targets an auxiliary resource, the server MUST respond with the `403`
+status code and response body describing the error.
+[[Source](https://github.com/solid/specification/issues/42#issuecomment-616688848)]
+
+Servers SHOULD NOT allow HTTP `POST`, `PUT` and `PATCH` to update a
+container's containment triples; if the server receives such a request, it
+SHOULD respond with a `409` status code.
```
acoburn, line 198: Is it legit for a server to simply ignore any user-supplied containment triples? For instance, say you have some user-managed triples in a container (along with some containment triples). Assume that you have two clients manipulating the data of that container. The first client fetches the RDF representation of that container (which includes the containment triples and an `<> rdfs:label "some label"` triple). The second client `POST`s to the container, adding a new child resource. Then the first client issues a `PUT` (missing the newly created child resource but with the updated `rdfs:label` value). Assume the clients are not using `If-Match` headers.

Should the sever accept the first client's PUT request (and just ignore the containment triples)? Or is it the responsibility of the client to remove those triples from the request? Or must a client use PATCH in this case? Or must a client use `If-Match` headers in all cases? (the section below suggests that conditional requests are at the `MAY` level)

main/resource-access.bs
```diff
@@ -130,31 +125,150 @@ Note:
...
+
+### Deleting Resources ### {#delete}
+
+When a server supports the HTTP `DELETE` method [[!RFC7231]] this
+specification imposes the following requirements:
+[[Source](https://github.com/solid/specification/issues/39#issuecomment-538017667)]
+
+When a `DELETE` request targets storage's root container or its associated ACL
+resource, the server MUST respond with the `405` status code. Server MUST
+exclude the `DELETE` method in the HTTP response header `Allow` in response to
```
acoburn, line 223: Just to clarify: does this mean that, in order to delete one's Pod, there needs to be some separate (implementation specific?) mechanism for deleting the root container?

csarven: Yes. Based on the available information, I'm inclined to say that how the root container come into existence (or commissioned) and cease to exist is unknown. It is an acceptable axiom. [The observable universe is after the big bang and just before it collapses? - /me pokes @kjetilk ]

Prohibiting root container or its ACL from getting deleted was to partly address loss of control. We could of course allow an authorized agent to delete the root container (provided that the container is empty), and subsequent requests to read the root container may get a 404/405/410. Practically speaking, I'm not sure if the difference between an empty root (and no public read) and one that does not really exist is significant. Even once deleted, there'd be no particular way to reinstate the root container given that the agent that was previously authorized is presumably no longer known to the system.

An alternative is to have an agent with the Control access (with the role of admin/controller/authority/creator..) on the root container where the authorization policy is sticky, immutable or maintains a specific shape and data. "But wait!" you say, how did that information come to be?

So, we are back again to a separate mechanism that somehow commissions the root container/storage as well as being able to decommission. We could specify an out of band management mechanism but that'd still be orthogonal.

Open to ideas / implementation experience that can give us a more useful axiom to work with.

acoburn: re: loss of control, I am experimenting with some implementation behavior in which there really is a fixed owner for a Pod. It would rely on an out-of-band provisioning/deprovisioning model, which is hinted at in the text. It also relates to authZ, but that's a separate discussion. Main point: it sounds like a separate Pod deprovisioning mechanism would be OK.

main/resource-access.bs
csarven: In the initial PR ( https://github.com/solid/specification/commit/51bbd733b2feaf2073ab6d8bf720332c67f9fb0c ) the `Allow` header was used in context of WAC. In https://github.com/solid/specification/pull/188#pullrequestreview-445094017 , Ruben and I agreed that the `Allow` header only indicates server's support - which was already stated in the spec: "Servers [..] indicate their support for HTTP Methods [..] on the resource's URL [..]"

Holding that perspective may mean that `Allow` does not need to distinguish between resource types to unauthenticated clients. Essentially `PATCH` is listed in `Allow` regardless of media types actually supported for a resource. Unless we are explicit about the methods to always list in `Allow` ie. equivalent to server requirements.

For `Accept-Patch` and `Accept-Post` headers in response to `OPTIONS` to be useful, it needs to distinguish between resource types. That would indeed expose information to unauthenticated clients. So, I agree to remove `OPTIONS` from the list, perhaps more specifically, the header should not be in the response to an authenticated request.

If `Accept-Patch` and `Accept-Post` values distinguish between resource types, would that mean that an agent needs to have `acl:Read` on the resource? Related issue: https://github.com/solid/specification/issues/43

main/resource-access.bs
csarven: If we are still in agreement on https://github.com/solid/specification/issues/40#issuecomment-573358652 , I think it entails that 409 will precede conditional errors (412, and probably for 428) as per https://tools.ietf.org/html/rfc7232#section-5 :

>   redirects and failures take precedence over the evaluation of
>   preconditions in conditional requests.

Does that make sense? If so, to take it a step further:

it may also make sense that the client should follow up with `If-Match` on their next write request. Of course they'll need to get a hold of the latest resource state before doing so. Without additional instructions to give to a client, it is possible that a simple client will not get its update through when the validator on a resource changes since their last interaction with the resource. Server could inform exactly what happened and what the client should do next in the response body of a 409 but it is not entirely reliable if this were to be automated - we'd have to bake in additional know-how on how clients can make use of structured problem details.

I'm also not confident about how sensible it'd be for a server to use an ETag in the response of a 409 for a write request.

Alternatively, we can revisit https://github.com/solid/specification/pull/188/commits/49cd2f18e55f21df339226cf8e63c5f5424354e5 in that client should use `If-Match` in requests if server includes an ETag with strong validator. However! I'm not aware of if/how clients can make decisions based on strong or weak validator in ETag. So, this may not actually work in practice?

--- 

We arrived here because we didn't indicate weak or strong validator use for servers, and especially on handling RDF sources. Reverting the commit (linked above) was the simple way out.

main/resource-access.bs
csarven: Revised: https://github.com/solid/specification/pull/188/commits/78507ffe0be507c8616073cd9dd487759efcdf77 . What do you think?

acoburn: @csarven that's great, thanks

main/resource-access.bs
acoburn: Just to circle back on this discussion, I am 👍  on the text here, as it relates to `DELETE`

main/resource-access.bs
acoburn: I think we are in agreement w/r/t conditional requests and all the complexities of ETags. What I was specifically interested in with my question relates to triples that are server-managed (e.g. `<> ldp:contains <./resource>`), and whether a server can ignore such server-managed triples on inbound client write requests. A client shouldn't be able to set these triples anyway and so (presumably) a server would need some mechanism to distinguish between those triples and other, client-managed triples.

To be clear, assume you have a resource called `./container/`. A GET requests produces the following:

```turtle
@prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix ldp:   <http://www.w3.org/ns/ldp#> .
@prefix rdfs:  <http://www.w3.org/2000/01/rdf-schema#> .

<>  rdf:type     ldp:BasicContainer ;
    rdfs:label "A Conttainer"@en ;
    ldp:contains <./resource> .
```

But I realize that the label was misspelled, so I want to perform a PUT to replace that label with the following Turtle: 

```turtle
@prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix ldp:   <http://www.w3.org/ns/ldp#> .
@prefix rdfs:  <http://www.w3.org/2000/01/rdf-schema#> .

<>  rdf:type     ldp:BasicContainer ;
    rdfs:label "A Container"@en ;
    ldp:contains <./resource> .
```

Assume that the `rdf:type` and `ldp:contains` triples are managed by the server. And assume that the client request isn't attempting to modify any containment triples.

But imagine that, between the GET and the PUT request, another agent has created a second child resource by POSTing to that container. Now there are two `ldp:contains` triples. That means that the PUT request by our client is now out of sync with the state of the container (a conditional request `If-Match` would catch that, but for the sake of argument, we're not using conditional requests). Specifically, the containment triples in the client request are not accurate from the perspective of the server, it is ambiguous whether a client is actually attempting to "update a container's containment triples".

The `SHOULD` language is probably sufficient, but I'm trying to tease out whether, for server-managed triples (such as containment) a server implementation can simply ignore those client-supplied triples or whether the server needs to respond with a 409 Conflict error.

csarven: This is great feedback, thanks!

Servers should not allow clients to modify containment triples. Put another way, two subsequent container states can't have different containment triples by way of client attempting to update the container. Server responds with 409.

The spec essentially states that clients MAY use `If-Match` by way of implementing RFC 7232's client parts. In order to encourage clients to opt-in to using the `If-Match` header, I've added an optional for servers to use the strong validator on RDF bearing representations: https://github.com/solid/specification/pull/188/commits/b519bf83c0962dfdd17647f946315e1cacaeac3c

Regarding the (SHOULD NOT) requirement level: Currently there is a requirement to have all resources be discoverable and no orphaned resources can be created ( https://github.com/solid/specification/issues/97#issuecomment-547459396 ). It follows that while SHOULD NOT may be sufficient, MUST NOT can ensure that all resources are discoverable. SHOULD NOT permits the system to have "hidden" resources by way of client removing a containment statement while leaving the resource (it contained just earlier) untouched. So, is there a reason to use SHOULD NOT?

acoburn: I agree that MUST NOT would ensure that all resources are discoverable. I would favor that approach.
In other words, I don't have a use case that would suggest using SHOULD NOT, though I'm also not against the use of SHOULD NOT.

csarven: Let's go with MUST NOT at least towards spec Draft until a case for SHOULD NOT is clear: https://github.com/solid/specification/pull/188/commits/d7b8039691f192fa58c53783f3bd4f13c8bde50e . Thanks.

Issue #195, opened 2019-07-17
PUT RDFa, then GET turtle


michielbdejong:
After storing an RDFa document it should be retrievable as Turtle or JSON-LD.

Two use cases:
* A page like https://csarven.ca/#i to be an html page, but also to act as my WebID for Solid
* A blog post that's a recipe as an html page, but a Solid app can retrieve it as Turtle to put the ingredients on my shopping list.

csarven:
Generally resolved by https://github.com/solid/specification/pull/196 . Re `Accept-Put`: https://github.com/solid/specification/pull/201

Issue #109, opened 2019-11-05
Representation URLs and interactions


csarven:
Should a resource's representation URL(s):
* exist?
* if exists, should a resource advertise it through `Content-Location`?
* if exists, should there be a redirect to the representation URL?
* if the representation can be accessed, are there any constraints on i) reading ii) writing?
* if writing is allowed on a representation URL, are there any side effects to other representation URLs?

RubenVerborgh:
My answers below:

> * exist?

MAY

> * if exists, should a resource advertise it through `Content-Location`?

SHOULD

> * if exists, should there be a redirect to the representation URL?

MAY

> * if the representation can be accessed, are there any constraints on i) reading ii) writing?

Reading: same as original resource.
Writing: I'd be inclined to say, only the specific representation then.

> * if writing is allowed on a representation URL, are there any side effects to other representation URLs?

Yes; representations stay connected to the resource.

kjetilk:
> > * if exists, should a resource advertise it through `Content-Location`?
> 
> SHOULD

With the guideline to avoid optional features, I am inclined to say MUST there. Other than that, I agree.

csarven:
One design consideration is to minimise the number of interactions around a resource. From that perspective, representation URLs would not be exposed. Alternatively, if exposed, certain requests methods (basically writing; `POST`, `PUT`, `PATCH`) should be disallowed.

There are of course some advantages to exposing the representation URLs (eg caching), so making that optional seems reasonable, with the expectation that if supported, there is a required behaviour.

Based on that, one possibility:

>* exist?

MAY

>* if exists, should a resource advertise it through `Content-Location`?

MUST

* if exists, should there be a redirect to the representation URL?

MUST NOT. (MAY could be an alternative but that explicitly introduces new affordances around the representation.)

>* if the representation can be accessed, are there any constraints on i) reading ii) writing?

Reading: don't see a need to set any. Anything in particular we need to consider?

Writing: `405`. Writing must happen through the original resource. Side effect: don't have to specify or manage access controls for representations. `Allow` only includes read methods; `GET`, `OPTIONS`, `HEAD`.

>* if writing is allowed on a representation URL, are there any side effects to other representation URLs?

By only allowing writing to go through the resource, server implementation is somewhat simplified ie. doesn't have to bounce from writing to a particular representation and then generating equivalents for the other representation URLs. Things should exist and die with the resource. Bonus: avoid write access control policies around representations.

RubenVerborgh:
[...]

Issue #45, opened 2019-09-19
Required RDF serializations for Solid


acoburn:
The LDP specification requires read support for both [`text/turtle`](https://www.w3.org/TR/ldp/#h-ldprs-get-turtle) and [`application/ld+json`](https://www.w3.org/TR/ldp/#h-ldprs-get-jsonld). For servers that allow write operations, support for those [two](https://www.w3.org/TR/ldp/#h-ldpc-post-turtle) [formats](https://www.w3.org/TR/ldp/#h-ldpc-post-jsonld) are also required.

Does Solid impose any additional requirements about the support of other RDF serializations for read and/or write operations?

csarven:
If `PUT` becomes a requirement (pending https://github.com/solid/specification/issues/39 , https://github.com/solid/specification/issues/40 ) or at least at the same level as LDP's `POST` ie. optional with some requirements for supporting servers, I'd suggest that the same RDF serializations are supported ie. Turtle, JSON-LD.

Clients intending to create an LDP-RS using a payload in other than Turtle and JSON-LD MUST fulfill https://www.w3.org/TR/ldp/#ldpc-post-createrdf for `POST` as well as `PUT` (pending agreement on support).

Servers supporting Turtle, JSON-LD and optionally other RDF serializations for LDP-RS SHOULD provide different RDF serializations based on client's proactive negotiation. For example, if a server allows the creation of an LDP-RS in `text/html` (including RDFa), it SHOULD respond to `GET` with `Accept: text/turtle` requests with `200`.

csarven:
Noting here that the WebID unofficial draft has MUST for Turtle, allows content-negotiation for other RDF serializations, and has MAY for RDFa. In that regard, the RDFa for WebID raises a slightly stronger recommendation than JSON-LD for WebID. This should also be taken into account in Authentication for both servers and clients.

Also noting here a general status update: the current spec does not require notion of LDP-RS but instead uses the notion of RDF document as in https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-document which includes Turtle, RDFa, JSON-LD, TriG as concrete syntaxes. Will clarify these bits in upcoming PR.

csarven:
Resolved by https://github.com/solid/specification/pull/196

PR #196, opened 2020-09-04
Add payload and representations in RDF

csarven:
Initial requirement on Representations. This particular requirement involves issues:

* https://github.com/solid/specification/issues/45
* https://github.com/solid/specification/issues/69
* https://github.com/solid/specification/issues/109
* https://github.com/solid/specification/issues/195

Will update this space to include related issues.

Feedback on making this more readable is most welcome.

### DIFF
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -286,11 +286,16 @@ such that the request's representation data encodes an *RDF document*
 [[!RDF11-CONCEPTS]] (as determined by the `Content-Type` header), the server
 MUST respect `GET` requests on this resource when the value of the `Accept`
 header requests a representation in `text/turtle` or `application/ld+json`.
-
 [[Source](https://github.com/solid/specification/issues/45)]
 [[Source](https://github.com/solid/specification/issues/69)]
 [[Source](https://github.com/solid/specification/issues/109)]
 [[Source](https://github.com/solid/specification/issues/195)]
+
+When a `PUT`, `POST`, `PATCH` or `DELETE` method request targets a resource's
+representation URL, the server MUST respond with a `307` or `308` status code
+and `Location` header specifying the preferred URI reference.
+[[Source](https://github.com/solid/specification/issues/109)]
+
 
 ## Auxiliary Resources ## {#rm}
 
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -284,7 +284,7 @@ https://github.com/solid/specification/issues/41#issuecomment-534679278
 When a server creates a resource on HTTP `PUT`, `POST` or `PATCH` requests
 such that the request's representation data encodes an *RDF document*
 [[!RDF11-CONCEPTS]] (as determined by the `Content-Type` header), the server
-MUST respect `GET` requests on this resource when the value of the `Accept`
+MUST accept `GET` requests on this resource when the value of the `Accept`
 header requests a representation in `text/turtle` or `application/ld+json`
 [[!Turtle]] [[!JSON-LD11]].
 [[Source](https://github.com/solid/specification/issues/45)]
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -286,7 +286,7 @@ such that the request's representation data encodes an *RDF document*
 [[!RDF11-CONCEPTS]] (as determined by the `Content-Type` header), the server
 MUST respect `GET` requests on this resource when the value of the `Accept`
 header requests a representation in `text/turtle` or `application/ld+json`
-[[!Turtle]] [[!JSON-LD]].
+[[!Turtle]] [[!JSON-LD11]].
 [[Source](https://github.com/solid/specification/issues/45)]
 [[Source](https://github.com/solid/specification/issues/69)]
 [[Source](https://github.com/solid/specification/issues/109)]
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -292,9 +292,10 @@ header requests a representation in `text/turtle` or `application/ld+json`
 [[Source](https://github.com/solid/specification/issues/109)]
 [[Source](https://github.com/solid/specification/issues/195)]
 
-When a `PUT`, `POST`, `PATCH` or `DELETE` method request targets a resource's
-representation URL, the server MUST respond with a `307` or `308` status code
-and `Location` header specifying the preferred URI reference.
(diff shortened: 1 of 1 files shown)


bblfish:
There seems to be another question which is if a PUT of Turtle doc is done at `<foaf.ttl>`, should the `<>` refer to `<foaf.ttl>` or to the content negotiable `<foaf>`? Should clients rather PUT at `<foaf>`, and have `<foaf.ttl>` be linked to from there? The same question for jpegs and image formats and video formats.

michielbdejong:
The way it currently works in Solid is that if you PUT turtle to `https://example.com/profile/card`, then that is its URL. You can then fetch it as Turtle or as JSON-LD or as human-viewable HTML at `https://example.com/profile/card` (or `https://example.com/profile/card#me` within that document). At no point is there any need to access the Turtle and JSON-LD (which is what we call "representation of the resource" right?) at a URL different from `https://example.com/profile/card`.

csarven:
@bblfish I've raised something similar in https://github.com/solid/specification/issues/69#issuecomment-581340858 in context of container representations. The suggested practice for clients to not explicitly set base URI in the document, however, if set, it should refer to container's URI (as opposed to a representation URL).

>Should clients rather PUT at <foaf>, and have <foaf.ttl> be linked to from there?

Yes. Some implementations may expose representation URLs (covered in issue #109 ) but we want writes to go through the preferred content negotiable URI of the resource (ie. `<foaf>`). So, `GET /foaf` with `Accept: text/turtle` may get a response including `Content-Location: /foaf.ttl`. I've updated this PR covering what's in 109.

Note https://tools.ietf.org/html/rfc7231#appendix-B :
>   The definition of Content-Location has been changed to no longer
   affect the base URI for resolving relative URI references, due to
   poor implementation support and the undesirable effect of potentially
   breaking relative links in content-negotiated resources.
   (Section 3.1.4.2)

So, if `/foaf.ttl` includes `<>` (or equivalent in other RDF syntaxes), `GET /foaf`'s response with `Content-Location: /foaf.ttl` entails that when the document is parsed `<>` refers to URI ending `/foaf`.

Having said that, I'm not sure what servers should do, if anything, when they encounter `<>` in PUT's payload. Preferably nothing. Clients should parse the content at the resource URI instead of the representation URL to obtain the RDF graph. Is this effectively covered in the specs or do we need to mention that at all? We want to avoid exposing multiple graphs where they may only differ with references resolved against the base URI.

csarven:
@michielbdejong Issue #109 raised whether implementations may advertise representation URLs through `Content-Location`. FWIW, servers are not forbidden from advertising representation URLs, so if some servers expose them, how should read and write operations work? The agreement was that write requests targeting the representational URL should be done through the URI of the resource.

michielbdejong:
OK then it's #109 I disagree with :)

main/resource-access.bs
```diff
@@ -279,6 +279,25 @@ Issue:
 Pertaining to events and loss of control mitigation:
 https://github.com/solid/specification/issues/41#issuecomment-534679278
 
+### Representations ### {#representations}
+
+When a server creates a resource on HTTP `PUT`, `POST` or `PATCH` requests
+such that the request's representation data encodes an *RDF document*
+[[!RDF11-CONCEPTS]] (as determined by the `Content-Type` header), the server
+MUST respect `GET` requests on this resource when the value of the `Accept`
```
RubenVerborgh, line 287: Yes to the idea, but the phrasing of `respect` is a bit vague/unclear to me

csarven: Updated to use "accept": https://github.com/solid/specification/pull/196/commits/4d8858654c4056ef53f0420cb9118f1f7ba2b9a7 ie. to honour, conform to, follow through client's wishes - matching 7231's "received, understood, and accepted" for 2xx. Will this do for now? Happy to revise later.

RubenVerborgh (approved):

akuckartz (approved):

csarven:
@michielbdejong On the contrary, we agree on the same requirement: clients should target the resource URI for write operations.

Two approaches for Solid where both would conform to RFC 7231 (`Content-Location` is optional for servers and clients):

1. Servers and clients MUST NOT use `Content-Location`.
2. Servers MUST redirect (eg. `307`, `308` - this PR) or fail (eg. `405`) in response to write requests on representation URLs.

Solid servers MUST implement the server parts of RFC 7231 and `Content-Location` still remains optional. Disallowing `Content-Location` seems like a significant limitation (approach 1) in comparison to specifying what should happen when server uses `Content-Location` (approach 2).

Open to alternative approaches or an update to what's proposed.

main/resource-access.bs
```diff
@@ -279,6 +279,25 @@ Issue:
...
+MUST accept `GET` requests on this resource when the value of the `Accept`
+header requests a representation in `text/turtle` or `application/ld+json`
+[[!Turtle]] [[!JSON-LD11]].
+[[Source](https://github.com/solid/specification/issues/45)]
+[[Source](https://github.com/solid/specification/issues/69)]
+[[Source](https://github.com/solid/specification/issues/109)]
+[[Source](https://github.com/solid/specification/issues/195)]
+
+When a `PUT`, `POST`, `PATCH` or `DELETE` method request targets a resource's
+representation URL, the server MUST respond with a `307` or `308` status code
```
acoburn, line 296: The redirect requirement makes sense in the case where the resource URL (e.g. `http://example.com/foo`) is different from the representation URL (e.g. `http://example.com/foo.jsonld`), but what if all representations are fetched from the same resource URL (`/foo`)? Should this sentence include something like "if the representation URL is different than the resource URL"? Or perhaps the term "representation URL" should be more rigorously defined?

csarven: I agree that'd be good to distinguish. Initial update here: https://github.com/solid/specification/pull/196/commits/e92d8bce5caaad42fcaad465ba326e44fc1357fc . Will this do for now? I like the idea of further clarifying how we are using "resource" and "representation" but we shouldn't define per se - perhaps just remind RFC's notions.

acoburn: @csarven thanks, that update works for me

acoburn (approved):

justinwb (approved):

michielbdejong:
Yes, approach 2 is fine.

csarven:
Great, thanks all! All approved, feedback and objections addressed.

PR #199, opened 2020-09-16
Add section for acl Link relation type

csarven:
With intent to submit to IANA for review, approval, and inclusion in the IANA Link Relations registry.

* https://github.com/solid/specification/issues/54
* https://github.com/solid/web-access-control-spec/issues/21

### DIFF
--- a/main/http-definitions.bs
+++ b/main/http-definitions.bs
@@ -1,2 +1,27 @@
 HTTP Definitions {#http-definitions}
 ===================================
+
+## Link Relations ## {#link-relations}
+
+The intent is that these link relations will be registered with IANA per [[!RFC8288]]. 
+
+### acl ### {#acl}
+
+The contents of this section were originally taken from [Web Access Control](https://www.w3.org/wiki/WebAccessControl).
+
+The following Link Relationship will be submitted to IANA for review, approval, and inclusion in the IANA Link Relations registry.
+
+: Relation Name
+:: `acl`
+: Description
+:: The relationship `A acl B` asserts that resource B provides access control description of resource A. There are no constraints on the format or representation of A. B MUST be a Access Control List document using a Resource Description Framework syntax.
+: Reference
+:: This specification.
+: Notes
+:: Consumers of ACL resources should be aware of the source and chain of custody of the data.
+
+[[Source](https://github.com/solid/specification/issues/54)]
+[[Source](https://github.com/solid/web-access-control-spec/issues/21)]
+
+Issue:
+Shape of ACL: https://github.com/solid/specification/issues/169
--- a/main/http-definitions.bs
+++ b/main/http-definitions.bs
@@ -14,7 +14,7 @@ The following Link Relationship will be submitted to IANA for review, approval, 
 : Relation Name
 :: `acl`
 : Description
-:: The relationship `A acl B` asserts that resource B provides access control description of resource A. There are no constraints on the format or representation of A. B MUST be a Access Control List document using a Resource Description Framework syntax.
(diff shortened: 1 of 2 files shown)


main/http-definitions.bs
```diff
@@ -0,0 +1,27 @@
...
+
+: Relation Name
+:: `acl`
+: Description
+:: The relationship `A acl B` asserts that resource B provides access control description of resource A. There are no constraints on the format or representation of A. B MUST be a Access Control List document using a Resource Description Framework syntax.
+: Reference
+:: This specification.
+: Notes
+:: Consumers of ACL resources should be aware of the source and chain of custody of the data.
+
```
bblfish, line 22: what is "the chain of custody of the data"?

csarven: Authority. Just an initial note. If it is not significant we can remove it. The note can be empty or anything ( https://www.iana.org/assignments/link-relations/link-relations.xhtml )

While Solid currently expects both the resource and its ACL on the same server, it has applicability beyond Solid - which is why we are considering it to push it through IANA. So, technically possible to have A and B on different servers and authority. I don't suggest that restriction is placed on the definition of the relation ie. okay to keep it only for Solid for now.

bblfish: I think it makes sense to have the ACL somewhere else. They could be on a future hyper-blockchain build on linked data for example.

bblfish: Anyway, I think the phrasing could be better. But presumably this will be reviewed a lot more before sending it to IANA.

csarven: In addition to more reviewing, it needs to refer to an actual spec - so a version of the Solid spec that can be acknowledged.

bblfish (approved):

kidehen:
I also recommend approval of these changes.

acoburn (approved):
LGTM

justinwb (approved):
✅

main/http-definitions.bs
```diff
@@ -0,0 +1,27 @@
...
+### acl ### {#acl}
+
+The contents of this section were originally taken from [Web Access Control](https://www.w3.org/wiki/WebAccessControl).
+
+The following Link Relationship will be submitted to IANA for review, approval, and inclusion in the IANA Link Relations registry.
+
+: Relation Name
+:: `acl`
+: Description
+:: The relationship `A acl B` asserts that resource B provides access control description of resource A. There are no constraints on the format or representation of A. B MUST be a Access Control List document using a Resource Description Framework syntax.
```
RubenVerborgh, line 17:
```suggestion
:: The relationship `A acl B` asserts that resource B provides access control description of resource A. There are no constraints on the format or representation of A. B MUST be an Access Control List document using a Resource Description Framework syntax.
```

RubenVerborgh: Why are we constraining B? Also, RDF syntax is a representation constraint, not a resource constraint.

csarven: I've considered this as well. So, okay to remove the constraint ( https://github.com/solid/specification/pull/199/commits/b6702a8f50ceebd9112071c5e244aa8bfe750ce3 ) as far as the generic use of the relation type is concerned for IANA.

For Solid's use of `acl` however, B indeed MUST be an ACL document in RDF.

bblfish: So that info could come be passed on to the client through the resource advertising itself as a `solid:Resource`.
We should keep a list of things that we can gain if we have `solid:Resource` type link type relation.

bblfish: I agree with @RubenVerborgh if you made it a requirement that it point to an RDF document, you'd have quite a lot of resistance in the IETF. (perhaps less so nowadays, but still)

csarven:
Great, thanks all! All approved and feedback addressed.

PR #209, opened 2020-11-10
Use consistent dash

janschill:
The n-dash has been used before, therefore it should be changed here as well.

### DIFF
--- a/main/resource-access.bs
+++ b/main/resource-access.bs
@@ -96,7 +96,7 @@ Behaviour pertaining to authorization MUST proceed this optional redirect
 ### Storage ### {#storage}
 
 When a server supports a data pod, it MUST provide one or more storages
-(`pim:Storage`) - a space of URIs in which data can be accessed. A storage is
+(`pim:Storage`) – a space of URIs in which data can be accessed. A storage is
 the root container for all of its contained resources (see
 [[#resource-containment]]).
 


TallTed:
Unfortunately, the m-dash is the appropriate one for this instance.

Github web search doesn't find either m-dash or n-dash, so it's not easy to find the other occurrence(s) to which this PR refers. Possibly someone who uses the command line tools can do a `sed` or similar.

janschill:
You are right. This is quite interesting, as in German and here I am not sure what other languages, you use the n-dash as a substitution for a comma, but it seems not in English typography. There it should be the m-dash without any space in between—just like this.

TallTed:
@janschill —

Whether spaces should wrap the m-dash remains a debated question. (I'm in favor of the wrapping spaces, as I believe they improve readability.)

In English, approximately: 

* hyphens (without wrapping spaces) connect two words into one, as in "tip-top"
* n-dashes (without wrapping spaces) join the ends of ranges, as in "3–10"
* m-dashes (with or without wrapping spaces) separate phrases (or interjections) within sentences, as in "I once thought — but no longer do — that 'solid' was just one of the three states of matter; now I know that [list is much longer)](https://en.wikipedia.org/wiki/List_of_states_of_matter), and Solid is also "[re-decentralizing the web](https://github.com/solid/)."

Issue #181, opened 2019-08-14
Document the WAC-Allow header in the spec.


timbl:
This allows client code to be smart about whether to give the user editing interfaces to data or just view interfaces. 

It also allows them to know whether it open to the public, which might affect for example, whether the user is invited to make a public link, like, bookmark, etc of the resource.

csarven:
Proposal/requirements merged via PR https://github.com/solid/specification/pull/210 . See further details/discussion on concerns of this header, as well as use cases for different permission groups. Implementation feedback is always useful/welcome. Please follow up with new issues/proposals.

Issue #170, opened 2019-05-06
Is WAC-Allow part of the solid spec?


Otto-AA:
I've seen that the node-solid-server implemented a `WAC-Allow` header for identifying which access is granted for the requested resource. Is this also part of the solid api spec, or just a feature of this specific implementation?

Related issue in NSS: https://github.com/solid/node-solid-server/issues/246

RubenVerborgh:
Let's also have an issue for the problem; as this is just one possible solution. Other solutions include the `Link` header.

csarven:
Transferred https://github.com/solid/specification/issues/171 as the original problem.

csarven:
Proposal/requirements merged via PR https://github.com/solid/specification/pull/210 . See further details/discussion on concerns of this header, as well as use cases for different permission groups. Implementation feedback is always useful/welcome. Please follow up with new issues/proposals.

PR #210, opened 2020-11-15
Add WAC-Allow definition and requirements

csarven:
In response to the following (and possibly other) issues on the `WAC-Allow` header:

* https://github.com/solid/specification/issues/170
* https://github.com/solid/specification/issues/181

Future considerations:

In order for this HTTP header to be reusable by any access control system, we can consider changing the header field-name as well as the parameters and values.

Introduce new modes eg. "delete", if and when needed.

At the moment, access modes corresponds to the ones defined in the ACL ontology (also kept in context of the Web Access Control section), however, we can relax this with some editing.

There is also the alternative approach to using URIs instead of human-friendly terms for the allowed access modes. While that'd be precise it may be less flexible if access control systems change.

### DIFF
--- a/main/authorization.bs
+++ b/main/authorization.bs
@@ -58,8 +58,8 @@ or `GET` request on the target URL, and checking the `WAC-Allow` header value
 for access parameters listing the allowed access modes per permission group.
 
 Clients' parsing algorithm for the `WAC-Allow` header should incorporate error
-handling. When the received message fails to match an allowed pattern,
-unrecognised access parameters or modes are found, clients MUST ignore the
+handling. When the received message fails to match an allowed pattern, finds 
+unrecognised access parameters or access modes, clients MUST ignore the
 received `WAC-Allow` header-field.
 
 [[Source](https://github.com/solid/specification/issues/171)]
--- a/main/authorization.bs
+++ b/main/authorization.bs
@@ -38,17 +38,29 @@ OWS                 = *(SP / HTAB)
 RWS                 = 1*(SP / HTAB)
 ```
 
-The `access-mode` corresponds to the modes of access (`acl:Read`, `acl:Write`,
-`acl:Append`, `acl:Control`) as defined in the ACL ontology.
+The `WAC-Allow` HTTP header's field-value is a comma-separated list of access
+parameters (`access-param`). `access-param` is a whitespace-separated list of
+access modes ( `access-mode`) granted to a permission group.
 
-Clients can discover their access privileges on a resource by making an HTTP
-`HEAD` or `GET` request on the target URL, and checking the `WAC-Allow` HTTP
-header value for `user` and `public` access paramaters listing the allowed
-access modes.
+This specification defines the following permission groups:
+
+<dl>
+  <dt>`user`</dt>
+  <dd>Permissions granted to the agent requesting the resource.</dd>
+  <dt>`public`</dt>
+  <dd>Permissions granted to the public.</dd>
+</dl>
+
+`access-mode` corresponds to the modes of access as defined in the
+ACL ontology (`acl:Read`, `acl:Write`, `acl:Append`, `acl:Control`).
+
+Clients can discover access privileges on a resource by making an HTTP `HEAD`
+or `GET` request on the target URL, and checking the `WAC-Allow` header value
+for access parameters listing the allowed access modes per permission group.
 
 Clients' parsing algorithm for the `WAC-Allow` header should incorporate error
 handling. When the received message fails to match an allowed pattern,
-unrecognised access paramaters or modes are found, clients MUST ignore the
+unrecognised access parameters or modes are found, clients MUST ignore the
 received `WAC-Allow` header-field.
 
 [[Source](https://github.com/solid/specification/issues/171)]
--- a/main/authorization.bs
+++ b/main/authorization.bs
@@ -3,12 +3,14 @@ Authorization {#authorization}
 
 ## Web Access Control ## {#web-access-control}
 
-Web Access Control (WAC) is a decentralized cross-domain access control
-system. The WAC mechanism is concerned with giving access to agents denoted by
-a [[#webid]] to perform various kinds of read-write operations on resources
-identified by URLs. The Access Control List (ACL) vocabulary is used to
-describe authorization policies about authorized agents with modes of access
-on target resources.
+Web Access Control (<abbr title="Web Access Control">WAC</abbr>) is a
+decentralized cross-domain access control system. The WAC mechanism is
+concerned with giving access to agents denoted by a [[#webid]] to perform
+various kinds of read-write operations on resources identified by URLs. The
+<cite><a href="http://www.w3.org/ns/auth/acl">Access Control List</a></cite>
+(<abbr title="Access Control List">ACL</abbr>) ontology is used to describe
+authorization policies about authorized agents with modes of access on target
+resources.
 
 Servers MUST conform to the Web Access Control specification [[!WAC]].
 
@@ -19,3 +21,31 @@ the HTTP `Link` header with a `rel` value of `acl` [[#acl]].
 In the event that a server can't apply an ACL to a resource, it MUST deny
 access.
 [[Source](https://github.com/solid/specification/issues/130#issue-532777017)]
+
+Servers exposing client's access privileges on a resource URL MUST advertise
+by including the `WAC-Allow` HTTP header in the response of HTTP `HEAD` and
+`GET` requests.
+
+The syntax for `WAC-Allow`, using the ABNF syntax is:
+
+```
+WAC-Allow           = "WAC-Allow" ":" OWS permissions OWS
+permissions         = user-permissions OWS "," OWS public-permissions
+user-permissions    = "user=" quoted-access-modes
+public-permissions  = "public=" quoted-access-modes
+quoted-access-modes = DQUOTE *( access-modes / SP ) DQUOTE
+access-modes        = "read" / "write" / "append" / "control"
+OWS                 = *( SP / HTAB )
+```
+
+The `access-modes` corresponds to the modes of access (`acl:Read`,
+`acl:Write`, `acl:Append`, `acl:Control`) as defined in the ACL ontology.
+
+Clients can discover their access privileges on a resource by making an HTTP
+`HEAD` or `GET` request on the target URL, and checking the `WAC-Allow` HTTP
+header value for `user` and `public` paramaters listing the allowed access
+modes.
+
+[[Source](https://github.com/solid/specification/issues/171)]
+[[Source](https://github.com/solid/specification/issues/170)]
+[[Source](https://github.com/solid/specification/issues/181)]
--- a/main/authorization.bs
+++ b/main/authorization.bs
@@ -31,18 +31,17 @@ Section 1.2 of [[!RFC7231]], is:
 
 ```
 wac-allow           = "WAC-Allow" ":" OWS #access-param OWS
-access-param        = 1*ALPHA OWS "=" OWS quoted-access-modes
+access-param        = permission-group OWS "=" OWS quoted-access-modes
+permission-group    = 1*ALPHA
 quoted-access-modes = DQUOTE OWS *1(access-mode *(RWS access-mode)) OWS DQUOTE
 access-mode         = "read" / "write" / "append" / "control"
-OWS                 = *(SP / HTAB)
-RWS                 = 1*(SP / HTAB)
 ```
 
-The `WAC-Allow` HTTP header's field-value is a comma-separated list of access
-parameters (`access-param`). `access-param` is a whitespace-separated list of
-access modes ( `access-mode`) granted to a permission group.
+The `WAC-Allow` HTTP header's field-value is a comma-separated list of
+`access-param`s. `access-param` is a whitespace-separated list of `access
+mode`s granted to a `permission-group`.
 
-This specification defines the following permission groups:
+This specification defines the following `permission-group`s:
 
 <dl>
   <dt>`user`</dt>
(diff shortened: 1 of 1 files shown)


main/authorization.bs
```diff
@@ -19,3 +21,31 @@ the HTTP `Link` header with a `rel` value of `acl` [[#acl]].
...
+by including the `WAC-Allow` HTTP header in the response of HTTP `HEAD` and
+`GET` requests.
+
+The syntax for `WAC-Allow`, using the ABNF syntax is:
+
+```
+WAC-Allow           = "WAC-Allow" ":" OWS permissions OWS
+permissions         = user-permissions OWS "," OWS public-permissions
+user-permissions    = "user=" quoted-access-modes
+public-permissions  = "public=" quoted-access-modes
```
acoburn, line 35: What is the use case for including "public" access modes?

csarven: Thanks for raising that. I forgot to mention that as well in the comment! The PR reflects the current issues and clarifies *solid-spec*.

If access modes for `user` can be determined for the current (un)authenticated agent, I don't see the use for `public`. Is there a UI that's currently using it or needs to differentiate between them? Some weird edge case where `user` has less privileges than `public`?

`user` doesn't need to be tied to authenticated agent either. On that note, spec needs to be more clear on what the parameter(s) are intended for.

I'd be in favour of dropping `public`. @RubenVerborgh @dmitrizagidulin

csarven: That's a bit of a stretch no? Linking/referencing doesn't imply network access. One can be aware of the existence of an identifier, but nothing necessarily beyond that.

RubenVerborgh: The idea for this header comes from @timbl. The inclusion for `public` was to mimic Google-docs-like functionality: _what I write here, can it be seen and/or edited by the world?_
The unrestricted linking id indeed part of that.

Definitely nothing to do with `public` having less privileges than `user` (is that even possible in current WAC, BTW?).

main/authorization.bs
```diff
@@ -19,3 +21,31 @@ the HTTP `Link` header with a `rel` value of `acl` [[#acl]].
 In the event that a server can't apply an ACL to a resource, it MUST deny
 access.
 [[Source](https://github.com/solid/specification/issues/130#issue-532777017)]
+
+Servers exposing client's access privileges on a resource URL MUST advertise
+by including the `WAC-Allow` HTTP header in the response of HTTP `HEAD` and
```
RubenVerborgh, line 26: Don't we have an issue where `WAC-Allow` is under discussion?
I understand that this PR is just an officialized version of the draft spec, but as the author of the original text and library, I would like to reiterate my point from https://github.com/solid/specification/issues/170#issuecomment-523401269 that I am _not_ in favor of a custom header. I would rather see the `Link` header here.

Should we at least acknowledge in the spec that this point is under discussion?

main/authorization.bs
```diff
@@ -19,3 +21,31 @@ the HTTP `Link` header with a `rel` value of `acl` [[#acl]].
...
+`GET` requests.
+
+The syntax for `WAC-Allow`, using the ABNF syntax is:
+
+```
+WAC-Allow           = "WAC-Allow" ":" OWS permissions OWS
+permissions         = user-permissions OWS "," OWS public-permissions
```
RubenVerborgh, line 33: The syntax we had in mind is actually more generic, as you can see from the implementation: https://github.com/solid/wac-allow/blob/v1.0.0/src/WacAllowParser.js#L24

The `permissions` block is a comma-separated, unordered list of an arbitrary number of `permission` blocks. A `permission` block is `group = quoted-access-modes`, where `group` is a string of word characters and arbitrary whitespace is allowed around the equals sign.
It just so happens there are two pre-defined groups `user` and `public`.

See some of the tests as well: https://github.com/solid/wac-allow/blob/v1.0.0/__tests__/WacAllowParser-test.js

main/authorization.bs
```diff
@@ -19,3 +21,31 @@ the HTTP `Link` header with a `rel` value of `acl` [[#acl]].
...
+user-permissions    = "user=" quoted-access-modes
+public-permissions  = "public=" quoted-access-modes
+quoted-access-modes = DQUOTE *( access-modes / SP ) DQUOTE
```
RubenVerborgh, line 36: The original code allows for an arbitrary non-zero number of whitespace characters. Access modes MUST be separated by whitespace (not the case here).

main/authorization.bs
```diff
@@ -19,3 +21,31 @@ the HTTP `Link` header with a `rel` value of `acl` [[#acl]].
...
+access-modes        = "read" / "write" / "append" / "control"
```
RubenVerborgh, line 37: The original code is case-insensitive.

RubenVerborgh (changes requested):
Good start, but:
- not fully in line yet with the original intentions as documented in code (it's fine if the intentions change, just wanted to document this)
- does not explicitly mention doubts regarding this header

main/authorization.bs
csarven: Ack re `public` use.

The general use case/need was originally documented recorded in https://github.com/linkeddata/dokieli/issues/96 , https://github.com/solid/specification/issues/171 .. (2016). Not the specifics.

main/authorization.bs
csarven: Either is fine with me. As mentioned, I wasn't crazy about the custom header either.

Will implement whatever comes out of it in dokieli eg. whether to disable the "save" button or not.

bblfish:
If I am not mistaken this is needed because in the current spec one needs `Control` rights on a resource to see the ACL. If so this should be made clear.

I believe there has been a shift in perception in the Authorization panel in the past 3 months, to acknowledge that allowing ACLs to be readable by a larger group is not necessarily a bad thing, and can in fact be useful. For example it makes sense for resources that can be accessed after proof of payment, proof of age, etc... to be world readable.  It is also possible to have publicly visible ACLs that give access to restricted groups without revealing the membership of the groups, reducing concerns regarding confidentiality. 

Whatever one's position on this is, the exercise should be quite simple. Would this be needed where access to the ACL visible to clients or their authentication agents? If not that should be documented, as that helps motivate the header. If even with such access the header would be needed then that could provide extra support for it, that would also help understand the reason for it.

RubenVerborgh:
> If I am not mistaken this is needed because in the current spec one needs `Control` rights on a resource to see the ACL. If so this should be made clear.

Correct, and important indeed!




> I believe there has been a shift in perception in the Authorization panel in the past 3 months, to acknowledge that allowing ACLs to be readable by a larger group is not necessarily a bad thing

Instead of this header, we could link to (a summarized version of) the ACL.

acoburn:
> I believe there has been a shift in perception in the Authorization panel in the past 3 months, to acknowledge that allowing ACLs to be readable by a larger group is not necessarily a bad thing

For example, the Access Policy proposal makes it possible to distinguish between read and write in the context of an ACL. There are certainly cases where the ACL resource should be kept private and other cases where it's OK for it to be readable.

csarven:
I'm sure we are all on the same page here but just to be clear we had the two (read and write) contexts all along. For clients to be able to write to ACL was there from early on obviously. And for clients to be able read / know about allowed privileges perhaps became more apparent from use cases based on actual applications ( eg. https://github.com/linkeddata/dokieli/issues/137 , circa 2016 ). All of that predates WAC-Allow, ACP, authorization-panel... The functional requirements are more or less known and was agreed. We just didn't agree on the exact solution and carry far with the implementations. WAC-Allow just happens to be one approach to meet the read case - whether that came through by use cases recorded elsewhere doesn't matter at this point. Some aspects of WAC-Allow are attractive (to me) and some not. That's all fine though.. But this is not to say that the alternative approaches have been around for a long time, agreed, and in wide use across different implementations.

While we had rel=acl under WAC for the write case, we didn't "think outside of the box" or at least didn't get to proposing something like rel=readable-acl-for-requesting-agent referring to another ACL document that clients can read. Or bothered - ie. actually implement and propose for spec - to scratch the Link header approach.

There can be 50 different ways of addressing the same functional requirement. We just need one approach. Link header with rel allow is perfectly fine if that seems attractive now.. or having a separate ACL document just to read allowed privileges. That's all in a way "accidental" as far as I can see. Different times.. people.. in the mix.

I'd only suggest that we consider making the solution flexible enough to handle different access control systems. Meaning, along the lines of WAC-Allow's idea of using simple values eg. "read" implying `acl:Read` but it doesn't need to be `acl:`. So a bit generic like `solid:Read` (or whatever) to tell the client that it can do "read operations", or `solid:Write` for "write operations" and so on that corresponds to the operations as defined (and to be improved) in the Solid ecosystem. In the end something like:

`Link: <http://www.w3.org/ns/solid/terms#Read>; rel="http://www.w3.org/ns/solid/terms#allow"`

(I don't particularly care which namespace to use or what the term is actually called but probably close enough.)

So, in about 1.61803 years from now when we are reviewing a yet another access control system, we don't flip tables (again).

bblfish:
I am not sure we need different Link headers for read and write.
All I was suggesting is that the reasons for why the proposed feature is needed at this time should be documented. The "WAC-Allow" feature may provide a good stop-gap answer until a deeper one is provided.

main/authorization.bs
timbl: Yes @bblfish basically the public readability , like the user write frankly, is for modifying the options to users.    If the user can't write it, then the UI will not make the fields editable, will remove action buttons to change things, and suppress blank empty fields.  Maybe grey color text slightly.

If the thing is NOT viewable by the public, then his will suppress social actions which are public about it, like liking it, bookmarking it, "add this to my profile", etc etc, dropping it into a public chat.

If the thing  IS viewable by the public, and editable by the user, then using colors or fonts etc and actual warnings make it clear that anything the user types will be public. "People who can see this: 3 billion".

Those are the biggest uses I can think of ..There are others I am completely missing.

There is no reason that other non-WAC systems can't still just **supply a WAC-Allow header anyway**. Suggest they all be required to forSolid, whatever the ACL language they use. ("Control" would mean "ACP control read and write" for example in ACP

main/authorization.bs
csarven: The initial PR was based off the draft spec, not the implementation. Updated ABNF: https://github.com/solid/specification/commit/6d1bc9e1a90f63294e6bc38e6645665e6904121f . If I'm not mistaken, it can parse inputs like:
`WAC-Allow: ,user=" READ  wriTe   Append control ", public=" read append" ,other="read " , group="  ",,  `

main/authorization.bs
csarven: https://github.com/solid/specification/commit/6d1bc9e1a90f63294e6bc38e6645665e6904121f#diff-6e4282542b3e3fcc1e8048e34c85b6e806de95de91f6ae62d3273d47ec2eaeb6R35 should do it.

main/authorization.bs
csarven: These strings are case-insensitive as per RFC 5234.

main/authorization.bs
csarven: As an editor, I have indeed noted what you've said ie. not in favour of custom header and that the Link header may be a preferable alternative. However, I didn't think that it was significantly under discussion considering that there was a draft spec and implementations in place. To make better editorial calls, we need more feedback about the experience re libraries and applications. What experience can you share from the implementations? Is there an issue/PR for an alternative approach or implemented somewhere?

If we can go through similar steps for an alternative, as we did with WAC-Allow, we can raise that as a possible replacement (eg. Link header in https://github.com/solid/specification/pull/210#issuecomment-728265065 ). Alternatively, have a survey and see if the existing WAC-Allow implementations are stale, broken, or not actually used by applications. That could also justify an alternative.

csarven:
`WAC-Allow` header is now being used by dokieli to determine whether the "Save" button in the menu will be actionable or not (disables the HTML button). Brief detail and screenshots:

https://github.com/linkeddata/dokieli/issues/259#issuecomment-734766140

main/authorization.bs
```diff
@@ -19,3 +21,47 @@ the HTTP `Link` header with a `rel` value of `acl` [[#acl]].
...
+`GET` requests.
+
+The syntax for the `WAC-Allow` header, using the ABNF syntax defined in
+Section 1.2 of [[!RFC7231]], is:
+
+```
+wac-allow           = "WAC-Allow" ":" OWS #access-param OWS
+access-param        = permission-group OWS "=" OWS quoted-access-modes
+permission-group    = 1*ALPHA
+quoted-access-modes = DQUOTE OWS *1(access-mode *(RWS access-mode)) OWS DQUOTE
```
csarven, line 36: I don't have a strong opinion about requiring or making DQUOTE optional since some HTTP headers include parameters with double quotes for the value and some don't. Requiring it would be simpler indeed and I do have a preference for it.

*If* we change it to optional, the guidance for servers should be to use double quotes, and clients should be able parse with and without double quotes.

main/authorization.bs
RubenVerborgh: > To make better editorial calls, we need more feedback about the experience re libraries and applications. What experience can you share from the implementations?

I think this is tricky.

Yes, the implementation works and it was easy.

But it was easy because we designed and implemented it in an afternoon.

My issue is rather with it not being sustainable as an extension mechanism; we shoudn't keep on minting custom headerss.

main/authorization.bs
RubenVerborgh: Check, lets allow either.

RubenVerborgh (approved):

main/authorization.bs
csarven: Done. If anyone disagrees or would like to provide implementation feedback regarding optional quotes, please raise an issue.

main/authorization.bs
csarven: Unless there is a strong objection, we can go with the proposal for the time being. We can always revisit this when there is more information that we can consider.

csarven:
Thanks all. Let's go for it. Please follow up with issues / concerns / implementation feedback.

PR #216, opened 2021-01-07
Created automated push to site action

jaxoncreed:
Whenever someone pushes to master of the `/specification` repo, it will clone the repo's contents to `solid/solidproject.org` at the folder `/root/TR/`.

This will be a DIRECT COMMIT TO MASTER on `solid/solidproject.org`.

Note: merging this PR will begin this functionality.

### DIFF
--- a/.github/workflows/copy-to-website.yml
+++ b/.github/workflows/copy-to-website.yml
@@ -1,0 +1,19 @@
+name: Copy
+on: 
+  push:
+    branches:
+      - master
+jobs:
+  copy:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Copy
+      uses: andstor/copycat-action@v3
+      with:
+        personal_token: ${{ secrets.PERSONAL_TOKEN }}
+        src_path: /.
+        dst_path: /TR/
+        dst_owner: solid
+        dst_repo_name: solidproject.org
+        exclude: .github/*
+        clean: true


csarven:
Thank you for putting this together!

I'd like to make sure we understand/document a few things. Some of these may be obvious but..

* is this the final automated solution (given the current solidproject.org server setup)?
* to be clear, resources at `/root/TR/` will be allocated to `https://solidproject.org/TR/` ?
* can there be merge conflicts? if so, what's the (expected) process to resolve them?
* can users with write access to solid/solidproject.org repository (master branch) change the TR/ and what's deployed to solidproject.org/TR/ ?
* related to above, can writes to TR/ in solid/solidproject.org be restricted to the synchronisation activity made from solid/specification repo?

Anything else we should record here?

jaxoncreed:
- is this the final automated solution (given the current solidproject.org server setup)?
    - Yes
 - to be clear, resources at /root/TR/ will be allocated to https://solidproject.org/TR/ ?
    - As far as I know, yes. @justinwb set up the build system there, so he can confirm.
 - can there be merge conflicts? if so, what's the (expected) process to resolve them?
    - This script clears the TR folder before putting the documents back in. So there cannot be merge conflicts. If someone manually commits to the TR folder in `solidcommunity.org`, it will be overwritten the next time someone merges to master on `specification`
 - can users with write access to solid/solidproject.org repository (master branch) change the TR/ and what's deployed to solidproject.org/TR/?
    - Yes, but any changes they do will not be permanent. They will be overwritten with the next merge to master on `specification`
 - related to above, can writes to TR/ in solid/solidproject.org be restricted to the synchronisation activity made from solid/specification repo?
    - To my knowledge, you cannot restrict access to specific folders. Only the repository itself.

jaxoncreed:
One other thing to note. The commit must be done by a specific github account. Right now it's set up to commit in my name (all commits to solidcommunity.net are from the jaxoncreed account). If that isn't desirable, we need to generate a key for a github account that is desirable.

csarven:
Thank you @jaxoncreed ! Let's follow-up on the account making the commits to solidproject.org repo / branch in https://github.com/solid/solidproject.org/issues/505 .

It looks like this is working (after merge) with one issue: server appears to use lowercase `tr` as opposed to `TR` eg. `https://solidproject.org/TR/` redirects to `https://solidproject.org/tr/` . Uppercase,`TR` is canonical. If anything, `tr` should redirect to `TR`

jaxoncreed:
@csarven this is probably a problem on the side of the website build, not the automated push. Is it absolutely necessary to have a capital `TR` in the pathname?

csarven:
Yes, I get that it is a server. I only tagged on to this issue here. I can create a whole new issue exploring this under solid/solidproject.org if you like..

The path can be anything we agree on. It is necessary and important to be consistent about its use. Identifiers `TR`and `tr` identify different things as per RFC 3986, 7230.. Which is why the practice is to either 404 or 301 (to the canonical). See also https://gitter.im/solid/chat?at=6009a565d8bdab47394e2f10 .

The uppercase `TR` that I proposed is what was generally agreed on in Editor and Team meetings.

I have in fact went to the extent of documenting URI Templates ( https://github.com/solid/solidproject.org/issues/249 ) as a way to set a practice to record and register URIs on what's needed for the Solid project. That's about close to two months ago.

There is an issue in the pipeline for URI Persistence ( https://github.com/solid/solidproject.org/issues/247 ) for about same time frame and a whole PR on Terms of Use including a URI Persistence section ( https://github.com/solid/solidproject.org/pull/489 ) for two weeks - using uppercase `TR`.

So, now, is it absolutely necessary to use uppercase `TR`? Of course not.. who am I to say? Could it be tr or Tr or tR or TtttttRRRRrrrrzzz? Sure, why not. But should `TR` be used at this point? Most likely yes! If not, say so in the existing issues/PR.

Pardon me but I don't feel like writing / updating documentation or designing identifiers after whatever random server configuration is happened to be in use or in flavour today.

If there are technical barriers to use uppercase `TR` as canonical as well as redirecting requests coming to `tr` to `TR`, then let's have a look at that and address what's necessary. If not a redirect, I would at the very least expect a 404 on `tr`. Definitely not redirect `TR` to `tr`. FWIW, W3C gives 301. FWIW, NSS and CSS give 404. Cool? :)

Issue #177, opened 2020-04-27
Clarify the notion and mechanisms for server-managed information


csarven:
What are the conceptual and implementation differences between server-managed containment triples and the triples in server-managed resources (auxiliary)?

It appears to be that we have two different mechanisms in the spec that are conceptually similar, if not the same. How can we reconcile this?

Containers are self-descriptive in that the containment information is part of the resource description and managed by the server. Contained resources are tied to the lifecycle of the container. Container may include other information provided by clients. 

Server-managed resources (auxiliary) are independent resources discovered through primary resources and are tied to their lifecycle. Server-managed resources are not writeable by clients.

We can see that although cumbersome, containment information can be found in a server-managed resource (auxiliary). We can also see that any information that can part of a server-managed resource (auxiliary) can instead be in a self-descriptive resource.

justinwb:
> We can see that although cumbersome, containment information can be found in a server-managed resource (auxiliary). 

It's true that you _could_ store containment triples in a server managed auxiliary resource associated with a container, but I'm not sure that we _have_ to.

> We can also see that any information that can part of a server-managed resource (auxiliary) can instead be in a self-descriptive resource.

It is true that you could store the same information in a regular Solid resource as a server managed auxiliary resource, but the provenance would not be the same. This is the key value in the server managed auxiliary resource type. No agent would have the ability to directly write or modify the data in a server managed resource. A consumer of the data in a server managed resource can have confidence that the data inside it (i.e. timestamps, creator, etc) was written by the server and no-one else.

csarven:
The point of this issue is to note and justify why we have one mechanism where a server controls certain kind of information and yet another mechanism for other information. It seems to be a bit of an arbitrary split.

---

>but the provenance would not be the same.

Explain.

> No agent would have the ability to directly write or modify the data in a server managed resource.

Holds true for containment triples in the container resource that is managed by the server:

>Containment triples can't be updated by clients.

Clients affect resources directly and indirectly eg:

Client requests to append/remove a resource to/from a container. Only the server manages the containment relationships.

Client requests to create a resource. Only the server manages the creator relationship.

It is not apparent why a container (or any other resource for that matter) can't be self-describing especially when it comes to information like "timestamps, creator etc." It is already doing that for containments as well as other information. Aside: Some of the Solid servers exposed posix information in the container, and client have nothing to do with it.

[...]

Issue #116, opened 2019-11-07
Specify container listing mechanism


csarven:
* Should container descriptions omit statements about its resources from agents that are not privileged to read (`acl:Read`) the resource? ( See also https://github.com/solid/solid/issues/29 )
* Should container descriptions omit statements about acl and meta resources? ( See also https://github.com/solid/specification/issues/105 )
  * Should it be possible to reveal them when requested (eg. `Prefer` `include=...all` analogous to `ls -a`)?
* ..

ericprud:
Concrete example:
### <http://alice.pod/share/photoAnnotations/>
``` turtle
<http://alice.pod/share/photoAnnotations/> a ldp:BasicContainer ;
  ldp:contains <annot123>,  <annot456> .

# data cached from contained resources
<annot123> a photo:Annotation ;
  dc:author "Alice" ;
   photo:photo <http://amy.pod/photos/photo123.jpg>, 
   photo:caption "Alice and Amy at work" .

<annot456> a photo:Annotation ;
  dc:author "Alice" ;
   photo:photo <http://bob.pod/photos/photo456.jpg>, 
   photo:caption "Bob sleeping at work" .
```
Timeline:
1. Alice created the container
2. Alice allowed Bob to "contribute" (append and own his contributions)
3. Alice creates annot123 and annot456
 -- unprivileged GET on /share/photoAnnotations/ gets a everything in the container
4. Bob asks Alice to restrict access to the "Bob sleeping at work" photo.
5. Alice sets ACLs on annot456 to only allow Alice and Bob to see it.
 -- unprivileged GET on /share/photoAnnotations/ ideall doesn't get either the link to <annot456> or the triples
``` turtle
<annot456> a photo:Annotation ;
  dc:author "Alice" ;
   photo:photo <http://bob.pod/photos/photo456.jpg>, 
   photo:caption "Bob sleeping at work" .
```

csarven:
>* Should container descriptions omit statements about its resources from agents that are not privileged to read (`acl:Read`) the resource? ( See also https://github.com/solid/solid/issues/29 )

In the context of access controls, there are generally two cases to consider pertaining to the binding relationship between a LDP resource and the container it is in:

[...]

PR #228, opened 2021-02-05
Add security consideration about information exposure

csarven:
Some related issues in which this topic is touched upon:
* https://github.com/solid/specification/issues/28
* https://github.com/solid/specification/issues/44
* https://github.com/solid/specification/issues/116
* https://github.com/solid/specification/issues/177
* https://github.com/solid/specification/issues/227
* ..


protocol.html
```diff
@@ -895,6 +895,8 @@ <h3 property="schema:name">Security Considerations</h3>
...
 
                   <p>Data pods SHOULD use TLS connections to protect the contents of requests and responses from eavesdropping and modification by third parties. Unsecured TCP connections without TLS MAY be used in testing environments or when the data pod is behind a reverse proxy that terminates a secure connection.</p>
+
+                  <p>When a server responds to an authorized HTTP <code>GET</code> request, the response MUST NOT expose information about resources that the agent is unauthorized to read. For example, when a <code>GET</code> method request targets a container, the server MUST NOT include information besides containment statements about the contained resources in the response eg. last modification, size, type or label.</p>
```
justinwb, line 899:
```suggestion
                  <p>When a server responds to an authorized HTTP <code>GET</code> request, the response MUST NOT expose information about resources that the agent is unauthorized to read. For example, when a <code>GET</code> method request targets a container, the server MUST NOT include information beyond containment statements about the contained resources in the response. Last modification time, size, type or label are examples of what is not allowed in that scenario.</p>
```

Looks good - suggestion is aimed to ensure people don't misread and lump those examples (mod time, etc) in the "besides these" group.

csarven: Makes sense, thanks.

justinwb (approved):

RubenVerborgh (changes requested):
Responding to an “authorized” request should not return “unauthorized information”? The inverse would make sense to me, but this does not at the moment. Also the phrasing is too vague even if I’d agree with the content.

csarven:
@RubenVerborgh Retake: https://github.com/solid/specification/pull/228/commits/7ec822891cee4f56afa74475d094142cab24fa29 - Better?

---

The intention of the security consideration is for servers to not expose information beyond what can be controlled by an agent through access policies. Thus far the system left a gap in which this PR attempts to close. For example, the initial text "authorized" referred to the target resource of a request while may be granted, it is possible that the response can include other information that the agent is unauthorized for because of the way server generates a representation. This is especially the case when a server can potentially expose composite or aggregate information about other resources (than the target of the request) eg. description of contained resources or links to certain auxiliary resources (eg. to an ACL document). As it appears to be case in for example: https://github.com/solid/node-solid-server/issues/1567

protocol.html
```diff
@@ -896,7 +896,7 @@ <h3 property="schema:name">Security Considerations</h3>
...
 
-                  <p>When a server responds to an authorized HTTP <code>GET</code> request, the response MUST NOT expose unauthorized information about resources to the agent. For example, when a <code>GET</code> method request targets a container, the server MUST NOT include information beyond containment statements about the contained resources in the response. Last modification time, size, type, creator or label are examples of what is not allowed without proper authorization.</p>
+                  <p>Servers MUST NOT expose information more than the minimum amount necessary to power a feature. For example, when a <code>GET</code> method request targets a container, the server MUST NOT include information beyond containment statements about the contained resources in the response. Last modification time, size, type, creator or label are examples of what is not allowed without proper authorization.</p>
```
TallTed, line 899:
```suggestion
                  <p>Servers MUST NOT expose information beyond the minimum amount necessary to enable a feature. For example, when a <code>GET</code> method request targets a container, the server MUST NOT include information beyond containment statements about the contained resources in the response. Examples of what is not allowed without proper authorization include size, type, creator, label, and last modification time.</p>
```

protocol.html
```diff
@@ -895,6 +895,8 @@ <h3 property="schema:name">Security Considerations</h3>
...
 
                   <p>Data pods SHOULD use TLS connections to protect the contents of requests and responses from eavesdropping and modification by third parties. Unsecured TCP connections without TLS MAY be used in testing environments or when the data pod is behind a reverse proxy that terminates a secure connection.</p>
+
+                  <p>Servers MUST NOT expose information beyond the minimum amount necessary to enable a feature. For example, when a <code>GET</code> method request targets a container, the server MUST NOT include information beyond containment statements about the contained resources in the response. Examples of what is not allowed without proper authorization include size, type, creator, label, and last modification time.</p>
```
RubenVerborgh, line 899: I think this is non-enforceable as currently written.

- What is "information beyond the minimum amount necessary"?
  - Does that mean I cannot send out `X-Powered-By` headers?
    - I know this is not the intention, but there's nothing that tells me either way.
- The example is enforceable; but I think this should then be the requirement, or part of the set of requirements.
- What does "without proper authorization" mean?
  - Is it per resource?

Another concern is that checking this quickly becomes expensive.
A valid way to satisfy this requirement seems to not output these things ever, at all. But that will break applications.

RubenVerborgh (changes requested):

protocol.html
csarven: The purpose of the first statement is to set an understanding about server's behaviour ie. to preserve privacy by default. It is a consideration for that reason, and not something intended to be enforceable in and itself. Examples that follow will fill in what's exactly required / enforceable. We need to add these general or specific considerations as they come up.

This paves the way to writing up the "Security and Privacy Review" section (as per https://github.com/solid/specification/issues/147 ).

>What is "information beyond the minimum amount necessary"? 

https://www.w3.org/TR/security-privacy-questionnaire/#minimum-data

>Does that mean I cannot send out X-Powered-By headers?

If server considers X-Powered-By unnecessary (or risky) for unauthorized agents, it shouldn't expose it.

>The example is enforceable; but I think this should then be the requirement, or part of the set of requirements.

It would be a requirement ( https://solidproject.org/TR/protocol#conformance ). I'm open to moving text around later on - and something like the example could go under Containment or Representations or somewhere else... but, for the time being, it is useful to have these type of considerations under Security (or Privacy) where it can all be checked. After all, it is completely possible to move Considerations to other parts of the spec, but that's not necessarily best way of going at it.

>What does "without proper authorization" mean?

Privileges being acquired within established parameters, so "proper" eg. would be if you are authorized to read the contained resources via what the protocol offers, and did not obtain it by other means.

>Is it per resource?

In that example, yes (eg. read on container, but may not have read on some of the contained resources.)

>Another concern is that checking this quickly becomes expensive.

Sure, it depends on the system and whether it wants to expose information to unauthorized users. The proposal here is based off the foundations or principles. It doesn't render the system useless. The point is that if a system wants to make that information available, it should do so based on the constraints. We can throw away the generally-agreed principle(s) or make cases for why some information doesn't fall under this category - that's also fine.

>A valid way to satisfy this requirement seems to not output these things ever, at all.

Right, that'd be one way. But consider the issue https://github.com/solid/specification/issues/227 raising the use cases for the minimal/useful information for applications to know. It is orthogonal to the consideration here. Even if for example 227 resulted in requiring certain information to be part of the container representation or needing querying or paging or something, the Consideration is still useful.

>But that will break applications.

To be precise, that will "break" existing applications relying on implementation specific feature. We need to first address whether that feature is wanted and then how it will work: issue 227 etc.

---

Happy to revise the text but it would still be indefinitely open to improvements. I suggest that if the general proposal is agreed, we can edit through as we go. More information will come in so we'll also need to look at it as a whole.

d-a-v-i--: Improper authorization usually means a control failure or bypass. In other words you are meant to be prompted for authorization but maybe you found a way to avoid being prompted, such as a bug in consistency or a weakness. Sometimes it can mean implied, although more often because of auditability it's a formal one. CWE language is "incorrectly performs an authorization check" https://cwe.mitre.org/data/definitions/285.html

RubenVerborgh: > if the general proposal is agreed

I'm not happy with neither the general proposal nor the specific example, but I don't think I will ever be. I'll just lift my reject and let others work it out.

csarven:
Merged this as is for the time being. Will follow-up with improvements/corrections. For starters. the security consideration section needs to be revised into non-normative, and all normative text can move elsewhere.

PR #247, opened 2021-03-05
Adjusted copy to site script to look at main rather than master

jaxoncreed:


jaxoncreed:
I am unsure why this script wasn't even in the main branch to begin with. Did someone remove it?

jaxoncreed:
Not needed

PR #248, opened 2021-03-10
Allow ACL doc discovery without acl:Control

michielbdejong:
This is a spec change proposal.

### Practical Reason
Most current implementations of Solid allow ACL doc discovery without acl:Control.
For instance, NSS and CSS just add `.acl` and ESS adds `?ext=acl`.
It would be a lot of work (including data migration on all existing servers) to change that.

### Simplicity Reason
Using a predictable scheme makes the life of the server developer easier.
Also, reporting the `Link` header in for instance OPTIONS makes it easier
to discover if a server *supports* WAC at all, which is useful both to app
developers and to end-users of apps. You can then see "ah, there is an ACL
link but I don't have access", rather than being kept in the dark about it altogether.

### Reasons against it
* If the ACL doc URL is predictable, there is a real risk that app developers skip
the discovery and just guess it. However, now that different servers already
use different schemes, I don't think many app developers will do this anymore.

* If the ACL doc URL is predictable, it might give an attacker a leg up (help them
overcome the first barrier). The could then for instance try to brute-force their
way into the ACL doc. However, if they can brute-force their way into `acl:Control`
then they would maybe also be able to use that technique to break the barrier to
discovery.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -563,7 +563,7 @@ left:4.5em;
 
                       <p>Servers MUST NOT directly associate more than one ACL auxiliary resource to a subject resource.</p>
 
-                      <p>To discover, read, create, or modify an ACL auxiliary resource, an <code>acl:agent</code> MUST have <code>acl:Control</code> privileges per the <a href="https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm">ACL inheritance algorithm</a> on the resource directly associated with it.</p>
+                      <p>To read, create, or modify an ACL auxiliary resource, an <code>acl:agent</code> MUST have <code>acl:Control</code> privileges per the <a href="https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm">ACL inheritance algorithm</a> on the resource directly associated with it.</p>
 
                       <p>A Solid server SHOULD sanity check ACL auxiliary resources upon creation or update to restrict invalid changes, such as by performing shape validation against authorization statements therein.</p>
                     </div>
--- a/protocol.html
+++ b/protocol.html
@@ -565,6 +565,8 @@ left:4.5em;
 
                       <p>To read, create, or modify an ACL auxiliary resource, an <code>acl:agent</code> MUST have <code>acl:Control</code> privileges per the <a href="https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm">ACL inheritance algorithm</a> on the resource directly associated with it.</p>
 
+                      <p>Whether or not <code>acl:Control</code> is also required to *discover* an ACL auxiliary resource, is left up to the server implementer.</p>
+
                       <p>A Solid server SHOULD sanity check ACL auxiliary resources upon creation or update to restrict invalid changes, such as by performing shape validation against authorization statements therein.</p>
                     </div>
                   </section>


RubenVerborgh:
There is no guarantee whatsoever that the URL is predictable.
It might be that ESS mints private URLs for specific resource for all we know.

-1 for me

michielbdejong:
> There is no guarantee whatsoever that the URL is predictable.

How is that relevant? This is about whether the `Link` header should be hidden from sight or not for unauthenticated agents.

michielbdejong:
> It might be that ESS mints private URLs for specific resource for all we know.

That is a good point, and good to leave that open. So the spec should say that whether or not discovery of the ACL doc URL requires `acl:Control` is up to the server.

RubenVerborgh (changes requested):
> So the spec should say that whether or not discovery of the ACL doc URL requires `acl:Control` is up to the server.

It should not; the existing text is just fine. Discovery requires `Control` permissions.
The existence of another channel to discover doesn't change that.

elf-pavlik:
[ACP](https://github.com/solid/authorization-panel/blob/master/proposals/acp/index.md) doesn't use 'Control' and [Access Control Resource](https://github.com/solid/authorization-panel/blob/master/proposals/acp/definitions.md#access-control-resource) include explicit policies for accessing the ACR itself. In that case discovery probably would depend on those policies granting (or not) Requesting Party access to the ACR. Maybe it would make sense to have mini spec just for Solid-WAC integration and separate one for Solid-ACP integration?

csarven:
WAC is a way of setting Protocol's default behaviour and taking security considerations into account. If required, an explicit policy for accessing ACL resources can also be done.

bblfish:
I think the text is ok. 
I want to make a proposal for ACLs to have ACLs that when enabled would allow servers to make their ACLs readable to a wider group, even public if needed.

justinwb:
> So the spec should say that whether or not discovery of the ACL doc URL requires `acl:Control` is up to the server.

This shouldn't be left up to implementation, regardless of direction. 

> There is no "[Source]" link to the requirement. I know the requirement was mentioned/repeated in discussions but underlying reasons seems to be left out.

Since I did that submission - the motivation behind it starts with principle of least privilege - which is specifically relevant here given the role of the acl in the security posture of the system. To @csarven's point - it minimizes the attack surface, and is a general best practice we should continue to employ where we can. I think that as part of the drafting process we need to look for reasonable opportunities to improve security posture, even if existing implementations may need to be adjusted to conform. This should be done judiciously, but this seemed like a reasonable case (and still does).

csarven:
As far as I can tell the removal of the text "discover" from the specification is appropriate and consistent with the Protocol. I'm in approval of the initial commit made to this PR ( https://github.com/solid/specification/pull/248/commits/4639a5d77468b1ddab55f211ae107566b1834d25 ):

---

RFC3986 https://tools.ietf.org/html/rfc3986#section-7 states:

>7.  Security Considerations

>   A URI does not in itself pose a security threat.

---

Introducing a security consideration for servers using a deterministic algorithm to generate a URI string does not fundamentally address the perceived concern on discovering the name of access control resources. Even if servers were to omit the link relation, it is trivial to look up the algorithm generating the URI string, especially in the case of open-source servers, and so server implementations and pods using them can be targeted en masse.

It is also worthwhile to consider scenarios where the access control source can be obtained in other ways eg. agent used to having Control access or having prior knowledge about the algorithm generating the string. (RFC 3986 blurb)

The only rule that can universally address this is by requiring servers to follow a certain URI Template for access control URIs eg. "Servers MUST use a non-deterministic (unguessable) algorithm to generate the URI string when allocating it to access control resources."

This adds unnecessary complexity to the system. Knowing an identifier != access. Similar to having Read access to a container and observing the containment statements does not entail having access to the contained resources. (RFC 3986 blurb)

The Solid Protocol must remain neutral and flexible about this by not requiring how servers name their resources. (Please do not bring up slash semantics into this discussion!)

Access control resources of WAC was never classified as "secret" - no documentation suggests this. Implementations simply followed this practice and at the very least did not deem them to be secret or a security threat. (RFC 3986 blurb)

---

Again, where needed, explicit policies controlling the discoverability of access control resources through HTTP headers (or anywhere) can still be prescribed/achieved per mechanism.

bblfish:
I agree 
 1. with the original removal of "discover" from the first paragraph, since it could be guessable in any case, and making it unguessable is a lot of work for nothing: we all know that someone can edit some acl for a resource. 
 2. I am not against addition of the second paragraph.
 
The specification should state what has to be, not what should not be, as that would be closing possibilities that are unknown. The same with tests. For example the proposal [ACLs on ACLs for WAC](https://github.com/solid/authorization-panel/issues/189) would make tests that just check for `:Control` Access mode wrong on systems that allow ACLs to be read more widely.

protocol.html
```diff
@@ -563,7 +563,9 @@ <h4 property="schema:name">Web Access Control</h4>
 
                       <p>Servers MUST NOT directly associate more than one ACL auxiliary resource to a subject resource.</p>
 
-                      <p>To discover, read, create, or modify an ACL auxiliary resource, an <code>acl:agent</code> MUST have <code>acl:Control</code> privileges per the <a href="https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm">ACL inheritance algorithm</a> on the resource directly associated with it.</p>
+                      <p>To read, create, or modify an ACL auxiliary resource, an <code>acl:agent</code> MUST have <code>acl:Control</code> privileges per the <a href="https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm">ACL inheritance algorithm</a> on the resource directly associated with it.</p>
+
+                      <p>Whether or not <code>acl:Control</code> is also required to *discover* an ACL auxiliary resource, is left up to the server implementer.</p>
```
TallTed, line 568:
```suggestion
                      <p>Whether <code>acl:Control</code> is required to <em>discover</em> an ACL auxiliary resource, is left to the implementation.</p>
```

bblfish:
I have [added a detailed example](https://github.com/solid/authorization-panel/issues/189#issuecomment-797638239) on how ACLs on ACLs could allow protected resources to have public ACLs without leaking information. The example shows how Alice may find it useful to know that she has access to `event` resource as she is a member of a closed group. 

What that shows is that acl discovery should put in terms not of `acl:Control` but in terms of the visibility of the ACL. If the ACL is only visible to members who have control then we are in the current situation. But for ACLs that are publicly readable, the ACL should be publicly discoverable. There are many intermediate possibilities too.

bblfish:
I opened up a complimentary PR that goes a little further than this one, taking into account use cases from the Authorization panel https://github.com/solid/specification/pull/250

michielbdejong:
Continued in #252

michielbdejong:
We can now consider this issue as resolved. A client needs (at most) `acl:Read`, not `acl:Control`, to see the `rel="acl"` link response header on a resource.

bblfish:
> We can now consider this issue as resolved. A client needs (at most) `acl:Read`, not `acl:Control`, to see the `rel="acl"` link response header on a resource.

I agree. Is that written out somewhere? It seems I have only seen text being removed.

PR #250, opened 2021-03-15
reduce requirements on acl:Control to edit and create

bblfish:
The ACL ontology states:

```Turtle
  acl:Control     a :Class;
         :comment "Allows read/write access to the ACL for the resource(s)";
         :label "control"@en;
         :subClassOf acl:Access .
```

This clearly gives read/write access to the controller. But that is different from disallowing read access to anyone else, which is what the removed text stated. 

There are use cases for making ACLs readable in order to guarantee [privacy of the client](https://solid.github.io/authorization-panel/authorization-ucr/#uc-minimalcredentials). There are ways to extend ACL so that ACLs become readable, see [ACLs on ACLs](https://github.com/solid/authorization-panel/issues/189), and this is also needed by ACP. 
Having the default be as it currently is specified in WAC that ACLs be only visible to those who control them, when no other information is available, is a default choice that makes sense. But closing doors to valid use cases that would allow such extensions is not justifiable.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -563,7 +563,7 @@ left:4.5em;
 
                       <p>Servers MUST NOT directly associate more than one ACL auxiliary resource to a subject resource.</p>
 
-                      <p>To modify an ACL auxiliary resource, an <code>acl:agent</code> MUST have <code>acl:Control</code> privileges per the <a href="https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm">ACL inheritance algorithm</a> on the resource directly associated with it.</p>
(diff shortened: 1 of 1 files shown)


bblfish:
In order to cover discoverability I would be fine with text such as the following:

```
In order to allow a client that has read access to an acl associated with a resource to discover it, a server MUST place a `Link: rel="acl"...` header in the response.
```

gibsonf1:
My vote is also to have default behavior be that an acl is an acl of itself unless the acl has its own acl.  (This is how our current acl algorithm is operating)

bblfish:
@gibsonf1 I am in favor of having ACLs of ACLs :-) It would be useful to open a separate PR for that. I am not sure where though. Here I am just trying to remove text that could make it seem like that behavior was not legal.

csarven:
Thanks for highlighting the problematic text and a clarification for it. Closing this PR as agreed in favour of https://github.com/solid/specification/pull/252

PR #254, opened 2021-04-09
Add server Cache-Control requirement forcing revalidation

csarven:
The requirement in this PR is a finding that was brought forward by @acoburn @Vinnl and co based on implementation experience. I hope it is captured accurately but this is why we have reviews :)

The requirement here supports browsers caching resources locally, but when re-fetching them, it will send conditional requests (If-None-Match, etc). A 304 Not Modified response allows the local browser cache to use the cached data, so the browser can avoid the network costs of re-transferring the data. If, however, the resource has changed, the new representation will be transferred, which is what an app would expect.

As per the Protocol, clients MAY implement RFC 7232. So, this is definitely something they can take advantage of -- and arguably shouldn't miss out out. The hard requirement (MUST) is set on servers so that all clients can take advantage of an intelligent caching strategy.

Feedback including direct experience with client/server behaviour can help a great detail. Some of the details involved here may be subtle and not always immediately apparent to server implementers... considering walking the fine balance between RFC7272, RFC7234.. and implementations in the wild.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -413,6 +413,8 @@ left:4.5em;
                   <p>A data pod MUST implement the server part of <cite>HTTP/1.1 Authentication</cite> [<cite><a class="bibref" href="#bib-rfc7235">RFC7235</a></cite>]. When a client does not provide valid credentials when requesting a resource that requires it (see <a href="#webid">WebID</a>), the data pod MUST send a response with a <code>401</code> status code (unless <code>404</code> is preferred for security reasons).</p>
 
                   <p>A Solid server MUST reject <code>PUT</code>, <code>POST</code> and <code>PATCH</code> requests without the <code>Content-Type</code> header with a status code of <code>400</code>. [<a href="https://github.com/solid/specification/issues/70#issuecomment-547924171" rel="cito:citesAsSourceDocument">Source</a>]</p>
+
+                  <p>When a server includes the <code>ETag</code> and/or <code>Last-Modified</code> headers, the server MUST include the <code>Cache-Control</code> header with the <code>no-cache</code> (alternative: <code>must-revalidate, max-age=0</code>) directive. In addition, for resources fetched with authentication, the server MUST include the <code>private</code> directive.</code>
                 </div>
               </section>
 


protocol.html
```diff
@@ -937,7 +939,7 @@ <h3 property="schema:name">Accessibility Considerations</h3>
...
+                  <p id="authoring-tool-accessibility"><strong>Authoring Tool Accessibility</strong>: To contribute to the proliferation of Web content that is accessible to a broad range of people, it is strongly encouraged to follow the <cite><a href="https://www.w3.org/TR/ATAG20/" rel="cito:citesAsAuthority">Authoring Tool Accessibility Guidelines</a></cite> [<cite><a class="bibref" href="#bib-atag20" title="Authoring Tool Accessibility Guidelines (ATAG) 2.0">ATAG20</a></cite>] in the design of authoring tools to support the production of accessible content through accessible user interfaces.</p>
```
kjetilk, line 942: It seems like two changes sneaked in here, the latter should probably be a different PR?

csarven: This change ("Update link") doesn't need to be in its own PR.. but I've pushed that commit into main just now. So, if/when we merge this PR, it'll only be the commit that's intended for this PR that'll be added to main,.. then things should magically work out, right, right?

If you have a better/simple solution, let me know. I'm no git-ninja.

kjetilk (commented):
It sounds like something is broken here... It should be possible to use all three caching strategies (conditional requests, presetting a ttl, notifications of changes). This seems to me to prohibit the two latter if the former is used. That would be bad.

Say that you have a chat where each day ends up in its own container, effectively making the previous days messages immutable. You'd be able to set a very long `max-age` on those resources, but you'd also assign them an `ETag` because why not. :-)

With this, it sounds like you can't do that, you'd have to opt for one of them. Conditional requests are nice, because it reduces network traffic, but a decent `max-age` is golden, because it eliminates the need to do a request altogether. Given that performance in decentralised systems is hard and caching is a very important strategy, we need to be very careful to get it right.

I would like to hear more about the background. :-)

protocol.html
kjetilk: OK! It could be done prettier, I think, but I can go with that :-)

acoburn:
> but a decent max-age is golden, because it eliminates the need to do a request altogether

very true!

The question comes down to what kinds of expectations a server can have about resources. If there is a way for the server to know that certain resources are no longer going to change, then a longer `max-age` property would be appropriate. But if anyone may write to a resource at any time, then a higher `max-age` value results in local data being out of date. For certain apps that may be fine; for others it may not be. Generally, we have found that serving stale data from a local cache just confuses users.

kjetilk:
> > but a decent max-age is golden, because it eliminates the need to do a request altogether
> 
> very true!
> 
> The question comes down to what kinds of expectations a server can have about resources. If there is a way for the server to know that certain resources are no longer going to change, then a longer `max-age` property would be appropriate. But if anyone may write to a resource at any time, then a higher `max-age` value results in local data being out of date. For certain apps that may be fine; for others it may not be. Generally, we have found that serving stale data from a local cache just confuses users.

Yes, that's very true. Stale data is very bad. To achieve decentralization, it falls on us to solve this problem, through good UIs where human users can explicitly indicate their expectations; through enabling organizations that have data maintenance and curation systems that has that kind of information; through machine learning estimating expectations; through the mentioned example of a chat program where certain resources effectively becomes immutable, etc.

What I'm concerned with here is that this proposal seems to cut off this direction, which I think is very important. I know that we don't see that much now, but it is, I believe, one of the things we have to be able to solve, the decentralization vision really depends on it.

csarven:
Point taken re MUST, and I agree. Long term stuff that's pledged to be immutable -- along the lines of Memento resources. FWIW, there is the proposal on "immutable" directive as Cache Control extension: https://bugzilla.mozilla.org/show_bug.cgi?id=1267474 , https://mailarchive.ietf.org/arch/msg/httpbisa/6gS9zGCh4tIB3hKa67wsoHdb4gY/

kjetilk:
Anyway, perhaps to add to that, it seems that you (@acoburn and @Vinnl ) had some practical experience where the `Cache-Control` header had to be set, otherwise the revalidation didn't work? It sounds like behavior that wouldn't be in line with the specs, but if this is Chrome acting up, then yes, we might have to do something to mitigate that problem despite my decentralization idealism :-)

acoburn:
I think @kjetilk's use cases, particularly the setting of a TTL value, are very interesting. It would be unfortunate if this change made those impossible to implement.

I wonder if a better approach here might be along the lines of "When a server includes the ETag and/or Last-Modified headers, the server MUST include the Cache-Control header" ... but stop there and leave the details of the _value_ of Cache-Control up to the implementation. IOW the specific directives could be implementation decisions: an impl might translate a certain RDF property into the max-age directive value. Some non-normative text could go a long ways toward guiding implementers in a good direction without expressly prohibiting certain advanced use cases.

Vinnl:
Would it not be possible to add a caveat like "for any resource for which the server cannot guarantee that it will be immutable for a particular length of time, it must..."? That leaves the door open for whatever mechanisms arise that lead to immutability (Memento, client apps that indicate expectations, etc.), and I don't suppose you'd want clients to skip checking for updates for resources that might have changed?

kjetilk:
I think we need to be clearer on what is a distinct Solid problem. 

As far as I can tell, there isn't much. Stale responses are a problem for everyone, that content with an `Authorization` header field can't be cached is also common. Those cases are already considered in RFC7234. 

The main different thing for Solid, IMHO, is that the stakes are higher, because the use of cached data, even if it is just for seconds, are very important for getting sufficient performance in query answering in decentralized ecosystems. It is also very important when you're drawing data from many sources.

If clients are not RFC compliant in a way that causes stale data to be used, then we certainly need to consider that. Normal RFC compliant behavior is what you propose, @Vinnl , it is part of the normal freshness and validation cycle. 

There are many nuances to caching, like what caches and clients should do if the server doesn't respond. The RFC leaves a lot of wiggle room there. We could certainly be much stricter than the RFC is, but should we? Is it conceivable that some clients operate in environments with poor internet connection, and would therefore need to have this wiggle room to use stale data if they lost connection? I think so... If we require `must-revalidate`, then we cut off the possibility to have Solid clients in environments with intermittent connections, like rapidly moving vehicles.

I started the caching panel to discuss such topics, I think we should discuss this at some length.

Vinnl:
@kjetilk The thing that's specific to Solid is that _every_ client-managed resource can be stale the moment it's been fetched (e.g. because it's been updated by a different user, the same user on a different device, or because it's a Container and a child resource has been created).

RFC7234, as I understand it (which is admittedly limited, since I'm basing myself off of a couple of paragraphs on MDN), specifies how a server can share the knowledge it has about how the resource will evolve with the client, which can then make its own decisions based on that.

So what this addition would be saying is that, if a Solid server communicates staleness data, it should communicate that the data can theoretically change at any point - because it can. It does not prevent clients from using cached data anyway if their specific situation calls for it and in fact, that's the default behaviour of client libraries like [SWR](https://swr.vercel.app/), so AFAICS, that wiggle room remains.

kjetilk:
> @kjetilk The thing that's specific to Solid is that _every_ client-managed resource can be stale the moment it's been fetched (e.g. because it's been updated by a different user, the same user on a different device, or because it's a Container and a child resource has been created).

I gotta disagree. I don't think that is specific to Solid, not even to the Web. BBSes had that property long before the Web, and when we got forum software on the Web that became commonplace issue. 

> RFC7234, as I understand it (which is admittedly limited, since I'm basing myself off of a couple of paragraphs on MDN), specifies how a server can share the knowledge it has about how the resource will evolve with the client, which can then make its own decisions based on that.
>
> So what this addition would be saying is that, if a Solid server communicates staleness data, it should communicate that the data _can_ theoretically change at any point - because it does. 

Yes, it does, but the whole HTTP stack of technologies are designed on that assumption, that any resource can change at any point and it does. This is the assumption that goes without RFC7234. Then, the question is, since the client might already have a copy, how may it use it?

Then, conditional requests are added in RFC7232 to allow for the client to ask the server if the copy it has can be used again. RFC7234 is added so that the server can tell the client for how long the client might use the copy without asking again, thus eliminating the need for a network roundtrip. RFC7234 also adds methods that tell the client it cannot use the data.


> It does not prevent clients from using cached data anyway if their specific situation calls for it and in fact, that's the default behaviour of client libraries like [SWR](https://swr.vercel.app/), so AFAICS, that wiggle room remains.

OK, there are some things to keep apart here, without knowing how SWR concrete does this beyond RFC5861. 

The patch in this issue adds `no-cache`. `no-cache` is a very strict prohibition on caching, RFC7234 says:
> The "no-cache" request directive indicates that a cache MUST NOT use a stored response to satisfy the request without successful validation on the origin server.

So, the use of `no-cache` does prevent clients from using cached data without a network roundtrip. That is the very purpose of using it. Then you can add the directives from RFC5861 to deal with stale data, but the more recent RFC7234 has some general guidelines there, so it is probably not that important anymore.

The normal operation of a cache is that if no `max-age` or similar is set, it won't reuse a response without validation. If `max-age` or similar is set, it will keep using the response until the freshness time is reached, then it will validate. 

So, I'm not sure what you are proposing, if it is that the client libraries you use are not behaving that way, or if it is that we shouldn't let servers make a freshness commitment into the future, or if it is something else. If it is the first case, we obviously have to look into it, if it is the second then I'm strongly opposed to it as it would destroy any hope of creating a decentralized query system. 

For that reason, I also don't see why just requiring a `Cache-Control` header would make any difference, as RFC7232 and RFC7234 covers rather different ground, and validation is what would happen anyway.

kjetilk (changes requested):
I'll just mark my opposition as a changes request, see other comments for discussion :-)

Vinnl:
> I gotta disagree. I don't think that is specific to Solid, not even to the Web. BBSes had that property long before the Web, and when we got forum software on the Web that became commonplace issue.

Sure, so I would argue that it is _also_ something specific to BBS's, i.e. if your server serves BBS resources, it should set the same header, and clients should attempt to revalidate if possible. And if we were defining a standard for BBS's, we'd also have to add such a provision.

Or to flip this around: if this were a standard for generic HTTP servers, i.e. we would not know what kind of resources they would be serving, we could not mandate such a Cache-Control setting - after all, the server might not be serving resources that can change at any time, so it might want to be more strict. But in the case of Solid, we _do_ know that the resources can change at any time, and so it seems that the server should not be telling clients that they cannot.

> The normal operation of a cache is that if no max-age or similar is set, it won't reuse a response without validation. If max-age or similar is set, it will keep using the response until the freshness time is reached, then it will validate.

I think this is the part that we experienced not to be the case, which triggered this addition. This is described [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_freshness_checking):

> If an origin server does not explicitly specify freshness (e.g. using Cache-Control [with a max-age - Vincent] or Expires header) then a heuristic approach may be used.
>
> In this case look for a Last-Modified header. If this header is present, then the cache's freshness lifetime is equal to the value of the Date header minus the value of the Last-modified header divided by 10.

So this means that if a browser requests a resource that hasn't been modified in the last 100 minutes, then it _will_ reuse_ the response without validation in the next ten minutes.

Thus, if you want the behaviour you describe, i.e. "it won't reuse a response without validation," then you have to also add `max-age: 0`, which supposedly is equivalent to `no-cache`.

Maybe it's a good idea to explicitly describe the behaviour we were observing when `max-age` was not set to 0: you'd do e.g. a POST to a container to add a new contained resource, immediately followed by a GET on that container. That GET would then not revalidate, leading to the previous response body being used, and the new resource not being listed.

kjetilk:
> > The normal operation of a cache is that if no max-age or similar is set, it won't reuse a response without validation. If max-age or similar is set, it will keep using the response until the freshness time is reached, then it will validate.
> 
> I think this is the part that we experienced not to be the case, which triggered this addition. This is described [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#heuristic_freshness_checking):
> 
> > If an origin server does not explicitly specify freshness (e.g. using Cache-Control [with a max-age - Vincent] or Expires header) then a heuristic approach may be used.
> > In this case look for a Last-Modified header. If this header is present, then the cache's freshness lifetime is equal to the value of the Date header minus the value of the Last-modified header divided by 10.
> 
> So this means that if a browser requests a resource that hasn't been modified in the last 100 minutes, then it _will_ reuse_ the response without validation in the next ten minutes.

OK, I see! But heuristic freshness is not a server-side measure, it is a client/proxy side thing. It is a terrible, terrible idea to have a caching client that does heuristic freshness without any influence from the programmer. 


> 
> Thus, if you want the behaviour you describe, i.e. "it won't reuse a response without validation," then you have to also add `max-age: 0`, which supposedly is equivalent to `no-cache`.

That is clearly not the intention of the RFC. This is a matter of a broken client library.

> Maybe it's a good idea to explicitly describe the behaviour we were observing when `max-age` was not set to 0: you'd do e.g. a POST to a container to add a new contained resource, immediately followed by a GET on that container. That GET would then not revalidate, leading to the previous response body being used, and the new resource not being listed.

Indeed, and that's why this is a broken client library. But we cannot break Solid because of a broken client library, there has to be some way for the programmer to tell the library not to do such silly things. 

The client library really should have heuristic freshness off by default, anything else is a terrible idea. Are you sure you at least can't tell the library to not use it?

Vinnl:
> It is a terrible, terrible idea to have a caching client that does heuristic freshness without any influence from the programmer.

That might be, but I'm afraid that ship has sailed:

> That is clearly not the intention of the RFC. This is a matter of a broken client library.

Please take note that this is not client library behaviour, but browser behaviour. This is what browsers do by default. Although it is [explicitly allowed by RFC 7234](https://datatracker.ietf.org/doc/html/rfc7234#section-4.2.2), as pointed out by the link I shared earlier.

kjetilk:
We need to consider this more carefully nevertheless. The ramifications for Solid are massive, it will make decentralized query answering extremely difficult, and without that, Solid is DoA.

Heuristic freshness was intended as a suggestion and the 10% was an example of what you might do, not something you should do blindly. There has to be a better way here. We should take some time to investigate.

kjetilk:
But I do see your suggestion of just adding `Cache-Control` in a different light now. Lets discuss when we have more time.

Vinnl:
> The ramifications for Solid are massive, it will make decentralized query answering extremely difficult

Could you elaborate on that? AFAICS the header is still just a suggestion, so if a query system wants to risk stale data and cache it anyway, could it not still do so?

> Heuristic freshness was intended as a suggestion and the 10% was an example of what you might do, not something you should do blindly.

That's something you'll have to take up with the browser vendors :)

kjetilk:
> > The ramifications for Solid are massive, it will make decentralized query answering extremely difficult
> 
> Could you elaborate on that?

Time is a bit limited now, I'd love to at some point.

> AFAICS the header is still just a suggestion, so if a query system wants to risk stale data and cache it anyway, could it not still do so?

No, `no-cache`  and `max-age` are MUSTs. Also, we need the whole Web ecosystem, including CDNs and stuff like that to pull this off, and so we need the standards to be clear.

Obivously, it would be nice if there was a `Cache-Control` directive for the server to say that the client/proxy may not use heuristics. I have never looked for that, I can't remember seeing it (`must-revalidate` means that you can't use a stale response even in the case of intermittent connectivity). It would solve this problem. Requiring everyone to do their damnedest to give meaningful `max-age`s would be an even better solution, but off-hand, I'm not sure how that would turn out in practice.

> 
> > Heuristic freshness was intended as a suggestion and the 10% was an example of what you might do, not something you should do blindly.
> 
> That's something you'll have to take up with the browser vendors :)

I guess so ;-) At least, we'd need to assess the impact of that problem.

Vinnl:
> Obivously, it would be nice if there was a Cache-Control directive for the server to say that the client/proxy may not use heuristics.

The client can only avoid heuristics if the server provides it with actual information it can base itself of instead, no? Which is what I understand this proposal to be: instructing the server to tell the client what guarantees it can give it about the freshness of the data (which is that it can have changed right after GETing it, so 0 is the most meaningful max-age it can give for mutable resources).

I think we can all agree that the following should not happen:

- A resource just created with a POST is not listed when doing a GET on its container.
- A server that notifies an app that a resource has changed then reports that resource to be unchanged when the app does a GET to that resource.
- A change that has been applied to a resource on one device is not visible when another device GETs that resource.

I see no way to avoid those without having the client getting in touch with the server, but if there are alternative approaches to avoiding them, it would be great if they could be shared here.

But until then, every server will want to apply the change in this PR, or the above _will_ happen. And if that is the case, then is at the very least a non-normative suggestion not an obvious addition?

kjetilk:
> > Obivously, it would be nice if there was a Cache-Control directive for the server to say that the client/proxy may not use heuristics.
> 
> The client can only avoid heuristics if the server provides it with actual information it can base itself of instead, no?

That is the question I don't feel I have the full answer to yet. :-) Can it really be that caching mechanisms in browsers are *that* broken? 

So, what I'm saying is that we need to set aside time to investigate how widespread this problem is, and what ways there are to mitigate that problem. 


> I think we can all agree that the following should not happen:
> 
>     * A resource just created with a POST is not listed when doing a GET on its container.
> 
>     * A server that notifies an app that a resource has changed then reports that resource to be unchanged when the app does a GET to that resource.
> 
>     * A change that has been applied to a resource on one device is not visible when another device GETs that resource.

Certainly!
 
> 
> I see no way to avoid those without having the client getting in touch with the server, but if there are alternative approaches to avoiding them, it would be great if they could be shared here.
>
> But until then, every server will want to apply the change in this PR, or the above _will_ happen. 

That's a fairly categorical statement, you know! :-) There are a few things that I'd like to investigate. You said that you had looked into what FF does, but how about the other browsers? Is there nothing you can do on the client side to disable heuristics? Can we find something in the top-level RFCs that could help us in other ways? Beyond the top-level RFCs, like RFC723*, it is actually fairly hard to get an overview of every RFC that could be relevant, there's a lot of them that browsers may have implemented that can be of use? 

> And if that is the case, then is at the very least a non-normative suggestion not an obvious addition?

The impact of such a decision is very large. It will be very difficult to scale Solid beyond two-three clients and two-three data sources, and that's why the Web today is centralized into a few silos. Everything is easier in centralized architectures, you have none of these worries. So, the reason why I'm resisting is that I worry about decentralization in the future. 

Consider a case where you have a few hundred actors, each contributing their own data. This could be humans having a conversation, gaming together, gathering for protests, resolving traffic jams (which is my favorite example of when humans engage in swarm behaviour without having swarm intelligence), or doing business at some scale. It could also be IoT, where you have a sensor network under the control of different actors coming together to produce a result. Your application wants to gather that information, analyze and present that to other actors somehow. To do this, you would need to make a lot of requests, and say that you could get a conditional request down to 50 ms or something in near foreseeable future. However, you could need to make hundreds of such requests, adding up to a user-perceived latency of seconds. With a cache, if the data changes on a certain timescale, this is achievable.

I'm not saying that `max-age` is the only thing we have to achieve this, to the contrary, there's a lot of opportunity, but it is pretty much what we have *now*. There are many cases where notifications are better, but we'd very quickly go into cases where unicast isn't enough. 

Without this, we will very quickly get into a situation where there is a massive, centralizing force. For example, for an application to be able to do anything, everyone has to put their data into somebody else's pod. If that happens, Solid will have achieved very little, we'd have given up on personal data control and we'd have given up on redecentralizing the Web.

So, that's what I think are at stake here :-) 

I think it is pretty important that we find time to work on the things that are necessary to ensure that we can decentralize, but I also understand that you'd like answers right now. So, I just hope we can set aside time to understand this in sufficient depth.

Vinnl:
> That's a fairly categorical statement, you know! 

It is. Is it not true?

> You said that you had looked into what FF does, but how about the other browsers? 

IIRC, Chrome appeared to use a different heuristic that caused it to skip the cache entirely, even without validating it.

> Is there nothing you can do on the client side to disable heuristics?

What would you want the client to do? If there's no clear-cut instruction of what to do, the client will have to make a decision somehow. I don't see how you could get the client not to use heuristics without telling it what rules it _should_ adhere to.

> Can we find something in the top-level RFCs that could help us in other ways?

As mentioned before, AFAIK it's literally impossible, by definition, for a client to ensure data is fresh without asking the server.

I understand the desire for a cache to avoid many requests, but that cache will have to make a conscious choice to live with the risk that the data it caches is outdated. I don't see why the server cannot tell it that it runs that risk in the first place - the cache is then free to ignore the server instructions and run that risk.

It's up to the spec team, but I can almost guarantee that servers will either be going to implement the behaviour described here (potentially after spending many hours trying to figure this out by themselves), or users will be experiencing the bugs I described above. And in practice, it's probably going to be the latter first, and then the former. That doesn't seem like something you should want, but up to you :shrug:

kjetilk:
> > That's a fairly categorical statement, you know!
> 
> It is. Is it not true?

I don't know. But in that, you say that you are 100% certain that there is no way to ensure that the heuristics are not used, and you have performed and exhaustive search of all RFCs, all 10000 of them, and you have verified that there isn't for example a `no-heuristics` directive, and you have verified that no browsers have implemented something like that. 

Isn't that a pretty bold claim?

> 
> > You said that you had looked into what FF does, but how about the other browsers?
> 
> IIRC, Chrome appeared to use a different heuristic that caused it to skip the cache entirely, even without validating it.

Hmmm, isn't that what you would derive from the RFC?

The expected behavior from the RFC is simply that if there is no `Cache-Control`, you don't cache it. If there is a `max-age`, then you cache it for as long as it is valid, you may validate it afterwards, or you may silently expire it. You may expire it before the `max-age` too, if it clutters up your cache. The heuristic that is proposed is something that you may do if the circumstances make it likely.

All I'm saying is that we need to investigate this. Preferably urgently. We can't just merge this without understanding the scope of the problem and what possible solutions there is.

The patch is most certainly not the solution. The solution may be a use of `most-revalidate`, or a requirement to always have a `max-age`, but it may also be something completely different. We need to assess that.

Vinnl:
> I don't know. But in that, you say that you are 100% certain that there is no way to ensure that the heuristics are not used

No, I am saying that servers espousing the behaviour that this change tries to steer them away from will result in at least one commonly used browser experiencing the undesirable effects I mentioned above :) However, I _am_ 100% certain there _is_ a way to ensure the heuristics are not used: that's what the proposed change would effect.

> Hmmm, isn't that what you would derive from the RFC?
> 
> The expected behavior from the RFC is simply that if there is no Cache-Control, you don't cache it, but you may if there is a conditional request that can be used. 

Yes, I think both browsers are behaving according to the RFC. If the server doesn't indicate the freshness of the resource, they are free to make a guess as to whether it's fresh and thus whether to serve it from cache: they can decide it's not and fetch it anew (like Chrome), or that it is and to retrieve it from their cache (like Firefox). But since the mutable resources you encounter in Solid are _not_ fresh after fetching them, the latter decision is incorrect and thus has the undesirable effects I mentioned above.

kjetilk:
> > We can't just merge this without understanding the scope of the problem and what possible solutions there are.
> 
> That's what I'm trying to help with by sharing what we saw during implementation :)

Yeah, and thank you for that! But there are still many dimensions that we haven't covered, so we need more. We clearly have to address this for version 1, so we'd need to get to it sooner rather than later.

kjetilk:
We resolved to close this issue and continue a discussion in #269 when we get there.

PR #261, opened 2021-05-05
Break lines at column 100

kjetilk:
Uhm... I wrote something on nearly every line in the spec ;-)

Basically, I prefer shorter lines when working with diffs because shorter lines means that less text is could be involved, which makes it far easier to review, you don't have to scroll sideways. Also, it makes it easier to see where the changes are, because a change is likely to touch more lines, rather than being on one line. It also makes the HTML easier to read when working on it. IMHO, all that of course. 

What do others think?

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -158,62 +158,107 @@ left:4.5em;
 
         <dl id="document-identifier">
           <dt>This version</dt>
-          <dd><a href="https://solidproject.org/TR/protocol" rel="owl:sameAs doap:homepage">https://solidproject.org/TR/protocol</a></dd>
+          <dd><a href="https://solidproject.org/TR/protocol" rel="owl:sameAs
+          doap:homepage">https://solidproject.org/TR/protocol</a></dd>
         </dl>
 
         <div id="authors">
           <dl id="author-name">
             <dt>Editors</dt>
-            <dd id="Sarven-Capadisli"><span about="" rel="schema:creator schema:author"><span about="https://csarven.ca/#i" typeof="schema:Person"><a rel="schema:url" href="https://csarven.ca/"><span about="https://csarven.ca/#i" property="schema:name"><span property="schema:givenName">Sarven</span> <span property="schema:familyName">Capadisli</span></span></a></span></span></dd>
+            <dd id="Sarven-Capadisli"><span about="" rel="schema:creator
+            schema:author"><span about="https://csarven.ca/#i"
+            typeof="schema:Person"><a rel="schema:url"
+            href="https://csarven.ca/"><span about="https://csarven.ca/#i"
+            property="schema:name"><span property="schema:givenName">Sarven</span> <span property="schema:familyName">Capadisli</span></span></a></span></span></dd>
 
-            <dd id="Tim-Berners-Lee"><span about="" rel="schema:author"><span about="https://www.w3.org/People/Berners-Lee/card#i" typeof="schema:Person"><a rel="schema:url" href="https://www.w3.org/People/Berners-Lee/"><span about="https://www.w3.org/People/Berners-Lee/card#i" property="schema:name"><span property="schema:givenName">Tim</span> <span property="schema:familyName">Berners-Lee</span></span></a></span></span></dd>
+            <dd id="Tim-Berners-Lee"><span about=""
+            rel="schema:author"><span about="https://www.w3.org/People/Berners-Lee/card#i"
+            typeof="schema:Person"><a rel="schema:url"
+            href="https://www.w3.org/People/Berners-Lee/"><span about="https://www.w3.org/People/Berners-Lee/card#i"
+            property="schema:name"><span property="schema:givenName">Tim</span> <span property="schema:familyName">Berners-Lee</span></span></a></span></span></dd>
 
-            <dd id="Ruben-Verborgh"><span about="" rel="schema:author"><span about="https://ruben.verborgh.org/profile/#me" typeof="schema:Person"><a rel="schema:url" href="https://ruben.verborgh.org/"><span about="https://ruben.verborgh.org/profile/#me" property="schema:name"><span property="schema:givenName">Ruben</span> <span property="schema:familyName">Verborgh</span></span></a></span></span></dd>
+            <dd id="Ruben-Verborgh"><span about=""
+            rel="schema:author"><span about="https://ruben.verborgh.org/profile/#me"
+            typeof="schema:Person"><a rel="schema:url"
+            href="https://ruben.verborgh.org/"><span about="https://ruben.verborgh.org/profile/#me"
+            property="schema:name"><span property="schema:givenName">Ruben</span> <span property="schema:familyName">Verborgh</span></span></a></span></span></dd>
 
-            <dd id="Kjetil-Kjernsmo"><span about="" rel="schema:author"><span about="http://www.kjetil.kjernsmo.net/foaf#me" typeof="schema:Person"><a rel="schema:url" href="http://www.kjetil.kjernsmo.net/"><span about="http://www.kjetil.kjernsmo.net/foaf#me" property="schema:name"><span property="schema:givenName">Kjetil</span> <span property="schema:familyName">Kjernsmo</span></span></a></span></span></dd>
+            <dd id="Kjetil-Kjernsmo"><span about=""
+            rel="schema:author"><span about="http://www.kjetil.kjernsmo.net/foaf#me"
+            typeof="schema:Person"><a rel="schema:url"
+            href="http://www.kjetil.kjernsmo.net/"><span about="http://www.kjetil.kjernsmo.net/foaf#me"
+            property="schema:name"><span property="schema:givenName">Kjetil</span> <span property="schema:familyName">Kjernsmo</span></span></a></span></span></dd>
 
-            <dd id="Justin-Bingham"><span about="" rel="schema:author"><span about="https://justin.bingham.id/#me" typeof="schema:Person"><a rel="schema:url" href="https://justin.bingham.id/"><span about="https://justin.bingham.id/#me" property="schema:name"><span property="schema:givenName">Justin</span> <span property="schema:familyName">Bingham</span></span></a></span></span></dd>
+            <dd id="Justin-Bingham"><span about=""
+            rel="schema:author"><span about="https://justin.bingham.id/#me"
+            typeof="schema:Person"><a rel="schema:url"
+            href="https://justin.bingham.id/"><span about="https://justin.bingham.id/#me"
+            property="schema:name"><span property="schema:givenName">Justin</span> <span property="schema:familyName">Bingham</span></span></a></span></span></dd>
 
-            <dd id="Dmitri-Zagidulin"><span about="" rel="schema:author"><span about="http://computingjoy.com/" typeof="schema:Person"><a rel="schema:url" href="http://computingjoy.com/"><span about="http://computingjoy.com/" property="schema:name"><span property="schema:givenName">Dmitri</span> <span property="schema:familyName">Zagidulin</span></span></a></span></span></dd>
+            <dd id="Dmitri-Zagidulin"><span about=""
+            rel="schema:author"><span about="http://computingjoy.com/"
+            typeof="schema:Person"><a rel="schema:url"
+            href="http://computingjoy.com/"><span about="http://computingjoy.com/"
+            property="schema:name"><span property="schema:givenName">Dmitri</span> <span property="schema:familyName">Zagidulin</span></span></a></span></span></dd>
           </dl>
         </div>
 
         <dl id="document-published">
           <dt>Published</dt>
-          <dd><time content="2020-12-16T00:00:00Z" datatype="xsd:dateTime" datetime="2020-12-16T00:00:00Z" property="schema:datePublished">2020-12-16</time></dd>
+          <dd><time content="2020-12-16T00:00:00Z" datatype="xsd:dateTime"
+          datetime="2020-12-16T00:00:00Z" property="schema:datePublished">2020-12-16</time></dd>
         </dl>
 
         <dl id="document-modified">
           <dt>Modified</dt>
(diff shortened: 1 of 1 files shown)


acoburn:
Big :+1: on breaking lines at a sensible length. 100 seems like a good number.

csarven:
Sorry, -1 from me on this PR. Not "reviewing" as per GitHub's settings because I don't wish to request changes.

* I find it more difficult to read HTML with arbitrary line breaks regardless of line length. Literally breaks my flow.
* Line breaks are information in and themselves - I find it especially concerning for text content eg. sentence level.
* Line breaks need not be preserved as part of the source data in order to satisfy the needs of applications. Remember that whole thing on not coupling data with application?
* Line breaks are preserved when source (HTML+RDFa) is reserialized to other concrete RDF syntaxes. (We have a lot more on the way eg. requirements)

I use the following git config settings for diffing which you might also find handy:

```
[color]
	ui = true
[alias]
        wdiff = diff --word-diff
        wlog = log --word-diff
```

In case of use to anyone, I find meld ( https://meldmerge.org/ ) indispensable. Kjetil: https://packages.debian.org/sid/meld

Besides... shouldn't we be using/working towards a Solid application of sorts... to author/annotate documents or something? :)

kjetilk:
> Sorry, -1 from me on this PR. Not "reviewing" as per GitHub's settings because I don't wish to request changes.
> 
>     * I find it more difficult to read HTML with arbitrary line breaks regardless of line length. Literally breaks my flow.

OK, I realize that you are the most important user of these files, so I don't want to pressure you to accept something that breaks your flow. But if you could live with it, then I think it would be nice for others, myself included. :-)

>     * Line breaks are information in and themselves - I find it especially concerning for text content eg. sentence level.

Yeah, if it had been a universal solution that would break on sentence level instead, that would be great. I personally don't see the information value of line breaks, and HTML doesn't either.

>     * Line breaks need not be preserved as part of the source data in order to satisfy the needs of applications. Remember that whole thing on not coupling data with application?

Absolutely! But the reality isn't there. Especially, if you look at this PR in Github's patch viewer, you'll see that the tool is practically unusable with these long lines. You'd have to scroll sideways back and forth. Which could explain why people have been put off from doing more extensive reviews. There's a threshold too high when the most immediate tool you have doesn't work.

> 
>     * Line breaks are preserved when source (HTML+RDFa) is reserialized to other concrete RDF syntaxes. (We have a lot more on the way eg. requirements)

Mmmmm, right. I can see that problem... But then, we could solve that by parsing those as XMLLiterals... 

> 
> I use the following git config settings for diffing which you might also find handy:
> 
> ```
> [color]
> 	ui = true
> [alias]
>         wdiff = diff --word-diff
>         wlog = log --word-diff
> ```

Thanks, I'll stuff that in my config.

> In case of use to anyone, I find meld ( https://meldmerge.org/ ) indispensable. Kjetil: https://packages.debian.org/sid/meld

OK, great, I installed that. But even with that, there's a workflow around reviewing, and Github has tools for reviewing that doesn't work so well.

> Besides... shouldn't we be using/working towards a Solid application of sorts... to author/annotate documents or something? :)

We should! It'll be a gargantuan task, so I don't know who could even get started doing such a massive thing! :-)

csarven:
>Which could explain why people have been put off from doing more extensive reviews.

hahaha, no! Excuse me. If that was remotely true or even significant, 1) the concerns would've been likely brought up long ago - it wasn't, 2) we had fairly consistent line breaks prior to switching to static HTML and reviews were still lagging and had poor quality - more like social reactions (with just thumbs up and stuff - literally *anyone* can do that) as opposed to actual proper technical reviews.. at this stage of the game. So, no, I completely reject that explanation that it had anything to do with line breaks or whatsoever :)

pmcb55:
I don't have full context here, so just my 2c - but if any text (be it Java code, JavaScript code, HTML, JSDoc, etc.) is to be reviewed in GitHub at all, then I find sensible line breaking is really important. Personally, I'd argue for breaking at 80 (and not 100 (but I can live with 100, I suppose)), just so that I can review on my phone without the 'really annoying' need to scroll left and right. Reviews (and 'good' reviews) are critical, and so any impediments at all that make that even slighter harder should be removed, if we justifiably can - e.g., I do, however, see problems with long RDF string literals, e.g., for Turtle (as I understand it), it doesn't support multi-line string literals that **_don't_** include the line-breaks in the literal itself.
For example, I can have long, multi-line RDF string literals:
```turtle
<> rdfs:comment """Some long textual description
for humans to read, but that if serialized as a multi-line
string will include all these line-breaks that I'm including here...""" .
```

...but this is encoding the line-breaks, which I _generally_ don't want hard-coded into the literal at all. What we would need to help GitHub reviewing would be something like this instead:
```non-turtle
<> rdfs:comment "Some long textual description" +
" for humans to read, but that if serialized as a multi-line" +
" string will include all these line-breaks that I'm including here.." .
```
...but we can't do that ('cos Turtle doesn't support that (or does it?)).
So can we live with RDF string literals being the only exception to the "forced line-break at 100" rule...!?

kjetilk:
> I don't have full context here, so just my 2c - but if any text (be it Java code, JavaScript code, HTML, JSDoc, etc.) is to be reviewed in GitHub at all, then I find sensible line breaking is really important. Personally, I'd argue for breaking at 80 (and not 100 (but I can live with 100, I suppose)),

Yeah, I had configured at 72 and then at 80 for many years, but with larger screens, I felt that for HTML, it made sense to increase to 100.

> just so that I can review on my phone without the 'really annoying' need to scroll left and right. Reviews (and 'good' reviews) are critical, and so any impediments at all that make that even slighter harder should be removed, if we justifiably can - e.g., I do, however, see problems with long RDF string literals, e.g., for Turtle (as I understand it), it doesn't support multi-line string literals that **_don't_** include the line-breaks in the literal itself.
> For example, I can have long, multi-line RDF string literals:
> 
> ```turtle
> <> rdfs:comment """Some long textual description
> for humans to read, but that if serialized as a multi-line
> string will include all these line-breaks that I'm including here...""" .
> ```
> 
> ...but this is encoding the line-breaks, which I _generally_ don't want hard-coded into the literal at all.

Yeah, that's a tension between the XML world, where whitespace+ is insignificant, and Turtle, where it is. So, we could bring in the XML world by using the `rdf:XMLLiteral` data type...

> So can we live with RDF string literals being the only exception to the "forced line-break at 100" rule...!?

I think @csarven 's point, correct me if I'm wrong, is that the Protocol spec, indeed anything that Sarven is lead editor of, is RDFa, because Sarven is an RDFa wizard, and so all the text in these specs are RDF string literals. That has a lot of strengths, but it also adds some restrictions.

In the end, we need a solution that @csarven can live with here. I just wanted to propose it as I feel it would be helpful, and since it was easy to not only propose, but also provide a PR, I did it (Emacs FTW!)

csarven:
I think you mean rdf:HTML in this case instead of rdf:XMLLiteral. The document already includes rdf:HTML but that doesn't change the situation with line breaks... it just helps to preserve the markup in the literal and parsing continues for new statements.

Again, my main concern is about the line breaks in text and attribute nodes. Whitespace between element nodes is a non-issue. So, as Pat suggests, we can compromise on forcing line break after a closing tag for lines that go over 100 or whatever. If MUST have :) Not sure how much that's really going to help reviewing through GitHub UI, but ok..

At the moment, the RDFa in the Protocol spec is pretty basic as far as information coverage. Soon it'll be along the lines of https://www.w3.org/TR/ldn/ ( `curl "http://rdf.greggkellogg.net/distiller?command=serialize&url=https:%2F%2Fwww.w3.org%2FTR%2Fldn%2F&format=rdfa&output_format=ntriples&raw"` ) and actually even better... as some of you already know/expect.

What shall we debate about next? 2, 3, 8 spaces or tabs? Favourite text editor? The answer is obviously...

csarven:
meh.. I mean, it gets complicated. Consider `<p>foo bar.. <em>baz</em> qux</p>`. I'd rather want line break before `<em>`. Okay to break After `</p>` I suppose but even then. Generally safe-enough around block-level elements,.. and that brings us back to the current pattern that I've applying any way.

kjetilk:
> The answer is obviously...

Emacs! ;-)

Shall we just say that we try out breaking lines manually in reasonable places to help reviewers using Github?

csarven:
As mentioned (twice) above... here is what the N-Triples serialization would be for the example above (snippet):

```turtle
<https://csarven.ca/#i> <http://schema.org/name> "\n                      Sarven\n                      Capadisli\n                    "@en .
```

Undisputedly fugly.

I ack having preferably shorter lines for the purpose of reviews/PRs, however for the published version, text nodes must not split. It may even be a marginal improvement to omit whitespace in the published version - but that introduces additional steps/machinery to the whole process... just to get a plain ol' HTML output.

kjetilk:
> As mentioned (twice) above... here is what the N-Triples serialization would be for the example above (snippet):
> 
> ```turtle
> <https://csarven.ca/#i> <http://schema.org/name> "\n                      Sarven\n                      Capadisli\n                    "@en .
> ```
> 
> Indisputably unwanted.

Absolutely. I just fail to understand where the problem really is. I mean, somewhere in this process, there has been an interpretation of whitespace as having a different significance than it has in HTML and XML. What is it  that makes this misinterpretation? Is it tooling? Is it an RDF* spec? (with RDF* I mean as something in the RDF ecosystem, which is what Tim intended in the Linked Data design issue back in the day). Can the problem be mitigated by being explicit about the parse type or something?

BTW, @pmcb55 , I think that error would mean that somewhere, there is a `<wbr foo="bar"></wbr>`, when it should be `<wbr foo="bar"/>`

pmcb55:
Ok, so just to close out my input on this (and more as an educational note to my future self, and perhaps anyone else as unfamiliar with RDFa as I am):

This HTML+RDFa:
```html
<dd id="Sarven-Capadisli"><span about="" rel="schema:creator schema:author"><span about="https://csarven.ca/#i" typeof="schema:Person"><a rel="schema:url" href="https://csarven.ca/"><span about="https://csarven.ca/#i" property="schema:name"><span property="schema:givenName">Sarven</span> <span property="schema:familyName">Capadisli</span></span></a></span></span></dd>
```
...is parsed into this Turtle (I used [this](https://www.w3.org/2012/pyRdfa/Overview.html#distill_by_input) to translate):
```turtle
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix schema: <http://schema.org/> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<> schema:author <https://csarven.ca/#i> ;
    schema:creator <https://csarven.ca/#i> .

<https://csarven.ca/#i> a schema:Person ;
    schema:familyName "Capadisli" ;
    schema:givenName "Sarven" ;
    schema:name "Sarven Capadisli" ;
    schema:url <https://csarven.ca/> .
```

Whereas this identical HTML+RDFa, which is merely nicely formatted (from a HTML perspective that is, **_not_** from an RDFa perspective!):
```html
<dd id="Sarven-Capadisli">
  <span about="" rel="schema:creator schema:author">
    <span about="https://csarven.ca/#i" typeof="schema:Person">
      <a rel="schema:url" href="https://csarven.ca/">
        <span about="https://csarven.ca/#i" property="schema:name">
          <span property="schema:givenName">Sarven</span>
          <span property="schema:familyName">Capadisli</span>
        </span>
      </a>
    </span>
  </span>
</dd>
```
...is parsed into this Turtle:
```turtle
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix schema: <http://schema.org/> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<> schema:author <https://csarven.ca/#i> ;
    schema:creator <https://csarven.ca/#i> .

<https://csarven.ca/#i> a schema:Person ;
    schema:familyName "Capadisli" ;
    schema:givenName "Sarven" ;
    schema:name """
                      Sarven
                      Capadisli
                    """ ;
    schema:url <https://csarven.ca/> .
```
So that's RDFa for you (I guess) - game over for trying to make it readable (and therefore maintainable, I'd argue). I had a quick look, but it seems there's no RDFa property to help tell parsers to explicitly ignore inter-HTML-tag whitespace whenever you might want, or need to. That seems a tad nuts to me, but that's just based on this simple example, so I could well be missing loads of complexity if such a simple property where to be attempted.

(As a complete aside, I suppose RDFa isn't completely alone here, as even our venerable Turtle has no way (that I know) to add 'soft' line breaks in RDF literal strings either (as I mentioned in my first comment on this issue). But with the extremely popular RDF-star necessitating the need to propose ['star' updates to the popular RDF serializations](https://w3c.github.io/rdf-star/cg-spec/editors_draft.html) (although that draft only refers to Turtle,TriG, N-Quads, and N-Triples, and **_not_** RDFa), would you guys think that that significant update might offer an opportunity to sneak in a soft line-break syntax option into Turtle-star/TriG-star at least...?)

csarven:
If one writes a person's name with line breaks in Turtle, then "game over..."

If one really cares about "readable" diffs, lets just commit to writing the spec in N-Triples. Who wants to go first? :)

kjetilk:
> I had a quick look, but it seems there's no RDFa property to help tell parsers to explicitly ignore inter-HTML-tag whitespace whenever you might want, or need to. That seems a tad nuts to me, but that's just based on this simple example, so I could well be missing loads of complexity if such a simple property where to be attempted.

Yes, I agree. That's really strange, so strange that I don't understand nobody has thought of it already... Is there something we're missing?

> would you guys think that that significant update might offer an opportunity to sneak in a soft line-break syntax option into Turtle-star/TriG-star at least...?)

I think it is worth trying :-) , one of us just need to have sufficient bandwidth to take it up in the appropriate forum.

kjetilk:
I guess this isn't happening.

PR #263, opened 2021-05-12
Make resource creation more exact and testable

RubenVerborgh:
I found the language to be inaccurate (`PUT and PATCH requests`), but also not testable.
There is no way to test the desires of a client; but we can test the behavior of the server.
So, in my opinion, it's the server that `MUST`, not the client.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -592,9 +592,9 @@ left:4.5em;
                 <div datatype="rdf:HTML" property="schema:description">
                   <p>When creating new resources, servers can determine an effective request URI’s type by examining the URI path ending (<a href="#uri-slash-semantics">URI Slash Semantics</a>).</p>
 
-                  <p>Clients who want to assign a URI to a resource, MUST use <code>PUT</code> and <code>PATCH</code> requests.</p>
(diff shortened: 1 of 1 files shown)


csarven (changes requested):
I think mention of client can be omitted altogether. Also focus on effective request URI.

Changing the requirement's focus (from clients) to servers is perhaps most useful for PATCH because it changes resource creation from MAY (RFC5789) to MUST. For PUT and POST, the requirement is actually not needed because it is already covered by RFC 7231.

RubenVerborgh:
@csarven Agree with all.

csarven:
Should we still guide (informative/note) the clients along the lines of:

Clients can use PUT and PATCH requests to assign a URI to a resource. Clients can use POST requests to have the server assign a URI to a resource.

csarven:
Added commit. Have a look. Interesting that GitHub doesn't allow me to request a review from you (sort of obviously).

RubenVerborgh:
Thanks, @csarven. Approved!

csarven (approved):

Issue #197, opened 2019-06-23
Which permissions are required for deleting a resource?


RubenVerborgh:
When deleting a resource R in a container C, do we require:
- Write permissions on R?
- Write permissions on C?
- Both?
- Something else?

RubenVerborgh:
+1 from me

michielbdejong:
Yeah, I guess both options make sense, but your argument for changing this makes sense. If you have write on R then you can already set it to zero bytes which is almost the same as deleting it. Since it reduces the required permissions, it will not break many clients. But this is a semver-major change, so I'll change the behaviour in NSS 6.

I hadn't added any tests for this in https://github.com/solid/web-access-control-tests yet, so I'll immediately create the tests to test the new behaviour.

acoburn:
+1 on _not_ requiring `acl:Write` on the container in order to allow deletion of a contained resource (assuming the agent has `acl:Write` permission on the contained resource).

justinwb:
Also a very strong +1 to not requiring `acl:Write` on the container to delete a contained resource R when the agent has acl:Write on R.

> By not requiring Write on container to delete a contained resource, agents are not granted more access than what they need. Granting an agent with Write on container allows them to alter the container description as well as potentially modify or delete other (but not necessarily all) resources in the container.

Agree with @csarven and echo these points 1000% 👆

csarven:
Great! Updated based on most recent consensus: https://github.com/solid/specification/commit/4fd701fcd451bd7a24c5a930d940a56297788991

timbl:
Here is a counter argument to keep the need for Write access to the container also.   We have at the moment an invariant property of the system is that **if I remove Write access to something it won't change**.  

I like that.  It's logical. For example, if I am reading a file which has no write (or append) access then I  can cache it on a much more serious basis.   If I don't give anyone write access to a folder, I can assume, like for a file, that it won't change. If you then delete one of its contents, then that breaks - I thought I had a valid copy of the list, and I don't.

One of the directions we have talked about was getting into immutable resources.  An immutable resource is one where is is not only unwritable now but will never be writable in the future.   That's a fairly short step. With immutable resources, we  can refer to them by their hash. They can be Merkle trees,  you can  stick them in IPFS and do lots of other things.  You can send them to the Internet Archive as a final version of this resource. 

Lots of things get optimizable when something can't change.  You don't need to lock access to it at all. 

Being able to tie these good things to the ACL only having Read access is something we would lose.

[...]

Issue #153, opened 2020-03-13
Should it be possible to find controller of storage via resources?


megoth:
(I think I've created an issue before that's close to this one, but I'm unable to find it right now - sorry if this is a duplicate, but think it's important to bump the issue for the specification editors to discuss.)

In the current data browser we have [a function that gets the profile that is "the owner" of the Pod](https://github.com/solid/solid-panes/blob/master/src/dashboard/homepage.ts#L55-L64), i.e. the profile that is located in `(Pod)/profile/card#me`. This is hardcoded against the current implementation of Node Solid Server, and will break if NSS changes or on any other Solid Server implementation that don't follow this pattern.

Currently there is no way of detecting the "owner" (in lack of a better word) of a Pod from a given resource. (Or possibly "owners", if multiple WebIDs can be said to be owners of the given Pod.) I think this can be useful, as it allows us to give more context to the current Pod that someone is visiting, at least when visitig the "root of the Pod" (another term I know might be problematic, hence the air quotes).

It might be problematic for people to expose their ownership of a Pod, so if it should be part of the standard, we might want to make it opt-in. (The Pod Provider would still have access to the info, but accessing the WebID through resources, or just the root, might be hidden for everyone else by default.)

elf-pavlik:
> I think this can be useful, as it allows us to give more context to the current Pod that someone is visiting, at least when visitig the "root of the Pod" (another term I know might be problematic, hence the air quotes).

I recall conversation about root of the storage in https://github.com/solid/data-interoperability-panel/issues/10

Following [terminology from auth draft](https://github.com/solid/authentication-panel/blob/master/oidc-authentication.md#17-terminology) *resource controller(s)* of the "root" in a solid storage would very likely set very restrictive ACL on that "level". If someone arrives to that "storage root" IRI for example following *space:storage* predicate in someone else's WebID Profile, they would most likely get HTTP 403 due to mentioned restrictive ACL on that resource/container.
For users who do have at least read access to the storage root, I think one could add statement to that container using *space:storage* predicate the same way as we use it in WebID Profile.

RubenVerborgh:
The bigger blocker for this is that, strictly speaking, we don't model the concept of a "pod" in the Solid tech stack. E.g., you won't find predicates like "hasPod".
It's the same on the Web, really. On the Web layer, there are only linked documents. The notion of "server" disappears in the layers below.

[...]

Issue #67, opened 2019-09-24
Loss of Control Mitigation


kjetilk:
As noted in #58 , we need a strategy to address the situation where a user is locked out from their pod through errors in the ACL. This can occur in several situation, e.g. syntax error in ACL files, that users pass end up not granting control permission to any agent, etc.

We have server-side validation to mitigate the former, and we have discussed introducing the concept of an agent that is understood to be the "owner", to mitigate other situations. We should have a more general discussion on it.

dmitrizagidulin:
In addition to missing or malformed ACL resources (_especially_ the root storage ACL), loss of control (user being "locked out" of their account or storage space) can also occur from the deletion of several important resources:

* Deletion of the WebID Profile document (prevents the user from being able to authenticate)
* Modification of the WebID Profile doc to remove a few key triples (location of storage, and authentication-related triples)
* Deletion of the root `.meta` resource. (This one's weird, but we currently store the recovery email for a user's account in the root `/.meta` resource. Not sure how that came about, but we should probably change that.) - renders the user unable to change their password etc.

kjetilk:
@justinwb has somet thoughts on how to use meta resources (#102) to address this.

csarven:
re critical changes to the WebID Profile description:

>Modification of the WebID Profile doc to remove a few key triples (location of storage, and authentication-related triples)

We may need the notion of "update-only properties" ie. clients are only allowed to update the object of a statement, but never remove the statement or the subject-predicate pair(s).

"update-only properties" is clearly different than than how "server-managed properties" are handled as well as what "write-once properties" (as LDP calls it).

gibsonf1:
Concerns as a Pod service provider:

It should be impossible for the user to compromise their account functionality in any way with the exception of an intentional account deletion.

To that end, there are several triples that the system requires and user should not be able to modify in any way.  Some examples are:
```
<https://frederick.trinpod.us/@>  
    solid:account <https://frederick.trinpod.us/> ;
    solid:oidcIssuer "https://trinpod.us"^^<xsd:string> ;
    solid:privateTypeIndex frederick:t_72 ;
    solid:publicTypeIndex frederick:t_6x ;
    space:preferencesFile frederick:t_8d ;
    space:storage <https://frederick.trinpod.us/> ...
```
I think state level permissions are needed to solve problems like system vs user data on the same resource, where a state is a uri representing a triple. 

[...]

PR #264, opened 2021-05-19
Add notion of "owner" and requirements

csarven:
Hello Reviewers,

In response to the following (and possibly other) issues on the notion of "owner":

* https://github.com/solid/specification/issues/67
* https://github.com/solid/specification/issues/153
* https://github.com/solid/specification/issues/197

If describing the relationship between Solid data pod / storage "owner" and AWWW's "URI ownership" is required, it can be done in another PR.

On merge of this PR, a follow-up PR will be made to the https://github.com/solid/vocab/ repository requesting to add and define the term "owner" in the Solid Terms vocabulary: `http://www.w3.org/ns/solid/terms#owner`. Proposal (derived from http://www.w3.org/ns/auth/acl#owner ):

```turtle
solid:owner
    a rdf:Property ;
    dc:issued "2021-05-19"^^xsd:date ;
    rdfs:comment "The person or other agent which owns the data storage. For example, the owner has implicit control of a data storage."@en ;
    rdfs:isDefinedBy <http://www.w3.org/ns/solid/terms#> ;
    rdfs:range foaf:Agent ;
    rdfs:label "owner"@en .
```

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -334,6 +334,8 @@ left:4.5em;
 
                   <p>A <dfn class="dfn-paneled" data-dfn-type="dfn" id="solid-app">Solid app</dfn> is an application that reads or writes data from one or more <a data-link-type="dfn" href="#data-pod" id="ref-for-data-pod">data pods</a>.</p>
 
+                  <p>An <dfn class="dfn-paneled" data-dfn-type="dfn" id="owner">owner</dfn> is a person or a social entity identified by a URI that has implicit control of a data pod.</p>
+
                   <p>A <dfn data-dfn-type="dfn" id="read-operation">read operation</dfn> entails that information about a resource’s existence or its description can be known. [<a href="https://github.com/solid/specification/issues/149#issue-568433265" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
                   <p>A <dfn data-dfn-type="dfn" id="write-operation">write operation</dfn> entails that information about resources can be created or removed. [<a href="https://github.com/solid/specification/issues/126#issuecomment-569920473" rel="cito:citesAsSourceDocument">Source</a>]</p>
@@ -434,6 +436,8 @@ left:4.5em;
           <section id="uri" inlist="" rel="schema:hasPart" resource="#uri">
             <h2 property="schema:name">Uniform Resource Identifier</h2>
             <div datatype="rdf:HTML" property="schema:description">
+              <p class="note" role="note"><span>Note</span>: This specification does not describe the relationship between a Solid data pod <q>owner</q> and Web architecture’s <cite><a href="https://www.w3.org/TR/webarch/#uri-ownership">URI ownership</a></cite> [<cite><a class="bibref" href="#bib-webarch">WEBARCH</a></cite>].</p>
+
               <section id="uri-slash-semantics" inlist="" rel="schema:hasPart" resource="#uri-slash-semantics">
                 <h3 property="schema:name">URI Slash Semantics</h3>
                 <div datatype="rdf:HTML" property="schema:description">
@@ -475,6 +479,14 @@ left:4.5em;
                   <p>Clients can discover a storage by making an HTTP <code>GET</code> request on the target URL to retrieve an RDF representation [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>], whose encoded RDF graph contains a relation of type <code>http://www.w3.org/ns/pim/space#storage</code>. The object of the relation is the storage (<code>pim:Storage</code>).</p>
 
                   <p>[<a href="https://github.com/solid/data-interoperability-panel/issues/10#issuecomment-598694029" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/153#issuecomment-624630022" rel="cito:citesAsSourceDocument">Source</a>]</p>
+
+                  <p>Servers MUST keep track at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way. The owner is first set at storage provisioning time and can be changed. The owner is allowed implicit control access on every resource in a storage.</p>
+
+                  <p>When a server wants to advertise the owner of a storage, the server MUST include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the URI of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</p>
+
+                  <p class="note" role="note"><span>Note</span>: When a server supports multiple storages, there must be complete trust between its owners.</p>
+
+                  <p>[<a href="https://github.com/solid/specification/issues/67" rel="cito:citesAsSourceDocument">Source</a>][<a href=" https://github.com/solid/specification/issues/132" rel="cito:citesAsSourceDocument">Source</a>][<a href="https://github.com/solid/specification/issues/153" rel="cito:citesAsSourceDocument">Source</a>][<a href="https://github.com/solid/specification/issues/197" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
                   <p>When using Web Access Control (<a href="#web-access-control">Web Access Control</a>):</p>
 
--- a/protocol.html
+++ b/protocol.html
@@ -334,7 +334,7 @@ left:4.5em;
 
                   <p>A <dfn class="dfn-paneled" data-dfn-type="dfn" id="solid-app">Solid app</dfn> is an application that reads or writes data from one or more <a data-link-type="dfn" href="#data-pod" id="ref-for-data-pod">data pods</a>.</p>
 
-                  <p>An <dfn class="dfn-paneled" data-dfn-type="dfn" id="owner">owner</dfn> is a person or a social entity identified by a URI that has implicit control of a data pod.</p>
+                  <p>An <dfn class="dfn-paneled" data-dfn-type="dfn" id="owner">owner</dfn> is a person or a social entity that is considered to have the rights and responsibilities of a data storage. An owner is identified by a URI, and implicitly has control over all data in a storage. An owner is first set at storage provisioning time and can be changed.</p>
 
                   <p>A <dfn data-dfn-type="dfn" id="read-operation">read operation</dfn> entails that information about a resource’s existence or its description can be known. [<a href="https://github.com/solid/specification/issues/149#issue-568433265" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
@@ -480,7 +480,7 @@ left:4.5em;
 
                   <p>[<a href="https://github.com/solid/data-interoperability-panel/issues/10#issuecomment-598694029" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/153#issuecomment-624630022" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
-                  <p>Servers MUST keep track at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way. The owner is first set at storage provisioning time and can be changed. The owner is allowed implicit control access on every resource in a storage.</p>
+                  <p>Servers MUST keep track of at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way.</p>
 
                   <p>When a server wants to advertise the owner of a storage, the server MUST include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the URI of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</p>
 


timbl (approved):
looks good to me

bblfish:
The domain of the relation is not defined explicit. I guess it is a Pod, or would it be a container?

csarven:
Intentionally undefined for the time being. Implementation feedback would help. In the Protocol, servers exposing the storage resource (root container) also has a `rel="type"` link relation `http://www.w3.org/ns/pim/space#Storage`.

If we lean towards pim/space, it could also be `pim:ControlledStorage`. The property could alternatively go under pim/space eg. `pim:owner`.

kidehen (changes requested):
> Intentionally undefined for the time being. Implementation feedback would help. In the Protocol, servers exposing the storage resource (root container) also has a `rel="type"` link relation `http://www.w3.org/ns/pim/space#Storage`.
> 
> If we lean towards pim/space, it could also be `pim:ControlledStorage`. The property could alternatively go under pim/space eg. `pim:owner`.

You should include the rdf:domain for clarity. 
As I see it, the rdf:domain would be rdfs:Resource if want to be broad.

```
solid:owner
    a rdf:Property ;
    dc:issued "2021-05-19"^^xsd:date ;
    rdfs:comment "The person or other agent which owns the data storage. For example, the owner has implicit control of a data storage."@en ;
    rdfs:isDefinedBy <http://www.w3.org/ns/solid/terms#> ;
    ## Property Domain Addition
    rdfs:domain rdfs:Resource ;
    rdfs:range foaf:Agent ;
    rdfs:label "owner"@en .
```

protocol.html
```diff
@@ -476,6 +480,14 @@ <h3 property="schema:name">Storage</h3>
 
                   <p>[<a href="https://github.com/solid/data-interoperability-panel/issues/10#issuecomment-598694029" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/153#issuecomment-624630022" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
+                  <p>Servers MUST keep track at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way. The owner is first set at storage provisioning time and can be changed. The owner is allowed implicit control access on every resource in a storage.</p>
+
+                  <p>When a server wants to advertise the owner of a storage, the server MUST include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the URI of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</p>
+
+                  <p class="note" role="note"><span>Note</span>: When a server supports multiple storages, there must be complete trust between its owners.</p>
```
matthieubosquet, line 487:
```suggestion
                  <p class="note" role="note"><span>Note</span>: When a server supports multiple owners, there must be complete trust between its owners.</p>
```
?

protocol.html
```diff
@@ -476,6 +480,14 @@ <h3 property="schema:name">Storage</h3>
 
                   <p>[<a href="https://github.com/solid/data-interoperability-panel/issues/10#issuecomment-598694029" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/153#issuecomment-624630022" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
+                  <p>Servers MUST keep track at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way. The owner is first set at storage provisioning time and can be changed. The owner is allowed implicit control access on every resource in a storage.</p>
+
+                  <p>When a server wants to advertise the owner of a storage, the server MUST include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the URI of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</p>
```
matthieubosquet, line 485:
```suggestion
                  <p>When a server wants to advertise the owner of a storage, the server MUST include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the WebID of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</p>
```

protocol.html
```diff
@@ -476,6 +480,14 @@ <h3 property="schema:name">Storage</h3>
 
                   <p>[<a href="https://github.com/solid/data-interoperability-panel/issues/10#issuecomment-598694029" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/153#issuecomment-624630022" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
+                  <p>Servers MUST keep track at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way. The owner is first set at storage provisioning time and can be changed. The owner is allowed implicit control access on every resource in a storage.</p>
```
matthieubosquet, line 483:
```suggestion
                  <p>Servers MUST keep track at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way. An owner is first set at storage provisioning time and can be changed. An owner is allowed implicit control access on every resource in a storage.</p>
```

protocol.html
csarven: I think when a storage has multiple owners, trust between them is generally expected or well-understood. This is really about when a server supports multiple storages with different owners under the same host.

protocol.html
csarven: Almost interchangeable. I did consider WebID but decided to go with a URI-reference only. As it stands, there is no additional requirement (and to test) whether the reference must resolve to a document including the WebID. I'm not against using WebID, just think URI would be more appropriate here.

csarven:
@kidehen No objection to including `rdfs:Resource` but I don't think it is needed. Wouldn't that always be inferred?

kidehen:
> @kidehen No objection to including `rdfs:Resource` but I don't think it is needed. Wouldn't that always be inferred?

rdfs:Resource is a broad solution to the issue at hand. There are narrower solutions that can also be applied e.g., whatever classes are associated with the notion of a Solid Pod etc. Fundamentally though, you need an rdfs:domain as an integral part of an rdf:Property definition. That isn't something that should be inferred.

protocol.html
```diff
@@ -334,6 +334,8 @@ <h3 property="schema:name">Definitions</h3>
 
                   <p>A <dfn class="dfn-paneled" data-dfn-type="dfn" id="solid-app">Solid app</dfn> is an application that reads or writes data from one or more <a data-link-type="dfn" href="#data-pod" id="ref-for-data-pod">data pods</a>.</p>
 
+                  <p>An <dfn class="dfn-paneled" data-dfn-type="dfn" id="owner">owner</dfn> is a person or a social entity identified by a URI that has implicit control of a data pod.</p>
```
RubenVerborgh, line 337: I think we're conflating the definition of an owner with the abilities arising from it (control).

Not perfect, but:

```suggestion
                  <p>An <dfn class="dfn-paneled" data-dfn-type="dfn" id="owner">owner</dfn> of a data pod is a person or other agent identified by a URI that is considered responsible for its storage. Such responsibility arises, for instance, when someone registers a pod, control its domain name, etc. Owners implicitly have control over all data in the pod.</p>
```

protocol.html
RubenVerborgh, line 483:
```suggestion
                  <p>Servers MUST keep track of at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way. The owner is first set at storage provisioning time and can be changed. The owner is allowed implicit control access on every resource in a storage.</p>
```

RubenVerborgh:
> ```turtle
>     rdfs:comment "The person or other agent which owns the data storage. For example, the owner has implicit control of a data storage."@en ;
> ```

The "for example" doesn't seem to work well.

protocol.html
matthieubosquet: I'm not sure I fully understand the relationship/geometry between server, owner and storage (and Pod).

bblfish:
So if I understand correctly in N3 we have 

```Turtle
# we first define the transitive closure of ldp:contains
:contains٭ owl:TransitiveProperty;
     rdfs:subPropertyOf ldp:contains .

# we then have the N3 rule
{ ?rootLdpc :owner ?agent } => {  
     [] acl:agent ?agent ;
        acl:mode acl:Control;
        acl:accessToClass [ a owl:Restriction;
              owl:onProperty [ owl:inverseOf :contains٭ ];
              owl:hasValue ?rootLdpc 
         ]
}
```

If it were a little clearer what the domain of the relation was (the domain-name perhaps?) one could adapt the definition.

protocol.html
csarven: So far the notion of "owner" was deemed to be in the scope of a data pod / storage and was not in the layer of controlling the domain name, URI ownership etc. The suggestion above conflicts with the Note. Of course nothing stops us from defining the owner to beyond the storage. I think Solid Protocol's terms should stay within or thereabouts the features it enables.

I like the take on "considered" and "responsible", so did some shuffling around and came up with this: https://github.com/solid/specification/pull/264/commits/6ba032d7556f42d2323a339adfeb03965d973024 . WDYT?

I also tried to keep it somewhat compatible with https://dpvcg.github.io/dpv/#hasDataController . "Owner" should decide on how data can be ultimately processed in a storage, and also be held accountable.

csarven:
@bblfish As mentioned, if a specific domain is necessary, the most suitable or closest is perhaps `http://www.w3.org/ns/pim/space#Storage` (set on root container). I was hoping to hear more about other possibilities within the context of the Protocol. I don't think we should equate, infer or make a specific connection to domain name owner/renter.

protocol.html
RubenVerborgh: Looks good, just added some small nits.

csarven:
* I think that a "social entity" is a living thing, an organisation or a group of sorts that can be held accountable (based on whatever definition, jurisdiction.. it is taken up). A software agent probably wouldn't be based on that viewpoint but I can see that it could be delegated to carry out the duties to some extent. (IANAL... and this is not a hill I want to die on.)
* Owner is the subject of the sentence.
* Yes, implementation dependent. The (initial) assignment and tracking of the owner is implementation dependent to begin with.

kidehen:
> > @kidehen No objection to including `rdfs:Resource` but I don't think it is needed. Wouldn't that always be inferred?
> 
> rdfs:Resource is a broad solution to the issue at hand. There are narrower solutions that can also be applied e.g., whatever classes are associated with the notion of a Solid Pod etc. Fundamentally though, you need an rdfs:domain as an integral part of an rdf:Property definition. That isn't something that should be inferred.

Please scratch this response. Basically, I didn't digest the rdfs:comment aspect of this matter. 
I am going to post an example in a separate response.

kidehen:
> @kidehen No objection to including `rdfs:Resource` but I don't think it is needed. Wouldn't that always be inferred?

Yes, based on my initial errant response. 
Here's a corrected response using RDF-Turtle. 

Here's what I didn't pick up first time around, by not digesting rdfs:comment . 

```
solid:owner
    a rdf:Property ;
    rdfs:comment "The person or other agent which owns the data storage. For example, the owner has implicit control of a data storage."@en;
   rdfs:range foaf:Agent ;
   rdfs:label "owner"@en .
```

That's very confusing, from my vantage point. 


Here's my suggested revision. 

```
 ## Turtle Start ##

@prefix solid: <http://www.w3.org/ns/solid/terms#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> . 
@prefix dc: <http://purl.org/dc/elements/1.1/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .

solid:owner
    a rdf:Property ;
    dc:issued "2021-05-19"^^xsd:date ;
    rdfs:comment """A property that associates a person (or other agent) with data storage over which they have paramount 
                   control. For example, the owner who has controlling privileges over data storage associated with a Solid Pod."""@en ;
    rdfs:isDefinedBy <http://www.w3.org/ns/solid/terms#> ;
    ## Property Domain Addition
    rdfs:domain foaf:Agent ;
    rdfs:range rdfs:Resource ;
    rdfs:label "owner"@en .


# For additional clarity
# using an additional Property definition

solid:hasOwner
    a rdf:Property ;
    owl:inverseOf solid:owner ;
    dc:issued "2021-05-19"^^xsd:date ;
    rdfs:comment """A property that associates data storage associated a Solid Pod with a person (or other agent) that has 
                    paramount control. For example, the owner who has controlling privileges over data storage associated with 
                    a Solid Pod."""@en ;
    rdfs:isDefinedBy <http://www.w3.org/ns/solid/terms#> ;
    ## Property Domain Addition
    rdfs:domain rdfs:Resource ;
    rdfs:range foaf:Agent ;
    rdfs:label "hasOwner"@en .

## Turtle End ##
```

I hope this helps.

csarven:
@kidehen , `solid:owner` doesn't seem right for the domain/range you're using. Perhaps you meant `solid:owns` or `solid:ownerOf` - "this agent owns or is an owner of that". But this is not the focus of this PR.

As for `solid:hasOwner`, no need for "has". `solid:owner` is sufficient.

kjetilk (commented):
I have :+1: a couple of comments I think should go in.

In addition, I'm :+1: on @csarven 's suggestion that `pim:Storage` should be the domain, and I think it should be added, since that ties in the definition of what storage is.

kidehen:
> @kidehen , `solid:owner` doesn't seem right for the domain/range you're using. Perhaps you meant `solid:owns` or `solid:ownerOf` - "this agent owns or is an owner of that". But this is not the focus of this PR.
> 
> As for `solid:hasOwner`, no need for "has". `solid:owner` is sufficient.

This is getting too confusing. 

I am responding to solid:owner to [this post](https://github.com/solid/specification/pull/264#issue-647658451#this) that includes:
```
solid:owner
    a rdf:Property ;
    dc:issued "2021-05-19"^^xsd:date ;
    rdfs:comment "The person or other agent which owns the data storage. For example, the owner has implicit control of a data storage."@en ;
    rdfs:isDefinedBy <http://www.w3.org/ns/solid/terms#> ;
    rdfs:range foaf:Agent ;
    rdfs:label "owner"@en .
```

My fundamental point is that solid:owner needs an rdfs:domain as part of its definition, whatever that happens to be :)

kidehen:
> @bblfish As mentioned, if a specific domain is necessary, the most suitable or closest is perhaps `http://www.w3.org/ns/pim/space#Storage` (set on root container). I was hoping to hear more about other possibilities within the context of the Protocol. I don't think we should equate, infer or make a specific connection to domain name owner/renter.

That's fine re rdfs:domain.

elf-pavlik:
> I think that a "social entity" is a living thing, an organisation or a group of sorts that can be held accountable (based on whatever definition, jurisdiction.. it is taken up). A software agent probably wouldn't be based on that viewpoint but I can see that it could be delegated to carry out the duties to some extent. (IANAL... and this is not a hill I want to die on.)

👍 

We have related issue in https://github.com/solid/data-interoperability-panel/issues/80 *(Agree on terminology to distinguish software agent from social agent)*. I mention there [Legal person](https://en.wikipedia.org/wiki/Legal_person) as one of possible concepts we could reference. Only *social agent* could own data, while *software agent* always acts on behalf of some *social agent* who controls it (even by fully automating it).

kjetilk:
> Your proposal to define it in terms of `pim:Storage` is better, but still seems somewhat vague then. It leaves open whether we are identifying electronic devices via a hardware serial number, Hardware UUID, Provisioning UDID (info I find available for my computer) or with a domain. Are we interested in the ownership of the hardware, or in the topological position of the device on the internet? For the latter, it could be a [MAC address](https://en.wikipedia.org/wiki/MAC_address) or a domain name.

Alright, perhaps we should have a `solid:Storage` and tighten the definition appropriately?

I agree that we should have the domain in there as it very clearly binds the definition of an owner to the definition of a storage, which is very clearly specified in terms of the URI  hierarchy in the protocol spec, which is again a rigorous definition AFAICS. It is therefore important to have a RDFS class that is equally strongly tied to the protocol's definition of storage. I can see that the `pim:Storage` definition can be interpreted rather broadly, so I say we tighten it.

protocol.html
```diff
@@ -476,6 +480,14 @@ <h3 property="schema:name">Storage</h3>
 
                   <p>[<a href="https://github.com/solid/data-interoperability-panel/issues/10#issuecomment-598694029" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/153#issuecomment-624630022" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
+                  <p>Servers MUST keep track of at least one <a data-link-type="dfn" href="#owner" id="ref-for-owner">owner</a> of a storage in an implementation defined way.</p>
+
+                  <p>When a server wants to advertise the owner of a storage, the server MUST include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the URI of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</p>
```
acoburn, line 485: > When a server wants to advertise the owner ...

Does this allow for the possibility that an owner may wish to remain anonymous? Another way to phrase this question: is it up to the server to decide whether to advertise the owner? Or is that an implementation detail?

csarven: [Good question. Can of worms on data minimisation...]

There is no constraint on reference to owner entailing a real, anonymous or pseudonymous identity. There is also no requirement (so far) that the URI of the owner needs to have a representation. While information could be inferred from the URI, and as you know, ultimately it is opaque. So, the URI doesn't tell us much about the social entity behind it. I also don't think the owner link relation should be coupled with access rights.

No need make the distinction in the spec about the owner's identity, anonymity, pseudonymity, as there are good cases for all; implementation detail. One side-effect of that is that identities could also be linkable by the system eg. the anonymous owner with the real identity.

The main question may just boil down to whether the owner must always be discoverable or not. If the spec requires owners to be discoverable, an implementation could for example choose to use an anonymous identifier, which may effectively be same as not advertising it to begin with, especially if the owner resource has no representation or no access provided to the world.

I'd be comfortable with the idea/requirement that  server always advertises the owner link relation. WDYT?

kjetilk: Great question indeed! It seems that in the interest of data minimisation, it shouldn't be a requirement to make it discoverable, and I don't see the usefulness of an anonymous identifier. A pseudonym could be useful though.

Generally, we should assume that a WebID is personally identifiable information (because it really is, even though good pseudonymisation should help with that). 

I imagine the exposure of an owner should be configurable, and so, yes, an implementation detail. I think we should note that in the privacy considerations in the spec though, so people aren't just making assumptions about it.

csarven: The usefulness of any identifier is that, if it resolves to a representation, the owner can be contacted eg. via its inbox, without reveal anything else.

kjetilk: right, but any attribute might also reveal identity... Can we even talk about an anonymous identity, aren't we then really talking about a pseudonym? 

I think a sophisticated treatment of pseudonyms is a very interesting topic, but not something we should do before this is merged.

protocol.html
elf-pavlik: How does it apply to https://solidcommunity.net/ ? One server supports multiple users but uses subdomain for each storage.

protocol.html
kjetilk: Basically, you have to ensure that pseudonymity or anonymity is maintained with all subsequent operations, i.e. you dereference the owner's URI, you get back some attributes amongst those, an inbox. 

With anonymity, you have to ensure that none of these things can tell you anything at all. With pseudonymity, you just have to ensure that these things cannot tell someone anything more than they already know about the identity. It is probably easier. I think... :-)

Anyway, there seems to be two open issues here, the rest can go in. How about this: We merge this PR as it currently is, and then we open two issues: 

- [ ] What is the domain of the `solid:owner` relation?
- [ ] whether the owner must be discoverable or not.

acoburn: Thanks, @csarven and @kjetilk. In the context of my question, I think it's possible to set aside the question of identifier resolution (n.b. this may be an interesting use case for DIDs, so for the domain/range conversation, that would be something to consider).

> I'm leaning on requiring [the owner identifier relation] without expecting anything further about the identity that it refers to.

:+1: I would support that.

kjetilk: I'm :-1: because there is an ocean of difference between anonymity and pseudonymity when you come to legal matters, for example GDPR applies to pseudonyms but not when the user is anonymous. Lets merge this and open new issues.

protocol.html
justinwb: > This is really about when a server supports multiple storages with different owners under the same host.

A Pod hosting scenario (per @elf-pavlik's point) would seem to fall under this category where a given server has multiple storages, with each having different owners, who may not know/trust each other.

csarven: ?

The key word that I used was "host" but perhaps I should've used "scheme, host, and port" to be more accurate. This is boils down to origin (see RFC 6454). solidcommunity.net designates a unique "scheme, host, and port" with one "storage" to one "owner". It is currently a non-issue for solidcommunity.net as I understand it.

The main point remains in that if a service (like solidcommunity.net) were to allow multiple storages under the same "scheme, host, and port", then the owners of the storages (which could be the same owner) need to establish "trust" among themselves.

protocol.html
csarven: The Protocol is not prescribing the nature of the owner URI, and leaves it as an implementation detail that could be configurable to be compliant with what's legally applicable at pod provisioning time (by its owner for example).

No, the discovery of the owner link relation - server-controlled HTTP header - is not coupled with access privileges.

protocol.html
```diff
@@ -334,6 +334,8 @@ <h3 property="schema:name">Definitions</h3>
 
                   <p>A <dfn class="dfn-paneled" data-dfn-type="dfn" id="solid-app">Solid app</dfn> is an application that reads or writes data from one or more <a data-link-type="dfn" href="#data-pod" id="ref-for-data-pod">data pods</a>.</p>
 
+                  <p>An <dfn class="dfn-paneled" data-dfn-type="dfn" id="owner">owner</dfn> is a person or a social entity that is considered to have the rights and responsibilities of a data storage. An owner is identified by a URI, and implicitly has control over all data in a storage. An owner is first set at storage provisioning time and can be changed.</p>
```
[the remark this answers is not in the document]
csarven: Owner is the subject of the sentence.

csarven:
@kidehen @bblfish @kjetilk  :

The intended definition of `solid:owner` draws from `acl:owner`. It is broader in the sense that "there is a sense of right of control" over all data in a storage. Obviously no disagreement to make `solid:owner` more clear. Part of the reason why I left it vague was to work it out based on what's intended by the spec. The current `rdfs:range foaf:Agent` is intentional, so the term is definitely not about "agent owns/ownerOf thing".

We use the `pim:storage` property eg. agent has storage, where the range is defined to be an instance of `pim:Storage`. The arc from a storage to an agent is intended to be `solid:owner`. I think that's fine (ie. we can do `rdfs:domain pim:Storage`. Unless `pim:Storage` gets new subclasses (eg. for new devices/hardware..), I don't see what an equivalent like `solid:Storage` brings to the table.

@elf-pavlik 

I borrowed/introduced "social entity" from AWWW's URI ownership. There are social expectations and processes. Possible to go from responsibility/accountability chain from data "ownership" to URI ownership.. domain name registration through IANA (another social entity). And as I've said at the beginning of this PR:

>If describing the relationship between Solid data pod / storage "owner" and AWWW's "URI ownership" is required, it can be done in another PR.

AFAIK, "owner" in Solid is deemed to be a "social entity" of sorts, as opposed to a software agent. If there is existing discussion/work that incorporates non-living organisms under "owner", references would be great. (And to our future non-living organism overlords, I don't mean to discriminate, please go easy on us.)

protocol.html
kjetilk: > The Protocol is not prescribing the nature of the owner URI, and leaves it as an implementation detail that could be configurable to be compliant with what's legally applicable at pod provisioning time (by its owner for example).

Still, lets open a separate issue on it. The use cases are really unclear to me, and the requirements around what is an anonymous identifier (is there such a thing?) and what you'd get out of dereferencing it is also unclear.

kjetilk:
> I don't see what an equivalent like `solid:Storage` brings to the table.

Again, lets open a separate issue. I'd like to discuss tying it very explicitly to Solid's definition, which is rooted in its hierarchy, that's something that doesn't exist in the general case.

csarven:
>I'd like to discuss tying it very explicitly to Solid's definition, which is rooted in its hierarchy, that's something that doesn't exist in the general case.

Perhaps `solid:Resource`: https://github.com/solid/specification/issues/194#issuecomment-694828342

protocol.html
justinwb: Pod hosting isn't necessarily exclusive to sub-domains for segregation. Pod hosting scenarios can, and do, use path based segregation (e.g. https://host.example/pod-one, https://host.example/pod-two). In cases like those, the scheme, host, and port are the same, but still you have the likely scenario of different hosted pods owned by social agents that do know know and/or trust each other.

kjetilk (approved):
I have opened #265 and #266 to continue the discussion of the exposure of the owner relation and the domain of the relation there. 

And given that these issues can be resolved there, I approve this PR.

timbl (approved):
looks good to me this is important for existing servers.

csarven:
Implemented owner discovery in https://github.com/csarven/mayktso 

```
curl -I https://linkedresearch.org/ldn/tests/

link: <http://www.w3.org/ns/pim/space#Storage>; rel="type"
link: <https://csarven.ca/#i>; rel="http://www.w3.org/ns/solid/terms#owner", <https://example.org/a43a1874-c51e-11eb-a893-3fad0d2fdf12#i>; rel="http://www.w3.org/ns/solid/terms#owner"
```

acoburn (approved):

justinwb (approved):
- I support the notion of a storage owner, because it's valuable for interoperability and data sharing workflows (e.g. where data in a given storage is shared by a non-owner with access to do so, and whose data they are sharing needs to be communicated)
- I support public advertisement because there are cases where a non-privileged agent should be able to know whose storage they're accessing.

It is important that public advertisement can be anonymous, and I believe that most pod owners will choose to do so. This may impact the viability of certain workflows where discovery of ownership is necessary. That needs to be addressed, but doesn't have to block this submission, as we now have a separate thread where this is being discussed (#266). Similarly, I've also created #267 to continue [this thread](https://github.com/solid/specification/pull/264#discussion_r635381752).

protocol.html
csarven: I responded to the original inquiry about subdomain based services and explained why that was a non-issue. Yes, multiple storages (on the same origin) with different owners has been possible for long time. It was never prohibited to begin with! (Heck, we've had different types of storages as well as workspaces... ) One of the reasons for the whole set of requirements around storage was acknowledging that reality (see PR https://github.com/solid/specification/pull/208 and https://solidproject.org/TR/protocol#storage ).. making sure non-overlapping URI ((path) space are used, among other things.

Trivial to get a wholesale scheme-host-port blocked (rendering it useless) by a service because of one bad actor, and hard to isolate the bad actors in that system. Not to forget CORS considerations.

The *note* doesn't prohibit servers/services that possibility. Both the service provider and the owners of the storages should be aware of what it entails.

The point in the note still stands and relevant for any set up.

protocol.html
csarven: >The use cases are really unclear to me

System enforcing owner. System enabling clients to discover owner.

>the requirements around what is an anonymous identifier (is there such a thing?) and what you'd get out of dereferencing it is also unclear.

That's because this PR did not set out to prescribe what the owner identifier should look like (or imply anything beyond it being opaque) or whether the system requires to generate a representation and what the description needs to look like. Orthogonal.

Sure, we can continue/extend the discussion in the issues as you've created but I didn't see these as blockers for this PR (FWIW).

kjetilk: Yup, and exactly because it is orthogonal, it should have a separate issue. :-)

bourgeoa (approved):

bourgeoa (approved):

csarven:
Was waiting for additional feedback from implementers before merging..

Mayktso currently only implements the discovery bit as mentioned earlier: https://github.com/solid/specification/pull/264#issuecomment-854629484

Noting here that CSS is in the process of implementing this feature: https://github.com/solid/community-server/issues/720 .

NSS has a PR for it: https://github.com/solid/node-solid-server/pull/1604

Solid-NextCloud has shown interest in implementing: https://github.com/pdsinterop/solid-nextcloud/issues/48

Will update this comment with more.. or others (/me friendly nudges @kidehen @gibsonf1 .. ) can show/link to their implementations in the comments

---

As for the `solid:owner` definition, here is an update based on earlier discussion that can be PR'd to solid/vocab - we can finalise the exact definition there, instead of here:

```turtle
solid:owner
    a rdf:Property ;
    dc:issued "2021-05-19"^^xsd:date ;
    rdfs:comment "The a person or a social entity that is considered to have control, rights and responsibilities of a data storage."@en ;
    rdfs:isDefinedBy <http://www.w3.org/ns/solid/terms#> ; # alternative: <https://solidproject.org/TR/protocol> ;
    rdfs:domain ws:Storage ;
    rdfs:range foaf:Agent ;
    rdfs:label "owner"@en .
```

---

As agreed, we'll work out related/orthogonal issues (that were deemed to be non-blocking).

PR #290, opened 2021-07-27
Trust between owners on same origin with multiple storages

### LINKED ISSUE #267: Should there be trust between owners when a server supports multiple storages on the same origin
Creating this issue to unblock #264, and to continue [this thread](https://github.com/solid/specification/pull/264#discussion_r635381752).

For context, a note is provided in the addition of ownership requirements in #264 that states:

> When a server supports multiple storages, there must be complete trust between its owners

Assuming that: 

- _server_ in this context is equivalent to a running instance of a solid server (e.g. community-server at `https://pod.example`)
- support of multiple storages with different owners equates to a typical pod hosting scenario (e.g. `https://solidcommunity.net`) 
 
In this case, it would be true more often than not that a server with multiple storages does not realize complete trust between the owners of those storages. For example, I have a solidcommunity.net pod, but I do not know the majority of that user population, let alone completely trust them.

If _server_ in this context was meant to uniquely identify scheme, host, and port (e.g. `https://justinwb.pod.example/`, that may address cases where _server_  represents a storage partitioned by subdomain, but not cases where a server partitions storages by path (e.g. `https://pod.example/justin/`, `https://pod.example/alice/`).

csarven:
Closes #267

### DIFF
--- a/ED/protocol.html
+++ b/ED/protocol.html
@@ -745,13 +745,6 @@ margin-top:1em;
 
                   <p><span about="" id="server-storage-link-owner" rel="spec:requirement" resource="#server-storage-link-owner"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="#Server">server</span> wants to advertise the owner of a storage, the server <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the URI of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</span></span></p>
 
-                  <div class="note" id="trust-between-owners" inlist="" rel="schema:hasPart" resource="#trust-between-owners">
-                    <h4 property="schema:name"><span>Note</span>: Trust Between Owners</h4>
-                    <div datatype="rdf:HTML" property="schema:description">
-                      <p>When a server supports multiple storages, there must be complete trust between its owners.</p>
-                    </div>
-                  </div>
-
                   <p>[<a href="https://github.com/solid/specification/issues/67" rel="cito:citesAsSourceDocument">Source</a>][<a href=" https://github.com/solid/specification/issues/132" rel="cito:citesAsSourceDocument">Source</a>][<a href="https://github.com/solid/specification/issues/153" rel="cito:citesAsSourceDocument">Source</a>][<a href="https://github.com/solid/specification/issues/197" rel="cito:citesAsSourceDocument">Source</a>]</p>
                 </div>
               </section>


kjetilk (approved):
This, I think is a sufficient explanation. In a subsequent best practice document, we should also say that this means that unless you know what you're doing (like making a server for your family), you should ensure that different users have different origins.

kjetilk (approved):
Great, I think this strengthens it even further.

protocol.html
```diff
@@ -901,6 +901,8 @@ <h3 property="schema:name">Security Considerations</h3>
...
 
+                  <p>Servers are discouraged from implementing <a href="#storage">multiple storages on the same-origin with untrusted owners</a> because Web apps running in one storage will have the same rights as Web apps running in the other.</p>
```
justinwb, line 904: > Because Web apps running in one storage will have the same rights as Web apps running in the other

I think we adding a bit more detail here on the context (e.g. browser-based apps relying on same origin policy) and the attack vectors would be helpful. Understanding the risk better should make it easier for the reader to understand why they are being discouraged from hosting multiple storages in same origin.

acoburn: To continue @justinwb's comment, it is important to be clear about the assumptions that underlie this statement. If, for instance, CORS is used as the basis of identifying applications, which in turn is used for access control, then yes, this would be an important point to make.

If, however, applications are identified through [another mechanism](https://solid.github.io/solid-oidc/#clientids), I struggle to understand why this statement would be needed.

I would also mention that CORS-based restrictions are _only_ going to be relevant for certain browser-based use cases. It is entirely irrelevant for non-browser contexts and even for browser contexts, there are fairly easy ways to circumvent same-origin restrictions (i.e. adding a proxy service)

elf-pavlik: 👍 both previous comments

> because Web apps running in one storage

I also find word *running* in this statement possibly leading to some confusion. While application source can be *published* to solid storage, that application *runs* in a web browser on someone's device.

Applications running on remote server (aka 'server side') are perfectly fine, in some ways can be even better, solid applications. Those run on the server at certain origin but currently can't be deployed to solid storage.

protocol.html
```diff
@@ -484,7 +484,7 @@ <h3 property="schema:name">Storage</h3>
 
                   <p>When a server wants to advertise the owner of a storage, the server MUST include the <code>Link</code> header with <code>rel="http://www.w3.org/ns/solid/terms#owner"</code> targeting the URI of the owner in the response of HTTP <code>HEAD</code> or <code>GET</code> requests targeting the root container.</p>
 
-                  <p class="note" role="note"><span>Note</span>: When a server supports multiple storages, there must be complete trust between its owners.</p>
+                  <p class="note" role="note"><span>Note</span>: When a server supports multiple storages on the same-origin [<cite><a class="bibref" href="#bib-rfc6454">RFC6454</a></cite>], there must be complete trust between its owners</p>
```
dmitrizagidulin, line 487:
```suggestion
                  <p class="note" role="note"><span>Note</span>: When a server supports multiple storages on the same-origin [<cite><a class="bibref" href="#bib-rfc6454">RFC6454</a></cite>], there must be complete trust between its owners.</p>
```

csarven:
Changes follow action group's review of this PR https://github.com/solid/specification/blob/main/meetings/2022-11-30.md#trust-between-owners-on-same-origin-with-multiple-storages :

The PR introduced advisory content under Security Consideration and the concerns in this PR was specifically about that. The group (2022-11-30 meeting) agreed to leave out this advisory. There was also no objection to removing the Note and some approval to remove it as well. This PR however only made an editorial change to the Note.

On second review, I took it on myself to remove the Note as well - given consent - based on the grounds that the advisory is too vague to be implementable by a server and/or actionable by owners involved.

PR #292, opened 2021-08-02
Put the If-None-Match: * requirement on the server

kjetilk:
Tests have shown inconsistent treatment of the `If-None-Match: *` header. I believe there is consensus that Solid Servers must support this feature based on the issues discussions in #40 and #108. I also think we cannot put such a requirement on the client, also we can't test that. Thus, I have changed the wording to make it a requirement on the servers.


protocol.html
```diff
@@ -637,7 +637,7 @@ <h3 property="schema:name">Writing Resources</h3>
...
+                  <p>Servers MUST ensure that HTTP <code>If-None-Match</code> header with a value of <code>"*"</code> will prevent an unsafe request method (e.g., <code>PUT</code>, <code>PATCH</code>) from modifying an existing representation of the target resource. A client MAY use this header if it believes that the resource does not have a current representation. [<a href="https://github.com/solid/specification/issues/108#issuecomment-567272797" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/40#issuecomment-566995240" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
csarven, line 640:
```suggestion
                  <p class="note" role="note"><span>Note:</span>Clients are encouraged to use the HTTP <code>If-None-Match</code> header with a value of <code>"*"</code> to prevent an unsafe request method (e.g., <code>PUT</code>, <code>PATCH</code>) from inadvertently modifying an existing representation of the target resource when the client believes that the resource does not have a current representation. [<a href="https://github.com/solid/specification/issues/108#issuecomment-567272797" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/40#issuecomment-566995240" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

csarven (changes requested):
>Tests have shown inconsistent treatment of the If-None-Match: * header.

Vague. [Citation needed]

>I believe there is consensus that Solid Servers must support this feature based on the issues discussions in #40 and #108.

Already in the Protocol:

>A data pod MUST implement the server part of HTTP/1.1 Conditional Requests [RFC7232] to ensure that updates requested by clients will only be applied if given preconditions are met.

Moreover, for clients the Protocol includes:

>A Solid client MAY implement the client parts of HTTP/1.1 Conditional Requests [RFC7232] to only trigger updates when certain preconditions are met.

Hence, neither the original or the proposed text is necessary.

---

As for:

>I also think we cannot put such a requirement on the client

The requirement is not pushed to the client. Client behaviour is compatible with servers irrespective of the header in the request.

>, also we can't test that.

Why? (Putting aside the fact that it is not the client that actually prevents undesired modification.)

>Thus, I have changed the wording to make it a requirement on the servers.

Redundant.

---

I don't agree with the proposed change.

As mentioned, since server support is a MUST and client support is a MAY (as per #http section), no text is necessary. The purpose of the existing text in the Protocol was to (strongly?) encourage clients to use the header. So, I propose to change it to a Note - it is not a whole lot different than a MAY here but at least it is not redundant and still gets to put some emphasis on its use.

kjetilk:
Right. I hadn't noticed that there was a general must on the conditional requests. Then, I agree that a note is appropriate.

PR #293, opened 2021-08-03
Adjust URL for Solid-OIDC specification

acoburn:
The Solid-OIDC specification has been split off from the Authentication Panel repository. The new URL is https://solid.github.io/solid-oidc/ and the corresponding repository is at https://github.com/solid/solid-oidc


csarven (approved):
Thanks. Acknowledging it as a follow up on https://github.com/solid/solid-oidc/issues/13

PR #298, opened 2021-08-17
Clarify 409 on containment triples

RubenVerborgh:
I thought there were 2 minor problems with the original phrasing:

1. The containment triples _can_ be updated via `POST`/`PUT`/`PATCH` to _other_ resources, just not to the container itself.
2. `POST` is only used to create new resources, which does update containment triples; so I think it should be removed.


justinwb (approved):
+1 - FWIW - I've not found it onerous in our client implementations to filter out server-managed triples when updating the (non-server managed) graph in a container resource, so I think a 409 is reasonable (as opposed to ignoring proposed server-managed triples in the request).

kjetilk (approved):
Yes, I think this is an improvement, but lets wait for @csarven to merge.

csarven:
The initial (current) requirement simply repurposed LDP's text eg. instead of saying LDPC, it said container.

>The containment triples can be updated via POST/PUT/PATCH to other resources, just not to the container itself.

The request target is the container - that's the context. So eg. PUT /foo is not relevant or covered by or intended by this requirement.

>POST is only used to create new resources, which does update containment triples; so I think it should be removed.

Right but again the current requirement was about client directly changing container's description - as in actually including containment statements.

---

At the time of writing, there was consideration or at least to leave some space / variability on how POST can be used to update a container description (via RDF Merge) eg. POST / + Slug . (or some other way) which could include containment statements, hence being covered by this requirement. So, the current text was intentional, and to later revisit:

>Servers MUST allow creating new resources with a POST request to URI path ending /. Servers MUST create a resource with URI path ending /{id} in container /.

I'm okay to approve the PR *if* there is consensus to disallow POST to update container's description - because if we allow it to happen somehow, then we need a requirement along the lines of what's currently in the spec.

acoburn:
+1 on this clarification

One use case I'd like to leave room for is a "bulk update" or "import". For instance, if a user is able to download the contents of an entire Pod, it would be convenient to be able to import it somewhere else in a single operation. `POST` would be the obvious method for this operation. From my reading, that would still be possible with this clarification.

kjetilk:
Ah, thanks for the reminder about the possibility of a `Slug: .`, @csarven ! Actually, I found a bit of time to write down my RDF Merge proposal this weekend, I'll post that shortly, and there, I did not include updating the container. I think we can tighten it now and see if there is an opportunity to include that later if we find a suitable mechanism for doing so.

RubenVerborgh:
@kjetilk Done.

protocol.html
```diff
@@ -700,13 +700,13 @@ <h3 property="schema:name">Writing Resources</h3>
...
...<code>/</code>.</span></span> <span about="" id="server-post-container-create-resource" rel="spec:requirement" resource="#server-post-container-create-resource"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> create a resource with URI path ending <code>/{id}</code> in container <code>/</code>.</span></span> <span about="" id="server-post-container-create-container" rel="spec:requirement" resource="#server-post-container-create-continer"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> create a container with URI path ending <code>/{id}/</code> in container <code>/</code> for requests including the HTTP <code>Link</code> header with <code>rel="type"</code> targeting a valid LDP container type.</span></span> [<a href="https://github.com/solid/specification/pull/160#issuecomment-636822687" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/pull/190" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
dmitrizagidulin, line 703: > create a resource with URI path ending <code>/{id}</code>

Missing a 'with', should be: "create a resource with a URI path ending with"

dmitrizagidulin (approved):
:+1: 

The `with URI path ending <code>/{id}</code> in container <code>/</code>.` phrasing is a bit hard to parse, but I'm ok with the overall change.

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+* name: text
+
+---
+
+## Topics
+
+### Level of SPARQL Update support
+URL: https://github.com/solid/specification/issues/125
+
+* KK: Not entirely clear how PATCH updates should be done especially in a way that Tim wants. That is in conflict with a requirement in SPARQL update.. which makes it a bit messy. Haven't had the chance to talk to Tim so not sure what he thinks. So ended up making four different PRs. The 4th is a bit inconsistent and deviates a lot from SPARQL. There may be chances to get it into SPARQL if we go for it. There is also the code that Ruben used for N3Patch which is an alternative way. I don't want to mangle sPARQL to omuch to the point that the SPARQL group woldn't go for it. This is the issue we need Tim to discuss with. So, lets look at different issues.
```
TallTed, line 46:
```suggestion
* KK: Not entirely clear how PATCH updates should be done, especially in a way that Tim wants. That is in conflict with a requirement in SPARQL update... which makes it a bit messy. Haven't had the chance to talk to Tim so not sure what he thinks. So ended up making four different PRs. The 4th is a bit inconsistent and deviates a lot from SPARQL. There may be chances to get it into SPARQL if we go for it. There is also the code that Ruben used for N3Patch which is an alternative way. I don't want to mangle SPARQL too much, to the point that the SPARQL group wouldn't go for it. This is the issue we need to discuss with Tim. So, lets look at different issues.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* JB: It seems N3Patch proposal not contested too much. It seems from Tim's perspective it is not either. It doesn't seem like it is violating or deviating from existing standards. seems like the path of least resistence. I'm not sure we can figure out a compliant way for SPARQL that would satisfy everyone.
+
+* KK: I haven't contested it yet but have reservations. That is basically.. I'm still not sure what Tim is trying to achieve. There may be a misunderstanding of SPARQL. would like to clear that possibility out first.
```
TallTed, line 50:
```suggestion
* KK: I haven't contested it yet but have reservations. That is basically, I'm still not sure what Tim is trying to achieve. There may be a misunderstanding of SPARQL. would like to clear that possibility out first.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* RV: That may be a misunderstanding.. but we can still get a proper subset of SPARQL or any deviation.
```
TallTed, line 52:
```suggestion
* RV: That may be a misunderstanding... but we can still get a proper subset of SPARQL or any deviation.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: If you know what triple it is that you want to change. That is easy.. basically. It is also legitimate to change the triple and insert ??? If it is something else/more, then yes, we probably shouldn't be going with (that?) SPARQL now.
```
TallTed, line 54:
```suggestion
* KK: If you know what triple it is that you want to change, that is easy. It is also legitimate to change the triple and insert ??? If it is something else/more, then yes, we probably shouldn't be going with SPARQL now.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* RV: Open question is: is this suitable for 1.0 or go for undisputed approach for 0.9.
```
TallTed, line 56:
```suggestion
* RV: Open question is: is this suitable for 1.0, or go for undisputed approach for 0.9.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: Perhaps. Would you implement it in CSS?
+
+* RV: The reason Tim wanted this.. he said you have 3 weeks to do it. Not my proudest work but it is simple. It is extensible. The support for N3 libs and support for CSS would be trivial. It is a low hanging fruit with a lot of option for extensibility. Lowest common demoniator so to speak.
```
TallTed, line 60:
```suggestion
* RV: The reason Tim wanted this, he said you have 3 weeks to do it. Not my proudest work but it is simple. It is extensible. The support for N3 libs and support for CSS would be trivial. It is a low hanging fruit with a lot of option for extensibility. Lowest common denominator so to speak.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: Yes. That'd be good.
+
+ACTION: RV to make initial draft. KK to assign relevant issues to RV. KK to record 'something' on that issue.
+
+
+### Write methods as requirements
+URL: https://github.com/solid/specification/pull/304
+
+* KK: Also for testing for compliance... if we have to test for eventuality that Solid server that does not support write operations then that makes the test suite more complex. We would want to support servers that are read-only but that is mostly just a Web server right.. At some point we want to introduce conformance classes. In addition, we want to have language that allows servers for read-only. Which is also in the spec.. Therefore to not tackle more than this then I'd like write methods for required.
```
TallTed, line 94:
```suggestion
* KK: Also for testing for compliance... if we have to test for eventuality that Solid server does not support write operations, that makes the test suite more complex. We would want to support servers that are read-only, but that is mostly just a Web server, right? At some point we want to introduce conformance classes. In addition, we want to have language that allows servers for read-only, which is also in the spec. Therefore, to not tackle more than this then I'd like write methods for required.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* RV: That clarifies a lot. I can imagine read-only servers. How do we know if a solid server announces whether htey are read or write?
+
+* KK: accept headers.
+
+* RV: or status codes.
+
+* SC: Test suite shouldn't be driving factor for the requirement
+
+* KK: I disagree with that idea because. The test suite is important for the health of the ecosystem. As a principle that is important but if there is no easy way to automate the testing then it is a toy.. you don't have a production system you have a toy.
```
TallTed, line 104:
```suggestion
* KK: I disagree with that idea because the test suite is important for the health of the ecosystem. As a principle, that is important, but if there is no easy way to automate the testing then it is a toy. you don't have a production system, you have a toy.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* SC: spec comes before test.. even has test modes have different ways of checking.
```
TallTed, line 106:
```suggestion
* SC: spec comes before test... even has test modes for different ways of checking.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: there is principle and pragmatism. In general I agree to adopt the test to requirements. But if we go to the
```
TallTed, line 108:
```suggestion
* KK: there is principle and pragmatism. In general, I agree to adopt the test to requirements.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: I disagree with that idea because. The test suite is important for the health of the ecosystem. As a principle that is important but if there is no easy way to automate the testing then it is a toy.. you don't have a production system you have a toy.
+
+* SC: spec comes before test.. even has test modes have different ways of checking.
+
+* KK: there is principle and pragmatism. In general I agree to adopt the test to requirements. But if we go to the 
+
+* SC: then we are requiring automated test - that needs to be a specific requirement of what we expect from the requirement.
+
+* KK: what can we do that's testable in one week. instead of later on. if we want to do thta. so it does infuence a lot of stuff. I have spoken.
```
TallTed, line 112:
```suggestion
* KK: what can we do that's testable in one week, instead of later on, if we want to do that? so it does influence a lot of stuff. I have spoken.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* JB: Trying think of a case where
+
+* DZ: test sensor..
+
+* SC: some mentioned in the PR/issues.
+
+* JB: if you can 405 and not .. would you set out a case where you have a solid server and just have only read-only. or is it authz to make it read-only.
```
TallTed, line 120:
```suggestion
* JB: if you can `405` and not ??? , would you set out a case where you have a solid server and just have only read-only? or is it authz to make it read-only?
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: I suspect that you can configure apache with configuration only to be read-only. yes, that thing will exist. at some point we just need a conformance class to have a read-only then that is very
```
TallTed, line 122:
```suggestion
* KK: I suspect that you can configure apache to be read-only. yes, that thing will exist. at some point, we just need a conformance class to have a read-only
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* SC: ...
+
+* RV: use case is read-only / private.
```
TallTed, line 126:
```suggestion

* RV: use case is read-only / private.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: that makes it more complex to implement.
+
+* DZ: 405?
+
+* SC: yes
+
+* RV: but maybe it needs to be 403 not 405 if we say that write is required.
+
+* JB: okay with 405 and not against 403.
```
TallTed, line 136:
```suggestion
* DZ: `405`?

* SC: yes

* RV: but maybe it needs to be `403` not `405` if we say that write is required.

* JB: okay with `405` and not against `403`.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* RV: fine with it.
+
+* KK: language in my patch leaves it open so that can be defined in different terms.
+
+* DZ: i'm on the side of 405 because 403 notes to me with diff credentials it oculd work.
+
+* SC: 403 can be forbidden for any reason. not just authorization.
+
+* JB: in our case 403 seems more associated with authz butnot strictly either.
```
TallTed, line 146:
```suggestion
* DZ: i'm on the side of 405 because 403 notes to me with diff credentials it oculd work.

* SC: `403` can be forbidden for any reason, not just authorization.

* JB: in our case, `403` seems more associated with authz but not strictly either.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: the lang in the spec 
+
+* SC: 405 is cheap.
+
+* DZ: Might raise less support requests. 403 might raise more issues?
```
TallTed, line 152:
```suggestion
* SC: `405` is cheap.

* DZ: Might raise less support requests. `403` might raise more issues?
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* DZ: We are encouraging UCs to allow read-only eg. archives, right?
```
TallTed, line 154:
```suggestion
* DZ: We are encouraging UCs to allow read-only, e.g., archives, right?
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: I think current lang in the spec that allows that.
```
TallTed, line 156:
```suggestion
* KK: I think current lang in the spec allows that.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* DZ: that seems not a hard requirement. not difficult to implement.
+
+* SC: that was intended. 'when', these are the rules.
+
+* SC: {inaudible back and forth}
+
+* DZ: KK, what's the downside to this approach?
+
+* KK: ends up being tweaks in the Protocol. I think that to make this clearer, I'd introduce conformance classes and I don't want ot do that now. in that case read-only would be a conformance class
```
TallTed, line 166:
```suggestion
* KK: ends up being tweaks in the Protocol. I think that to make this clearer, I'd introduce conformance classes, and I don't want to do that now. in that case, read-only would be a conformance class
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: if we do that, how would i test it? it is hard to test things in the spec. that is a problem.
+
+* SC: i ack that concern.
+
+* KK: even if we make must or not it doesn't.
+
+* SC: but then people write the write operation bits but the code path will never be used.
+
+* JB: what if the wording in the PR said - keep it exactly as you wrote it for read-write as MUST - but have OR must respond with 405 if they don't support these methods. if you add that at the tail of each statement, then you're being more explicit. so someone that's reading it, if i want to make a read-only server i can be compliant by with these specific methods for reading-writing then i return 405.
```
TallTed, line 194:
```suggestion
* JB: what if the wording in the PR said - keep it exactly as you wrote it for read-write as MUST - but have "OR must respond with `405`" if they don't support these methods. if you add that at the tail of each statement, then you're being more explicit. so someone that's reading it, "if I want to make a read-only server, I can be compliant by returning `405` for these specific methods for reading-writing."
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* KK: that is possible.
+
+* SC: but we acknowledge that 405 can be returned for any reason (for any method ) for a particular resource.
+
+* KK: every test we do.. we can in principle.. and not fail.
```
TallTed, line 200:
```suggestion
* SC: but we acknowledge that `405` can be returned for any reason (for any method) for a particular resource.

* KK: every test we do, we can in principle, and not fail.
```

meetings/2021-11-03.md
```diff
@@ -0,0 +1,204 @@
...
+
+* SC: what would be the point of that?
+
+* JB: if you're reading/writing this section on the MusT, tailing the statement with or provides room to someone that they can do read as an option. the key there is the or.
```
TallTed, line 204:
```suggestion
* JB: if you're reading/writing this section on the `MUST`, tailing the statement with `OR` provides room to someone that they can do read as an option. the key there is the `or`.
```

TallTed (changes requested):

csarven:
Ruben, would you mind removing meetings/2021-11-03.md from this PR?

RubenVerborgh:
Seems like you added it, but sure.

csarven:
Thanks. That's strange. Where does it show that I've added? I did a force push but it had nothing to do with getting the minutes into this PR.

RubenVerborgh:
Force push to `main` probably created a situation where the `fix/clarify-containment-409` contained old `main` commits.

protocol.html
```diff
@@ -706,7 +706,7 @@ <h3 property="schema:name">Writing Resources</h3>
...
+                  <p><span about="" id="server-protect-containment" rel="spec:requirement" resource="#server-protect-containment"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> NOT allow HTTP <code>PUT</code> or <code>PATCH</code> on a container to update its containment triples; if the server receives such a request, it <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.</span></span> [<a href="https://github.com/solid/specification/issues/40#issuecomment-573358652" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
csarven, line 709:
```suggestion
                  <p><span about="" id="server-protect-containment" rel="spec:requirement" resource="#server-protect-containment"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST-NOT">MUST NOT</span> allow HTTP <code>PUT</code> or <code>PATCH</code> on a container to update its containment triples; if the server receives such a request, it <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.</span></span> [<a href="https://github.com/solid/specification/issues/40#issuecomment-573358652" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

csarven (approved):
I think my "review" was more or less captured in the comments. I'll just mention two bits:

>I'm okay to approve the PR if there is consensus to disallow POST to update container's description.

While there may be silent understanding or agreement by the group, I didn't see an explicit response besides the effect of this PR. So, taking that as is. Bottom line being:

>The PR actually removes the possibility of variable behaviour.

kjetilk:
OK, fair enough, l'll merge. 

Actually, in the meantime, @csarven had convinced me that we could well do without this, but lets get it in and have another look for 1.0, perhaps.

csarven:
>Actually, in the meantime, @csarven had convinced me that we could well do without this, but lets get it in and have another look for 1.0, perhaps.

YOU'VE GOT TO BE KIDDING ME.

RubenVerborgh:
There's always revert, peeps 🙂

PR #302, opened 2021-08-23
Add id attributes to server-side requirements that can be tested

kjetilk:
This does not change anything in the specification itself, it merely adds `id` attributes to requirements on the server that can be tested.

For now, I have left in particular auxiliary resources out, since we probably need to develop some more thinking around that, since it should be an extensible system. 

I have also not added the things that are in flux, such as notifications, or where testing should be done independently from the testing the protocol, for example authz.


protocol.html
```diff
@@ -406,15 +406,15 @@ <h2 property="schema:name">Hypertext Transfer Protocol</h2>
...
                 <div datatype="rdf:HTML" property="schema:description">
-                  <p>A <a data-link-type="dfn" href="#data-pod" id="ref-for-data-pod①">data pod</a> MUST be an HTTP/1.1 server [<cite><a class="bibref" href="#bib-rfc7230">RFC7230</a></cite>][<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]. It SHOULD additionally be an HTTP/2 server [<cite><a class="bibref" href="#bib-rfc7540">RFC7540</a></cite>] to improve performance, especially in cases where individual clients are expected to send high numbers of successive requests.</p>
+                  <p>A <span id="sr-http-http-11-server"><a data-link-type="dfn" href="#data-pod" id="ref-for-data-pod①">data pod</a> MUST be an HTTP/1.1 server [<cite><a class="bibref" href="#bib-rfc7230">RFC7230</a></cite>][<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]</span>. <span id="sr-http-http-2-server">It SHOULD additionally be an HTTP/2 server [<cite><a class="bibref" href="#bib-rfc7540">RFC7540</a></cite>]</span> to improve performance, especially in cases where individual clients are expected to send high numbers of successive requests.</p>
```
acoburn, line 409: Unrelated to this PR, but what is "①" character doing in the `id="ref-for-data-pod①"` element? Is that intentional?

kjetilk: I don't know, @edwardsph and I noted it too... We need to bring it to @csarven 's attention when gets back.

edwardsph (approved):
LGTM

protocol.html
csarven: "①" is garbage from bikeshed. Didn't get around to clean that up - along with adding ids all over the document.

csarven:
True that it doesn't change the requirements but it does put identifiers out there. Stuff links to it.. and when ids change, stuff "breaks". So, should really follow a proper pattern/Template. No big deal if we are willing to update stuff (like tests..)

I will follow-up with a PR that's a bit more thought out for requirements (IMO) - FWIW, WAC is the testing ground at the moment.

Issue #39, opened 2019-09-19
Which LDP features are required by Solid


acoburn:
[Section 2.2](https://solid.github.io/specification/#ldp) of the Solid specification requires (`MUST`) conformance with the [LDP specification](https://www.w3.org/TR/ldp/). The LDP specification, however, lists many features as optional.

LDP requires that a server support `GET`, `HEAD` and `OPTIONS`.

At what level of support does a Solid server implementation need to implement `POST`, `PUT`, `DELETE` and `PATCH`? It is worth noting that, via `OPTIONS`, a client can easily discover which methods a server supports for a given resource (via the `Allow` response header). It is also worth noting that `PATCH` is, effectively, an optimization of `PUT`.

RubenVerborgh:
My suggestion is to make `POST`, `PUT`, `DELETE`, and `PATCH` all MUST for Solid; except that of course a server is free to only offer read-only resources, in which case it would not need to support them.

I would also make `PATCH` a MUST because of concurrent editing use cases.

csarven:
Need to be clear on "support". Checking for support can be a grey area in that if `MUST` is given, server's capability needs to be essentially testable. Whether they offer resources as read-only is a separate case and we can incorporate that in the test. For example, the test can offer a check against read-only resources, and when tested as such, all `MUST`s would be expected to return a `405` and so it will *pass* the test, and fail otherwise.

csarven:
Just so that we document some reasons: an overarching reason as to why methods like `POST`, `PUT`, `DELETE` and `PATCH` should be required (at some level) for servers to support is that the *writing* - creating, deleting, updating - capability is central to the Solid ecosystem.

We need the servers at `MUST` to enable an even ground for clients. I don't think `SHOULD` will cut it.

csarven:
I tentatively suggest the following which extends LDP:

Generally:

`GET`: ...
`HEAD`: ...
`OPTIONS`: ...
`POST`: creating, appending, containment management
`PUT`: creating, replacing
`PATCH`: creating, updating
`DELETE`: remove, containment management

Specifically:

Intention for server-side containment management *SHOULD* be communicated through `POST` and `DELETE`. `PUT` and `PATCH` *SHOULD NOT* effect containment management.

Servers *MAY* use both the `Slug` and the `Link` header values in `POST` requests to determine hints for containment management. In the following example, the server uses the slashes in `Slug` as a separator for nested containers based on the interaction model(s) as specified in the `Link` header.

```
POST /
[...]

PR #304, opened 2021-08-30
Write methods as requirements

kjetilk:
The current spec does not require write methods to be implement. I disagree with this, as it introduces doubt as to what a Solid server really is. Solid should be very clear about that it is read-write. Without write, there's no social interaction, no setting of authorization, no notifications.  That is not to say that a read-only server isn't a part of the ecosystem, it is, but there's so little that is Solid-specific in a read-only scenario that it doesn't make sense IMHO to call it a Solid server. It'd still be a Web server, it'd still serve up resources that can be used on the Web and into Solid apps. It just isn't a Solid server. 

Therefore, I suggest we make the write methods `PUT`, `POST`, `PATCH` and `DELETE` required. If something does not implement all of these, it will not be a fully compliant Solid server, though it may be useful.


kjetilk:
BTW, in the back of my head, I had a comment in the back of my head that @timbl had expressed similar sentiments, and I finally [found a comment to the LDP comments list](https://lists.w3.org/Archives/Public/public-ldp-comments/2013Aug/0006.html):

>  You allow servers which do not support PUT or PATCH or POST.  Why?   A client using such a server will have no write  ability at all, and so your spec as a protocol delivers zero value on top of HTTP GET.  Suggest change all to MUST, or make two levels of server,  one which supports PUT and PATCH and no collections, and one which supports everything.

acoburn:
A point of clarification: must _all_ resources be read/write? Must only a subset of resources be read/write? Must there be at least one resource that is read/write?

pmcb55:
A big +1 to the clear intent behind this issue, although I'd tend to be a bit more nuanced in my support of the specifics - e.g., if I wanted to offer a "_Wayback-machine Pod service_", I may not want to offer any **DELETE** functionality whatsoever.
Likewise my Blockchain-backed Pod Server implementation might need to be cautious about a "**MUST** offer PATCH" requirement.
But yeah, absolutely - the general 'writeability' of a Solid Pod does indeed seem like a fundamentally defining feature, so it's apparent omission as a requirement does indeed seem like a great 'shout out' :) !

csarven:
This PR should respond to https://github.com/solid/specification/issues/39#issuecomment-907738789 .

kjetilk:
@acoburn & @pmcb55 I think that's covered by the sentence in the current spec that says:

> Servers MUST respond with the 405 status code to requests using HTTP methods that are not supported by the target resource.

That is, the requirement is on the server as a whole, if that particular target resource doesn't support a certain method, then it is legitimate to respond with a `405`. The spec also mentions the `Allow` header for that. Now, there are cases where none of the resources support a certain method, it doesn't change this, you will have an `Allow` header and respond with `405`.

@csarven I thought that I did, what do you feel I didn't touch upon? I'm concerned that if we do not emphasize the rw nature of Solid, we could end up as the early Web, where everything were designed read-write but that aspect was not appreciated by a lot of people, and therefore browsers were designed read-only.

kjetilk:
This is a source of pain, suffering and untold hours of madness while writing tests... 

I have added another commit, now to the security considerations, hopefully addressing the concern that deployments may want to restrict certain operations entirely.

protocol.html
```diff
@@ -902,6 +902,8 @@ <h3 property="schema:name">Security Considerations</h3>
...
 
                   <p>Servers are encouraged to use TLS connections to protect the contents of requests and responses from eavesdropping and modification by third parties. Unsecured TCP connections without TLS may be used in testing environments or when the data pod is behind a reverse proxy that terminates a secure connection.</p>
+
+		  <p>Even though servers must support write operations, not all deployments are required to allow write operations for all resources. For security, servers should indicate any such restrictions on HTTP methods by using the <code>Allow</code> header and should respond to methods they do not allow using the <code>405</code> status code for the relevant resources.</p>
```
csarven, line 906:
```suggestion
```

I don't see this as a security consideration. It is rather about affordances.

I think 405 is already sufficiently covered at the beginning of #reading-writing-resources:

>Servers MUST respond with the 405 status code to requests using HTTP methods that are not supported by the target resource.

protocol.html
```diff
@@ -625,7 +625,7 @@ <h3 property="schema:name">Reading Resources</h3>
               <section id="writing-resources" inlist="" rel="schema:hasPart" resource="#writing-resources">
                 <h3 property="schema:name">Writing Resources</h3>
                 <div datatype="rdf:HTML" property="schema:description">
-                  <p>When a server supports the HTTP <code>PUT</code>, <code>POST</code> and <code>PATCH</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] this specification imposes the following requirements: [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>]</p>
+                  <p>A server MUST supports the HTTP <code>PUT</code>, <code>POST</code> and <code>PATCH</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]. [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/pull/304" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
csarven, line 628: Minor: match to the language currently used for HTTP GET, HEAD and OPTIONS:

```suggestion
                  <p>Servers MUST support the HTTP <code>PUT</code>, <code>POST</code> and <code>PATCH</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]. [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/pull/304" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

Alternative that mentions the purpose (closer match to GHO text):

>Servers MUST support the HTTP <code>PUT</code>, <code>POST</code> and <code>PATCH</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] for clients to create or update resources or to process the request according to resource's semantics.

protocol.html
```diff
@@ -646,7 +646,7 @@ <h3 property="schema:name">Writing Resources</h3>
               <section id="deleting-resources" inlist="" rel="schema:hasPart" resource="#deleting-resources">
                 <h3 property="schema:name">Deleting Resources</h3>
                 <div datatype="rdf:HTML" property="schema:description">
-                  <p>When a server supports the HTTP <code>DELETE</code> method [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] this specification imposes the following requirements: [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>]</p>
+                  <p>A server MUST support the HTTP <code>DELETE</code> method [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]. [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/pull/304" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
csarven, line 649: Minor: match to the language currently used for HTTP GET, HEAD and OPTIONS:

```suggestion
                  <p>Servers MUST support the HTTP <code>DELETE</code> method [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]. [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/pull/304" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

Alternative that mentions the purpose (closer match to GHO text):

>Servers MUST support the HTTP <code>DELETE</code> method [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] to remove resources.

csarven (changes requested):
Going to quote myself from https://github.com/solid/specification/issues/39#issuecomment-907738789 because it requests clarity as to why the bar must be higher. Given two different ways to implement (one server with write capabilities and other not), and that they can both return 405. Some servers will never need to allow write requests (and so blocking immediately with 405) - e.g., archives, backups, audit logs, provenance records, memento resources.. - meanwhile still implementing an authorization system that would only need to allow read operations for specific resources to specific agents/clients:

>While I still think that servers MUST implement PUT, POST, PATCH, DELETE, I can also see that there is room for servers only implementing GET, HEAD, OPTIONS and participating in the ecosystem. I don't see significant or practical difference between a server implementing the PPPD and GHO methods but configured by its owner to return 405 on PPPD, with a server implementing only GHO and returning 405 for PPPD. GHO-only is a low bar entry and supports sufficient use cases, quickly implemented, run on environments with minimal resources, and can always be upgraded to support PPPD.

>I think authorization system is not particularly relevant here. The system doesn't need to resort to the authorization system to disallow write operations. That would only cover user/client level access control with 403, not 405.

protocol.html
kjetilk: Alright, it certainly works for me.

RubenVerborgh (approved):

justinwb (approved):

protocol.html
```diff
@@ -625,7 +625,7 @@ <h3 property="schema:name">Reading Resources</h3>
               <section id="writing-resources" inlist="" rel="schema:hasPart" resource="#writing-resources">
                 <h3 property="schema:name">Writing Resources</h3>
                 <div datatype="rdf:HTML" property="schema:description">
-                  <p>When a server supports the HTTP <code>PUT</code>, <code>POST</code> and <code>PATCH</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] this specification imposes the following requirements: [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>]</p>
+                  <p>Servers MUST support the HTTP <code>PUT</code>, <code>POST</code> and <code>PATCH</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]. [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/pull/304" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
TallTed, line 628:
```suggestion
                  <p>Servers MUST support the HTTP <code>PUT</code>, <code>POST</code>, and <code>PATCH</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>]. [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/pull/304" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

kjetilk: I think I shall leave it to native English speakers to debate that comma. :-)

csarven:
This PR was discussed in this meeting: https://github.com/solid/specification/blob/main/meetings/2021-11-03.md#write-methods-as-requirements .

In summary:

The underlying argument to require all Solid servers to have write-capability was due to a particular testing architecture/software requiring changes that would be too much of a task to have it ready for the 0.9 release of the Protocol specification.

While there was mutual understanding that specs, as opposed to testing software, drives the requirements, the release of a spec version for "pragmatic" reasons was deemed to be an exception by some of the editors.

The possibility to allow server implementations where their sole purpose is to be read-only without having any code paths for write operations and still be acknowledged to be part of the Solid ecosystem did not receive legitimate response or deemed to have merit by some of the editors.

No resolution was recorded for the 1.0 release.

---

For the record, I'm not satisfied with the arguments given to dismiss read-only servers. As mentioned, there are legitimate reasons to require write but not at the expense of dismissing use cases where only read is required. The merge of this PR also goes contrary to enabling an inclusive and equitable ecosystem. The Solid Protocol mentions how we use the Ethical Web Principles to orient ourselves. As is stands, the first sentence in the Introduction of the Solid Protocol:

>The aims of the Solid project are in line with those of the Web itself: empowerment towards "an equitable, informed and interconnected society".

I will follow with a PR to reverse this for the 1.0 release, and hope that specific testing software and/or lack of resources to develop them, does not dictate the behaviour of the Solid ecosystem by that point.

csarven:
As discussed in Editors meeting https://github.com/solid/specification/blob/main/meetings/2021-11-10.md#write-methods-as-requirements , it turns out that this PR was accidentally merged. The agreement was to have this as is for 0.9 - given that there is some approval -  but that we'll revisit for 1.0.

Issue #305, opened 2021-08-30
Append operations on non-container resources


kjetilk:
Append operations on non-container resources are currently not defined in the spec, and it is clear that what it means to append to a resource will depend on the media type of the target resource. There are some media types that have a straightforward interpretation, among them notably RDF sources. I therefore suggest that we have generic language in the spec and that we define append operations for certain media types, in the continuation of #118.

Moreover, that an append operation can be done using `POST` is very
clearly stated in [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.3):

> Appending data to a resource's existing representation(s).

It has actually been in since the very first description of [HTTP](https://www.w3.org/Protocols/HTTP/Methods/Post.html):

> Extending a document during authorship. 

So, given that simply adding some data is a common requirement, we have an append operation in Solid, a clear description in HTTP and a straigthforward interpretation in RDF, it seems very clear that we should do it.

I therefore suggest that we add something like this to the Protocol:

## Append operations to a non-container resource

When a `POST` method request targets a non-container resource with an existing representation, the server MUST interprete that as an Append operation. The behaviour of an append operation depends on the media type. Servers MAY implement append algorithms for any media type. If the request contains a media type that the Server has no append algorithm for, it MUST respond with the `415` status code. Servers MUST implement the two following append algorithms:

### Media types for RDF serializations

If the target resource and the request payload has a RDF bearing representation, then the server MUST use the [RDF Merge](https://www.w3.org/TR/rdf11-mt/#shared-blank-nodes-unions-and-merges) algorithm to append the request payload into the target resource. This is equivalent but simpler than to `PATCH` the target resource using `INSERT DATA` from SPARQL Update.

### `text/plain` media type.

If the target resource and the request payload has the `text/plain` media type, then the server MUST append the request payload to the end of the target resource.

jeff-zucker:
This sounds very useful.  How does this relate to the property of POST that the server can always assign  name?  If I append to foo.ttl, I would want to expect that foo.ttl is where I can retrieve the merged data.

kjetilk:
Yes, this is `POST` to an existing non-container resource (`POST`-to-append, we might call it), it is different from the `POST`-to-create. With `POST`-to-create, you `POST` to a container, and the server assigns a name, or takes a suggestion that you supply with the `Slug` header.

With `POST`-to-append, you `POST` to an existing resource, and then it doesn't create, and so, the server does not assign any name, it is strictly a modification of an existing resource.

[...]

Issue #198, opened 2020-09-16
Aligning representations of document and container resources with REST via single and compound state


RubenVerborgh:

acoburn:
I am very much in favor of this proposal and I agree with all of the normative language provided here.

csarven:
I agree that this proposal ties up roughly agreed conclusions from issues and current spec/PRs.

I find the normative text and examples above on HTML+RDFa clear. In the PR to follow, I'd however suggest that we place emphasis on RDFa without requiring a particular host language - as per RDF 1.1 Concepts. So then for example, a container resource could be served with an SVG+RDFa representation.

elf-pavlik:
Would this approach still work if instead of HTML+RDFa someone uses plain HTML with script tags embedding Turtle or JSON-LD ? From content negotiation those two cases seem identical since both use `text/html` as content type.

elf-pavlik:
> So then for example, a container resource could be served with an SVG+RDFa representation.

How would the client know that `image/svg+xml` is available? Would it rely on some prior knowledge or when available server would advertise it in HTTP header with

```http
Link: <https://solid.example>; rel="preconnect"; type="image/svg+xml"
```

csarven:
@RubenVerborgh 

>but it SHOULD represent the same RDF graph.

I suggest MUST.

>SHOULD include any part of the client-managed state that can be represented as RDF

would not ensure documents at `/` eg. profiles, slideshows, be usable by all clients eg. for authentication. *If* server accepts resource state for the client-managed part of a container resource, then using MUST (instead of SHOULD) would persist RDF encoded information (in RDF documents) across RDF representations. Adding a human-readable label of a container for example. This is also aligned with https://github.com/solid/specification/pull/196

csarven:
I'd like to clarify/propose a bit to above because some key aspects of the container representations are not quite jumping out but the good news is that the fundamental requirements are already in place.

Re spec:

>When a server creates a resource on HTTP PUT, POST or PATCH requests such that the request’s representation data encodes an RDF document (as determined by the Content-Type header), the server MUST accept GET requests on this resource when the value of the Accept header requests a representation in text/turtle or application/ld+json

>Servers MUST NOT allow HTTP POST, PUT and PATCH to update a container’s containment triples; if the server receives such a request, it MUST respond with a 409 status code.

>The representation and behaviour of containers in Solid corresponds to LDP Basic Container and MUST be supported

[...]

PR #309, opened 2021-09-14
Add that all RDF representations should represent the same graph

kjetilk:
I just occurred to me (in #305) that we do not seem to say that all RDF representations must represent the same graph, it seems to have been an implicit assumption. This is meant to make it explicit.


csarven (approved):
Perhaps an assumption to some but I felt that I had to regularly state/argue this in issues, e.g. https://github.com/solid/specification/issues/60#issuecomment-538350810 , many discussions including RDFa (somehow people wanted to classify RDFa as non-RDF, despite its name or in fact listed as a *concrete RDF syntax*), or even whether LDP-RS description made any sense...

..which is also consistent with my comment in 309 re "RDF document" as per RDF 1.1 Concepts".

/me takes a breath.

I don't object to including this statement in the PR but okay with it being included. I don't actually think it needs to be stated. It is a given if RDF 1.1 Concepts is used normatively in the Protocol - which it is.

RubenVerborgh (changes requested):
I need to vote against actually, for a major and a minor reason

1. major: content negotiation could happen in other dimensions (profile etc.), leading to representations of the same conceptual resource in different ontologies and/or shapes; these _are_ different RDF graphs, but can still be valid representations. I also point here to a recent discussion where we were considering different representations for containers at some point, which would have contained different RDF graphs.
2. minor: if the resource has RDF 1.1 named graphs, it cannot be adequately represented in RDF 1.0 formats, so subsetting should happen.

kjetilk:
> I need to vote against actually, for a major and a minor reason
> 
>     1. major: content negotiation could happen in other dimensions (profile etc.), leading to representations of the same conceptual resource in different ontologies and/or shapes; these _are_ different RDF graphs, but can still be valid representations. I also point here to a recent discussion where we were considering different representations for containers at some point, which would have contained different RDF graphs.

Right, there is some tension here. However, part of it is stretching the meaning of a representation. If profiles, `Prefer` headers, etc, select some subset of the resource, it is not a representation of the resource IMHO, in the same way that if the resource is a fruit basket, a picture of a banana is not a representation of that fruit basket. RFC7240 seems to support this view. IIRC, Fielding did not consider subsets in his REST definition, only conneg over full representations of the state resource.

A transformation is different, I can see that representing the graph in a different ontology could arguably be such a case. Pragmatically, would you actually do this? I could see a `CONSTRUCT` query being used for this purpose, but server side, that would be a different resource, and client side, it wouldn't matter. Perhaps it is in fact appropriate to say that this would be a different resource?

Now, the situation I'm trying to avoid here is that people write servers where the client has to update both the JSON-LD and the Turtle representation of the same resource. It is not sufficiently clear from the current spec that you shouldn't do that.

How would you formulate a requirement to resolve that problem? 


> 
>     2. minor: if the resource has RDF 1.1 named graphs, it cannot be adequately represented in RDF 1.0 formats, so subsetting should happen.

If that happens somewhere down the road, then the definition needs changing, but that would be a substantial change to Solid as currently defined, and should not be in scope for 1.0 IMHO.

RubenVerborgh:
> but server side, that would be a different resource

> Perhaps it is in fact appropriate to say that this would be a different resource?

Not necessarily; if I define `/foo/bar` to be "Ruben Verborgh's profile", then many representations in many ontologies could be acceptable. If I define it as "Ruben Verborgh's profile in Schema.org", that's something else; and so is "Ruben Verborgh's profile in Schema.org in Turtle". Note that "being a representation of" is a relationship, not an absolute. See https://rubenverborgh.github.io/WebFundamentals/web-apis/#conceptual-resource, https://rubenverborgh.github.io/WebFundamentals/web-apis/#representations, and beyond (but yeah, those are written by me, so not the best source to back up my own point 🙂).

kjetilk:
> Not necessarily; 

I'm certainly fully aware of that. There has always been room for a certain flexibility in this area, when should two language negotiated representations actually be different resources, etc. To make that not black and white, we have `q` factors and all that. Similarly, ontologies may have substantial alignment problems that would raise similar concerns. There is substantial room for pragmatism here, but perhaps that room should be restricted.

> 
> > Now, the situation I'm trying to avoid here is that people write servers where the client has to update both the JSON-LD and the Turtle representation of the same resource.
> 
> Then write that, I would say. Updating one representation invalidates all others.

"Invalidate" would not be my favorite word to describe that, then we'd have to say what happens after invalidation, and we shouldn't need to get into implementation details. This would have to be formulated more generally, it goes for any serialization. 

Alright, I hoped this was low-hanging fruit, lets leave it and get back to it in the ordinary process.

kjetilk:
After discussing with @csarven , it is unlikely to be a practical problem in near future, so we don't need to prioritize this now.

kjetilk:
This issue _is_ itching me a bit. I think we need to define it. 

A pragmatic problem is for example that a server applies a transformation to the FOAF data using some mechanism upon a `GET` request, and then the client `PUT`s that right back at the server. Then, suddenly, the ontology which you'd like to use for the representation of yourself has changed. 

Then, what if your data initially included both `foaf:givenName` and `foaf:familyName` and then, what you get back is `schema:name`, so you had data loss. You could argue that this problem is unavoidable unless you implement input data (shape) validation, and yes that's true. 

We could get into deeply philosophical arguments over whether a boat that has replaced all the planks is still the same boat. However, I'd like to keep it pragmatic, because I suspect the way that you'd do this is that there is an abstract representation (which is what your sensor returns or what you keep in your database or on file), but you allow people to read that representation in a variety of concrete syntaxes, and apply various queries and transformations over them. You allow writes using different concrete syntaxes to the same resource, but you don't really want clients to write back the result of the transformation to the resource.

You could, though, if that transformation had its own URI, so that's what I think we should be looking at. It is also an extension of the discussion of origin-form: #368 . If we did define (at least) the resources in terms of absolute-form, then, any transformation (e.g. using a different ontology over the same data), could return a response with a `Content-Location` header which may use origin-form to identify the actual representation? 

Personally, I think I would make that representation read-only, but that could be left as an implementation detail. You could allow the server to transform back upon writes if you had an appropriate ontology alignment engine.

There is a lot of complexity that you could run into by keeping this too open an philosophical for little practical value, and I think therefore makes sense to constrain it to keep it pretty straightforward.

kjetilk:
Actually, I think I'll close this and instead have a more generic discussion in #198 .

Issue #125, opened 2019-11-26
Level of SPARQL Update support


kjetilk:
A certain level of SPARQL Update support is [expected in Solid](https://github.com/solid/solid-spec/blob/master/api-rest.md#alternative-using-sparql-1), to be used with the `PATCH` method (#85). This discussion has [begun](https://github.com/solid/solid-spec/issues/159), and this issue is to discuss some details that we need to decide upon. 

The main questions are:
1. What subset of SPARQL Update is suited as a minimal requirement for Solid?
1. How does WAC apply to the minimal subset? 
1. How does WAC apply to SPARQL Update for implementations that will use a fully compliant SPARQL implementation?
1. What would be the URI of SPARQL Endpoint(s)?
1. Should any SPARQL Update operations be forbidden?
1. Should complex HTTP Verbs rather be SPARQL operations?

## A short overview of [SPARQL Update](https://www.w3.org/TR/sparql11-update/)

SPARQL Update has three operations that are of relevance to us, `INSERT DATA`, `DELETE DATA`, and `INSERT`/`DELETE`. SPARQL always operates over quads and quad patterns, whether they are quads that are passed directly as data to the two former operations, or used with the keywords `WITH` and `USING` to the `INSERT`/`DELETE` operation. In the context of Solid, each resource is represented with triples, and since the graph part is optional, we can safely ignore it for now.

`INSERT DATA` takes triples in their curly brackets, and [RDF merges](https://www.w3.org/TR/rdf11-mt/#shared-blank-nodes-unions-and-merges) the triples into the resource. `DELETE DATA` deletes exactly the triples that it has in curly brackets if they exist. It is important to note that **there are no variables** with these two operations, so no pattern matching is going on. They are the simplest forms of SPARQL Update.

If pattern matching is required, i.e. you need variables, then those variables goes in a `WHERE` clause, and thus, the more advanced `INSERT`/`DELETE` operation must be used. 

## Minimal SPARQL Update requirement

[...]

PR #320, opened 2021-10-11
Drafting SPARQL Update support

kjetilk:
Here is a first pass over #125 with a suggested BNF and some text.

I'll mark it WiP, because I have a feeling we should be able to simplify it further especially around `GroupGraphPattern`, which shouldn't need that, it seems.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -1062,12 +1062,22 @@ access-mode      = "read" / "write" / "append" / "control"</pre>
           <section class="appendix" id="sparql-update-subset" inlist="" rel="schema:hasPart" resource="#sparql-update-subset">
             <h2 property="schema:name">Appendix - A subset of SPARQL Update for Solid</h2>
             <div datatype="rdf:HTML" property="schema:description">
+              <section id="sparql-subset-overview" inlist="" rel="schema:hasPart" resource="#sparql-subset-overview">
+                <h3 property="schema:name">Overview</h3>
+		 <div datatype="rdf:HTML" property="schema:description">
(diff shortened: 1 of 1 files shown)


ericprud:
> Should we already mention something about atomicity?

This text precludes one way to provide atomic updates over multiple Resources (e.g. to keep them in a mutually-consistent state):

> except that servers MUST NOT allow a request with a `PATCH` method to change other resources than the target resource. [[Source](https://github.com/solid/specification/issues/125#issuecomment-873035679)]

@RubenVerborgh , is that what you meant by "atomicity"?

(Precluding could be a good thing 'cause it leaves us room to spec it later.)

csarven:
FWIW, HTTP, LDP specs use ABNF. wac-allow, Accept-Put is also in ABNF.

kjetilk:
> The particular subset looks good to me; I'm just not sure whether BNF is the most sustainable way to capture this (see other comment).
> 
> Should we already mention something about atomicity? Note that NSS implements a subset of SPARQL 1.1 _with a different semantics_, namely it requires exactly one match for the `WHERE` clause.

I need to re-review the NSS code on this point. I am aware of the willful violation of SPARQL when it comes to the semaphore mechanism, but I thought the matching was connected to that.

kjetilk:
> Which makes me wonder… is the semaphore mechanism the only reason? If so, Read should probably be dropped until we agree on the semaphore.

Yes, it is the only reason, AFAICT. Standard SPARQL doesn't have this problem since it always succeeds. But the semaphore is required for 0.9, so we might as well have it.

Issue #19, opened 2019-07-29
Proposal: Support COPY use case


dmitrizagidulin:
(This is a more detailed proposal continuation of issue https://github.com/solid/solid/issues/49.)

### Motivation / Problem Statement

We need a bandwidth-efficient method to copy data to and from Solid pods.

Imagine you're building a 'Save to Solid' widget / app / browser extension. The idea is - the user is browsing some Web resource (a PDF file, or an image, or a video, etc), and would like to save it to their pod (to be able to tag it and do other sorts of CMS stuff on it).

Currently, the only way to perform this operation would be multi-step:

1. GET resource and somehow store it temporarily in the client (in browser memory?)
2. PUT the temp resource on the pod (upload it)

Putting aside the implementation details, this presents two problems: temporary storage space (if the file is being held in a Javascript variable in a web app, or in LocalStorage, this quickly presents challenges when the resource is large), and bandwidth.

This is especially problematic on resource-constrained clients (like mobile apps or browsers) -- the user first has to use their mobile data to download a file temporarily, and then use mobile data to upload that resource to their pod.

### Proposed Solution 1: `COPY` Method

Note: As @RubenVerborgh points out, we should separate the problem / use case from the proposed solution.

Add a new Solid-specific LDP method, `COPY`, inspired by the existing [WebDAV COPY method](http://www.webdav.org/specs/rfc4918.html#METHOD_COPY).

This proposed solution lets pods play to their strength, to act as always-connected high(er) bandwidth servers. Specifically, it allows a client to issue a single COPY command, and the server would perform the necessary data transfer (using its own bandwidth, not the client's).

This would be just a single step:

1. issue a command to COPY resource from source URL to destination URL (server would proceed with the operation).

#### COPY Example

To copy FROM an external URL, `https://example.com/example.pdf`, TO a user's POD, `alice.inrupt.net`:

```
COPY /papers/example.pdf HTTP/1.1
Host: alice.inrupt.net
Source: https://example.com/example.pdf
```

```
HTTP/1.1 201 Created
Date: Mon, 23 May 2019 22:38:34 GMT
Content-Type: application/pdf
Location: https://alice.inrupt.net/papers/example.pdf
```

#### `COPY` Method Specs

This method is idempotent, but not safe (see [Section 9.1 of RFC2616](https://tools.ietf.org/html/rfc2616#section-9.1)). Responses to this method must not be cached.

#### Copying non-Container resources

[...]

PR #323, opened 2021-10-15
Behavior for other SPARQL Update operations

kjetilk:
As noted in #125, SPARQL Update contains several operations beyond DELETE and INSERT. It has been an open question what to do with them. Since they are not currently supported and thus falls outside of the supported subset, I suggest we simply disallow them on a SHOULD level. 

I think we should review this soon with respect to `COPY` and `MOVE` in light of #19 , but I think this is the most straightforward way to address it right now.

I made this a separate PR to not tie up #320, but I hope we can get it in for completeness.


protocol.html
```diff
@@ -709,7 +709,7 @@ <h3 property="schema:name">Writing Resources</h3>
...
-
+		    <span about="" id="server-patch-sparql-not-update" rel="spec:requirement" resource="#server-patch-sparql-not-update"><span property="spec:statement">SPARQL Update [<cite><a class="bibref" href="#bib-sparql-overview">SPARQL</a></cite>] also defines operations <code>LOAD</code>, <code>CLEAR</code>, <code>CREATE</code>, <code>DROP</code>, <code>COPY</code>, <code>MOVE</code> and <code>ADD</code>. <span rel="spec:requirementSubject" resource="spec:Server">Servers</span> that receives a <code>PATCH</code> request body containing these operations <span rel="spec:requirementLevel" resource="spec:SHOULD">SHOULD</span> respond with a <code>422</code> status code [<cite><a class="bibref" href="#bib-rfc4918">RFC4918</a></cite>] and a message body that explains the error, or MUST ensure that any access control measures apply.</span></span>
```
dmitrizagidulin, line 712: > or MUST ensure that any access control measures apply.

The 'or' there is slightly confusing -- shouldn't that be an 'and'?

dmitrizagidulin (approved):
:+1:

protocol.html
TallTed, line 712:
```suggestion
		    <span about="" id="server-patch-sparql-not-update" rel="spec:requirement" resource="#server-patch-sparql-not-update"><span property="spec:statement">SPARQL Update [<cite><a class="bibref" href="#bib-sparql-overview">SPARQL</a></cite>] also defines operations <code>LOAD</code>, <code>CLEAR</code>, <code>CREATE</code>, <code>DROP</code>, <code>COPY</code>, <code>MOVE</code> and <code>ADD</code>. <span rel="spec:requirementSubject" resource="spec:Server">Servers</span> that receives a <code>PATCH</code> request body containing these operations (1) <span rel="spec:requirementLevel" resource="spec:SHOULD">SHOULD</span> respond with a <code>422</code> status code [<cite><a class="bibref" href="#bib-rfc4918">RFC4918</a></cite>] and a message body that explains the error, and (2) MUST ensure that any access control measures apply.</span></span>
```

michielbdejong: I think it's an OR, not an AND.
EITHER ensure measures apply OR respond with an error.
Not both.

TallTed: I don't think "EITHER MUST ensure measures apply OR SHOULD respond with an error" makes sense. Does one of these work? 
* "MUST EITHER ensure measures apply OR respond with an error"
* "SHOULD EITHER ensure measures apply OR respond with an error"

Issue #322, opened 2021-10-14
Implications of semaphore mechanism for SPARQL Update


kjetilk:
While writing up a text around the semaphore mechanism from [the RWW Design Issue](https://www.w3.org/DesignIssues/ReadWriteLinkedData.html), also discussed in #125 , https://github.com/solid/solid-spec/pull/193 , https://github.com/solid-archive/query-panel/issues/3 and others, I found that I have not yet fully understood the problem. 

Is it also the idea that the same mechanism can be used in a situation where there isn't actual concurrency? I could see a case for that, say that one client `GET`s a certain resource, looks at it, and then wants to change something using a `PATCH` but meanwhile another client has changed the same? This is more the archetypal case for conditional requests, so I didn't think of it as a valid situation for the semaphore mechanism, but I guess a case for that this is actually a simpler mechanism could be made, as maintaining validators are more work for servers, and since it is a validator for the entire document, it scales badly if many people are co-editing a large document. 

This issue has usually been discussed in terms of _atomicity_, including [rdflib's documentation](https://github.com/linkeddata/rdflib.js/blob/8b405a1cef8a576cad0b1d4233201bd40e45b959/Documentation/webapp-intro.html#L473). I have not quite managed to understand that angle in light of the SPARQL Update spec, [which says](https://www.w3.org/TR/sparql11-update/#deleteInsert):

> If any solution produces a triple containing an unbound variable or an illegal RDF construct, such as a literal in a subject or predicate position, then that triple is not included when processing the operation: INSERT will not instantiate new data in the output graph, and DELETE will not remove anything. 

In our case, it means that if the `WHERE` clause doesn't match anything, the variable will be unbound and so nothing will be deleted. In other words, it sounds like we have standard behavior from the _query language_. 

However, it is true that no error is raised by standard SPARQL, as a measure to not leak information. Thus, we must expect that standard SPARQL implementations will need possibly substantial modifications to accommodate for this case now.

To accommodate for this situation on the Solid protocol, I wonder if it would suffice to say something like 

"If any solution produces a triple containing an unbound variable or an illegal RDF construct, then the server MUST abort any modifications and respond with a `409` status code."

That seems like a relatively minor change compared to standard SPARQL. This wouldn't be a violation of the language, AFAICS, it is a protocol thing.

I have only seen the semaphore mechanism mention in connection to the `DELETE INSERT WHERE` case, does that mean we only have to consider it in that context? 

[...]

PR #324, opened 2021-10-15
Add possible semaphore mechanism wording

kjetilk:
I'm proposing a draft of the wording I proposed in #322. I think this is the minimal change we can do to SPARQL that also legitimizes the behaviour in NSS with regards to the semaphore mechanism.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -713,7 +713,7 @@ left:4.5em;
 		  </p>
 
 		  <p>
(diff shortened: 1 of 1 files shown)


kjetilk:
> Should this be "if" or "when"? (I know sparq11-update says 'if' but context may be a bit different here.) Trying to first get a sense of where you want to go with this.

Yes, I couldn't see the practical difference, and so I chose to align with the terms that was used in most of the sentence.


> 
> Didn't make this change but would consider whether the language should be closer to "abort the sequence of operations, causing the subsequent operations to be ignored" instead of "any modification".

As explained in #322 , this is a different thing, that applies when the request contains several operations, in this case `DELETE INSERT WHERE` is a single operation, so it doesn't apply. However, I recognize that "any modification" might not be sufficiently precise, as there might be side effects, you could have an audit log for example that is modified as a result of an aborted operation. I'm not sure that should tie us up right now though. I think we should stay focused on what would be sufficient for defining the current NSS behavior without departing too far from SPARQL.

Issue #301, opened 2021-08-18
Slash semantics and conflicting requests on resource creation


kjetilk:

kjetilk:
We're working on converting tests for the new harness, and without going into too much detail on that work-in-progress, we have noted that there is a lot of divergence on how the implementations deal with the tension between slash semantics and LDP Interaction Models. I think this means that the heuristics from #128 is not sufficiently clear. My main fear is that since there is quite a lot of assumptions connected to containers and resources in the access control, lack of clarity in this area could result in opening attack vectors that we can't imagine right now. So, the intention isn't to change the good path (i.e. these are errors an end-user should never see), but to define error behaviour in the other cases, e.g. when something is declared a Non-RDF resource but doesn't have a `/`.

In https://github.com/solid/specification/issues/40#issuecomment-566995240 I tried to introduce the two concepts of _consistency_ and _exactness_. I still think they are useful and that we need them.

By consistency, I mean that different parts of the request carries the same semantics. If something has a `Link` header that says something different from the request URI, there are a conflicting statements and so, there is an error. There are several sources of information that could cause these conflicts, but to advance this issue, I think we should consider three, namely the possible conflict between LDP Interaction Models as given by a `Link` header, the `Content-Type` header that can distinguish between RDF Sources and Non-RDF Sources and the request URI, which can determine if the resource is a container.

As previously mentioned, by exactness I mean to constrain the freedom the server has to adapt the request. Exactness is a criterion that applies differently to `POST` than to other HTTP verbs. With `PUT` request, the HTTP spec affords very little freedom, from [RFC7221](https://www.rfc-editor.org/rfc/rfc7231.html#page-28):

> Proper interpretation of a PUT request presumes that the user agent knows which target resource is desired.  A service that selects a proper URI on behalf of the client, after receiving a state-changing request, SHOULD be implemented using the POST method rather than PUT.  If the origin server will not make the requested PUT state change to the target resource and instead wishes to have it applied to a different resource, such as when the resource has been moved to a different URI, then the origin server MUST send an appropriate 3xx (Redirection) response; the user agent MAY then make its own decision regarding whether or not to redirect the request.

So, it is not appropriate to change a request URI to fit a different interaction model when `PUT` is used, I think, as that implies it is a different resource than the UA intended to make.

[...]

PR #341, opened 2021-10-29
Tighten consistency requirements for resource types

kjetilk:
With the sound of the deadline of the milestone approaching with rising pitch, I decided to draft what I think is coming out of #301. There doesn't seem to be significant opposition to most of this given the test results, with the possible exception of the status code. I put `409` in anyway. We can conclude that debate in #301. 

The PR should address both my initial concern, and @langsamu's first concern of changing resource types. The text isn't very extensive, and does not cover it in as much detail as we did on the issue, but I think we can enforce this by encoding the status codes in tests.


protocol.html
```diff
@@ -665,6 +665,22 @@ <h3 property="schema:name">Resource Type Heuristics</h3>
...
+		    <span about="" id="server-reject-inconsistency" rel="spec:requirement" resource="#server-reject-inconsistency"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> reject inconsistent requests using a <code>409</code> or <code>415</code> status code as per Section 4.3.4 of [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>].</span></span>
+		    <span about="" id="server-existing-type-inconsistency" rel="spec:requirement" resource="#server-existing-type-inconsistency"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="spec:Server">server</span> recieves a request that attempts to change the resource type, it <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> must verify that the resource type of the request is compatible with the current resource type and reject the requests using a <code>409</code> status code if it is not.</span></span>
```
acoburn, line 671:
```suggestion
		    <span about="" id="server-existing-type-inconsistency" rel="spec:requirement" resource="#server-existing-type-inconsistency"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="spec:Server">server</span> recieves a request that attempts to change the resource type, it <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> verify that the resource type of the request is compatible with the current resource type and reject the requests using a <code>409</code> status code if it is not.</span></span>
```

There is an extra "must" in this line.

RubenVerborgh:
```suggestion
		    <span about="" id="server-existing-type-inconsistency" rel="spec:requirement" resource="#server-existing-type-inconsistency"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="spec:Server">server</span> receives a request that attempts to change the resource type, it <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> must verify that the resource type of the request is compatible with the current resource type and reject the requests using a <code>409</code> status code if it is not.</span></span>
```

protocol.html
```diff
@@ -665,6 +665,22 @@ <h3 property="schema:name">Resource Type Heuristics</h3>
...
+		     [<a href="https://github.com/solid/specification/issues/301" rel="cito:citesAsSourceDocument">Source</a>]
+		  </p>
+
+                  <div class="note" id="resource-type-interpretation" inlist="" rel="schema:hasPart" resource="#resource-type-interpretation">
+                    <h4 property="schema:name"><span>Note</span>: Interpretation of Resource Types</h4>
+                    <div datatype="rdf:HTML" property="schema:description">
+                      <p>
+			Clients and servers may indicate resource types using a HTTP <code>Link</code> headers with <code>rel="type"</code>, but should be aware that they come with assumptions as to what media types can be associated with them. In particular, Solid requires containers to be represented by a <em>RDF Document</em>, which again requires an appropriate RDF serialization with a corresponding media type for consistency. Other media types may have similar consistency requirements.
```
RubenVerborgh, line 679:
```suggestion
			Clients and servers MAY indicate resource types using a HTTP <code>Link</code> headers with <code>rel="type"</code>, but should be aware that they come with assumptions as to what media types can be associated with them. In particular, Solid requires containers to be represented by an <em>RDF Document</em>, which again requires an appropriate RDF serialization with a corresponding media type for consistency. Other media types may have similar consistency requirements.
```

protocol.html
```diff
@@ -665,6 +665,22 @@ <h3 property="schema:name">Resource Type Heuristics</h3>
...
 
+		  <p>
+		    <span about="" id="server-ensure-consistency" rel="spec:requirement" resource="#server-ensure-consistency"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> only accept requests where there is consistency between the path ending (<a href="#uri-slash-semantics">URI Slash Semantics</a>), any HTTP <code>Link</code> headers with <code>rel="type"</code> and the media type given in the <code>Content-Type</code> header of the request.</span></span>
+		    <span about="" id="server-reject-inconsistency" rel="spec:requirement" resource="#server-reject-inconsistency"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> reject inconsistent requests using a <code>409</code> or <code>415</code> status code as per Section 4.3.4 of [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>].</span></span>
```
RubenVerborgh, line 670: Maybe provide a small hint like

> 415 (for media type inconsistency) or 409 (for all other inconsistencies)

RubenVerborgh (approved):
LGTM; I'm still in favor of `400`, but let's settle that in #301 indeed.

kjetilk (commented):
Responses to @RubenVerborgh . The github workflow isn't very handy at these short things :-)

protocol.html
RubenVerborgh: OK!

protocol.html
```diff
@@ -665,6 +665,22 @@ <h3 property="schema:name">Resource Type Heuristics</h3>
...
 
+		  <p>
+		    <span about="" id="server-ensure-consistency" rel="spec:requirement" resource="#server-ensure-consistency"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> only accept requests where there is consistency between the path ending (<a href="#uri-slash-semantics">URI Slash Semantics</a>), any HTTP <code>Link</code> headers with <code>rel="type"</code> and the media type given in the <code>Content-Type</code> header of the request.</span></span>
```
csarven, line 669: How will this be tested without knowledge of specific required types? It appears to introduce unnecessary variance.

protocol.html
```diff
@@ -665,6 +665,22 @@ <h3 property="schema:name">Resource Type Heuristics</h3>
...
+		    <span about="" id="server-reject-inconsistency" rel="spec:requirement" resource="#server-reject-inconsistency"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> reject inconsistent requests using a <code>415</code> (for media type inconsistency) or <code>409</code> (for all other inconsistencies) status code as per Section 4.3.4 of [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>].</span></span>
```
csarven, line 670: Isn't this repeating RFC 7231? It is already covered in #http-server.

protocol.html
```diff
@@ -665,6 +665,22 @@ <h3 property="schema:name">Resource Type Heuristics</h3>
...
+		    <span about="" id="server-existing-type-inconsistency" rel="spec:requirement" resource="#server-existing-type-inconsistency"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="spec:Server">server</span> receives a request that attempts to change the resource type, it <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> verify that the resource type of the request is compatible with the current resource type and reject the requests using a <code>409</code> status code if it is not.</span></span>
```
csarven, line 671: For this to be effective, the resource types need to be a normative requirement.

csarven (changes requested):
I like the idea of introducing information on consistency. However, I do not think that we have (significant) normative requirements in SP to specify this just yet. We should get those up to speed (in either direction) to know what consistency will be, especially with the role of `type` link relation. Details: https://github.com/solid/specification/issues/301#issuecomment-970150949

protocol.html
kjetilk: To some extent, it is, but there is considerable room for interpretation since HTTP isn't entirely consistent itself at this point, as the discussion between @RubenVerborgh and myself shows. Therefore, it needs to be explicit.

protocol.html
kjetilk: I did not understand how it introduces variance, to the contrary, it is a constraint... 

It is clear that it cannot be tested in the general case, there's an open world out there, but we do have knowledge of some specific resource types, the LDP ones, and we can test them.

protocol.html
kjetilk: I disagree, will discuss in #301.

protocol.html
csarven: It is introducing a (vague) requirement that depends on variable resource typing. Not only there is no useful testing for the general case but the specific testing is very limited. As mentioned, there is currently only one place where LDP's specific type is used - and that's being blown out of proportion as it seems.

protocol.html
csarven: https://github.com/solid/specification/issues/301#issuecomment-970334856

protocol.html
kjetilk: I assume that you have inspected the existing tests?

csarven: Can you refer to the tests corresponding to the requirements in specs?

kjetilk: Not at present, but if this was merged, I could.

TallTed:
```suggestion
		    <span about="" id="server-ensure-consistency" rel="spec:requirement" resource="#server-ensure-consistency"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> only accept requests where there is consistency between the path ending (<a href="#uri-slash-semantics">URI Slash Semantics</a>), any HTTP <code>Link</code> headers with <code>rel="type"</code>, and the media type given in the <code>Content-Type</code> header of the request.</span></span>
```

Issue #332, opened 2021-10-27
Alternative update support with N3 patch


RubenVerborgh:
## Context
The Solid protocol specifies a mechanism to update resources via HTTP `PATCH` (https://github.com/solid/specification/issues/85) whereby the contents of the change are expressed in the HTTP body.

## Description of current status and problems

The currently proposed media type for these updates is that of the [SPARQL 1.1 Update](https://www.w3.org/TR/sparql11-update/) specification. However, Solid servers deviate from this specification in important ways:
- Only a subset of the SPARQL 1.1 Update language is supported:
  - https://github.com/solid/specification/issues/125
  - https://github.com/solid/specification/pull/320
- There is a desire to have different matching semantics:
  - https://github.com/solid/specification/issues/322
  - https://github.com/solid/solid-spec/pull/193
  - https://github.com/solid-archive/query-panel/issues/3

A currently proposal suggests to use a different (but SPARQL-inspired) syntax altogether in order to avoid confusion with the actual SPARQL 1.1 Update language:
  - https://github.com/solid/specification/pull/330

Whereas I understand the reasoning behind that, I see a couple of problems:
- There are no existing server or client implementations
  - All servers would have to be changed
  - All clients would have to be changed
- The proposed syntax is non-standard
  - This is concerning from a Solid philosophy where we strive to use standards
  - It also means that implementations cannot reuse existing stacks, but have to roll their own

## Proposed solution
My proposal is an N3 patch format as a solution for the 0.9 release.

### Advantages
- Support has been implemented and tested in NSS since 2017
- RDFlib.js code exists for it
  - So JavaScript clients can be migrated easily with an RDFlib.js upgrade
- It is a largely standards-based approach
  - Off-the-shelf parsers already exist for multiple programming languages
- We are in control of the semantics
  - And the semantics are extensible because it is an RDF format

### Disadvantages
- N3 is not an actual W3C standard
- We still require changes to clients and servers

### Examples
A working example patch looks as follows (MIME type `text/n3`):

```turtle
@prefix solid: <http://www.w3.org/ns/solid/terms#>

<> solid:patches <https://tim.localhost:7777/read-write.ttl>;
   solid:where   { ?a <y> <z>. };
   solid:inserts { ?a <y> <z>. };
   solid:deletes { ?a <b> <c>. }.
```

Thee above terms are all defined in https://www.w3.org/ns/solid/terms.

[...]

Issue #244, opened 2021-02-25
Use inclusive language in Solid technical reports


csarven:
Inspired by https://github.com/w3c/idcg/issues/17

csarven:
Why would anyone need to be knowledgeable about science fiction? I think Guinan and Darmok are considerably better than Alice/Bob/Carl.. which happen to be very English-centric, and boring if I may add.

RubenVerborgh:
I will just describe the effect it had on me: I don't know anything about science fiction, and when reading the specs, I came across names that I did not know or understand, and they felt they are an in-joke or reference that I was not part of. So I Google them, and see that it is a Star Trek reference, which I did not get, so I felt alienated/excluded.

That said, I am okay with that feeling for myself, because I have other ways of feeling included within Solid. So no need to change it for me.

But to people with a different background than myself, the alienation coming from references to something outside of their frame of reference might have a stronger reaction.

> Alice/Bob/Carl.. which happen to be very English-centric

How about https://forebears.io/earth/forenames?

jeff-zucker:
I agree with @RubenVerborgh on this one.   Insider jokes that give people the feeling "everyone gets this but me" are not inclusive.  Names that emphasize ethnic/cultural diversity seem more appropriate here.

TallTed:
I don't care if the chosen names are boring, as the point is not the names nor the non-people at all, but the activities in which they are supposed to be participating.  The only reason for using names at all is to make the usage scenario easier to think about.

No matter what language(s) we choose, someone will feel alienation or disinclusion, because we can't use every language at once.  Given that the docs are in English, it seems to me to make some sense for the placeholding names to be English-centric, as names from any other language/ethnicity/culture will be jarring from the textual flow to some degree.

Maybe, since every group that works on Use Cases & Requirements (among other) docs will run into this sooner or later, we should consider pushing for a cross-WG/CG/IG/XG/*G Task Force to build such an inclusive name set that everyone can use going forward, instead of all of us reinventing this wheel on a regular basis?

csarven:
Happy to celebrate people/names from underrepresented etnical/cultural backgrounds.

I disagree with the example of a techie white guy not getting in on minor fun while working at Google... Using Guinan as example was to celebrate a wise black women. Yes, a scifi character. I picked one. The intention wasn't to exclude people but point taken if that offends some. We can improve. I created this issue and used those examples to change the frame .. hopefully for the better.

csarven:
[...]

Issue #220, opened 2021-01-17
Is Read required on top of Write for PATCH delete and where ?


bourgeoa:
On PATCH spec seems to require only `Write` for sparql DELETE. This is implemented in solid/test-suite.
But NSS is requiring `Read and Write`, with the following explanations in the code comments in NSS PATCH implementation.

```
// Read access is required for DELETE and WHERE.
// If we would allows users without read access,
// they could use DELETE or WHERE to trigger 200 or 409,
// and thereby guess the existence of certain triples.
// DELETE additionally requires write access.
```
What should be the spec ? What should test-suite implement if the response is not straight forward ?

RubenVerborgh:
I've come up with the above explanation together with @timbl, and stand by it.

csarven:
I'm not sure which spec you're referring to but will take it as underspecified/unknown at the very least in the issues here and current Protocol spec. But right that the current web-access-control-spec alludes to `acl:Write` allowing PATCH.. and the wiki WebAccessControl doesn't require Read.

---

When using WAC, I agree on requiring `acl:Read` and `acl:Write` in order to accept PATCH operations including `DELETE DATA` or `WHERE` - at least those with application/sparql-update data type if anything.

When agent doesn't have `acl:Read` and `acl:Write` on target, rejection should use the 403 status code.

When agent has `acl:Read` and `acl:Write` on target and there is *no match* on `DELETE DATA`, rejection should use the 409 status code.

I've documented some PATCH requests in a table based on `INSERT DATA` and `DELETE DATA` in https://github.com/solid/specification/issues/14#issuecomment-683480525 as those appear to be the basic level of SPARQL Update we seem to agree on to date. Related issue: https://github.com/solid/specification/issues/125 . So, `WHERE` support needs an explicit agreement.

See also for recent-ish description of the current state of things: https://github.com/solid/community-server/issues/15#issuecomment-668615622

RubenVerborgh:
Note that this also relates to #139. If SPARQL UPDATE does _not_ act as a semaphore mechanism (so if it _does_ follow the SPARQL 1.1 spec), then we might not be leaking any information.

So perhaps WebAccessControl was initially assuming that semantics; and its only when the semaphore semantics were introduced that the leakage was created.

RubenVerborgh:
@acoburn What does ESS implement?

acoburn:
With ESS, a PATCH operation does not _require_ READ access. 

[...]

Issue #57, opened 2019-09-20
Does the acl:Authorization need to be materialized in an ACL resource


acoburn:
Via simple RDFS entailment, one can infer the presence of `acl:Authorization` in a valid ACL document:

```turtle
acl:mode rdfs:domain acl:Authorization .
acl:agent rdfs:domain acl:Authorization .
acl:accessTo rdfs:domain acl:Authorization .
```

And from the implementation of [ACL-Check](https://github.com/solid/acl-check/blob/master/test/unit/permission-set-test.js#L288-L301) it appears that the materialized presence of `<...> a acl:Authorization` is not necessary. It would be helpful if the WebAC specification clarified whether an implementation really needs to materialize that triple.

RubenVerborgh:
I would say it does not; and we probably want to spec a minimal shape (cfr. #56).

pmcb55:
I think I must be missing something here, but basically I don't like **_relying_** on entailment (even simple RDFS entailment), and in fact I also have concerns about even using `rdfs:range` and `rdfs:domain` (I'm sure they might be fine here in the ACL vocab, but for example, Schema.org doesn't use them (instead using the much looser `schema:rangeIncludes` and `schema:domainIncludes`), and professor Robert Stevens from Manchester University recommended against overusing `range` and `domain` in an OWL course I attended years ago!)).
So I'd just ask why would the spec **not** mandate materializing the `<...> a acl:Authorization` triple? Are we just trying to save a triple? Personally I prefer being as explicit (and readable) as possible.

RubenVerborgh:
> I don't like **_relying_** on entailment

We should not.
It's just that clients can be designed for this specific case such that they don't _need_ the type triple to be present.

pmcb55:
Sorry @RubenVerborgh, but I don't follow what you mean. Are you saying it might be good if we make it just a little bit easier for clients if they don't _need_ to provide this simple type triple (or parse it from server responses)? I really don't like that idea (since behaviour is then implicit, or is relying on inference), so I'm guessing you mean something else...

RubenVerborgh:
Consumers of an ACL should not need the type triple to function properly. They don’t even need to infer it. They should just look for the relevant predicates.

pmcb55:
[...]

PR #346, opened 2021-11-10
Describe N3 Patch

RubenVerborgh:
Closes https://github.com/solid/specification/issues/332


protocol.html
```diff
@@ -717,6 +717,31 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                      <li>A patch resource contains at most one triple of the form <code>?patch solid:where ?conditions</code>.</li>
+                      <li>When present, <code>?deletions</code>, <code>?insertions</code>, and <code>?conditions</code> must be non-nested <em>cited formulae</em> [<cite><a class="bibref" href="#bib-n3">N3</a></cite>] consisting only of triples and/or triple patterns. When not present, they are presumed to be the empty formula <code>{}</code>.</li>
+                    </ul>
+                    <p>While other specifications might provide a structure and interpretation for a wider class of N3 Patch documents, the present specification only governs the application of N3 Patch documents that additionally adhere to the following structural constraints:</p>
+                    <ul>
+                      <li>The patch document contains exactly one patch resource.</li>
+                      <li>The <code>?target</code> is an exact match with the request URI.</li>
+                      <li>The patch resource may optionally contain one triple of the form <code>?patch solid:matchingStrategy solid:SingleMatch</code>.</li>
```
RubenVerborgh, line 737: This is new behavior; proposing this for future compatibility. It makes the semaphore behavior explicit.

RubenVerborgh (commented):
This is the initially proposed text; if approved, I can also add an example.

protocol.html
```diff
@@ -717,6 +717,37 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                      <li>A patch resource contains at most one triple of the form <code>?patch solid:inserts ?insertions</code>.</li>
+                      <li>A patch resource contains at most one triple of the form <code>?patch solid:where ?conditions</code>.</li>
+                      <li>When present, <code>?deletions</code>, <code>?insertions</code>, and <code>?conditions</code> must be non-nested <em>cited formulae</em> [<cite><a class="bibref" href="#bib-n3">N3</a></cite>] consisting only of triples and/or triple patterns. When not present, they are presumed to be the empty formula <code>{}</code>.</li>
+                    </ul>
+
+                    <p>While other specifications might provide a structure and interpretation for a wider class of N3 Patch documents, the present specification only governs the application of N3 Patch documents that additionally adhere to the following structural constraints:</p>
+
+                    <ul>
+                      <li>The patch document contains exactly one patch resource.</li>
+                      <li>The <code>?target</code> is an exact match with the request URI.</li>
```
kjetilk, line 740: To me, this indicates that it is actually a superfluous triple, you shouldn't have to specify the target document, since it is always known as the request URI. Moreover, we should not allow using `PATCH` to modify other resources than the target specified by the request URI, since it may end in exploits abusing improper input sanitation. It could also be a practical issue, as it means that you would have to modify the patch document if you want to apply it to many resources individually. Could we leave this out?

csarven: May also need to consider possible redirects into account.

protocol.html
```diff
@@ -717,6 +717,37 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                <section id="n3-patch" inlist="" rel="schema:hasPart" resource="#n3-patch">
+                  <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
+                  <div datatype="rdf:HTML" property="schema:description">
+                    <p><span about="" id="server-patch-n3" rel="spec:requirement" resource="#server-patch-n3"><span property="spec:statement">When the target resource of a <code>PATCH</code> request is an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>], <span rel="spec:requirementSubject" resource="spec:Server">servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support modifying that resource using a request body containing an <em>N3 Patch</em>.</span></span> [<a href="https://github.com/solid/specification/issues/125#issuecomment-959518598" rel="cito:citesAsSourceDocument">Source</a>]</p>
+
+                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-n3">N3</a></cite>], identified by the MIME type <code>text/n3</code>, conforming to the following structural constraints:</p>
```
kjetilk, line 726: The anchor doesn't point right, #bib-notation3 was used below.

protocol.html
RubenVerborgh: I understand why it seems superfluous; but:
- This is the one identifying triple for the patch (it's the triple patches will always have)
  - Counter-argument: we can have another way
- It allows the patch to be understood in isolation
  - Counter-argument: it indeed cannot be reused verbatim for multiple resources
- It allows future extensibility towards supporting (atomic?) patches across multiple documents
  - Counter-argument: then it's up to the future extension point to define `solid:target`

> we should not allow using PATCH to modify other resources than the target specified by the request URI

Agree for 0.9, hence the constraint. Just leaving options open for 1.x.

> Could we leave this out?

We could, but:
- We would need to change implementations (whereas 0.9 is about documenting existing systems).
  - However, we could still optionally allow it.
- We would need another way of identifying the main patch object.
  - "Any resource to which `solid:inserts`, `solid:deletes`, `solid:where` is attached"?

protocol.html
RubenVerborgh: Thanks, fixed.

protocol.html
RubenVerborgh: Will drop `?target` in a separate commit for now.

RubenVerborgh: Added an example in b043d364694efe0ab95fc0bb00198149bd8716a8; looks a bit "naked" now without the leading `solid:patches` triple IMHO, hard to predict what's going on. I wonder if we (optionally) want to allow an `rdf:type solid:Patch` triple instead.

RubenVerborgh:
@csarven @justinwb @kjetilk @timbl Ready for review.

protocol.html
```diff
@@ -717,6 +719,51 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                  <div datatype="rdf:HTML" property="schema:description">
+                    <p><span about="" id="server-patch-n3" rel="spec:requirement" resource="#server-patch-n3"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support a <code>PATCH</code> request with an <em>N3 Patch</em> body when its target is an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>], and indicate such support by listing <code>text/n3</code> as a value of the <code>Accept-Patch</code> header [<cite><a class="bibref" href="#bib-rfc5789">RFC5789</a></cite>] of relevant responses.</span></span> [<a href="https://github.com/solid/specification/issues/125#issuecomment-959518598" rel="cito:citesAsSourceDocument">Source</a>]</p>
+
+                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>], identified by the MIME type <code>text/n3</code>, conforming to the following structural constraints:</p>
```
TallTed, line 728:
```suggestion
                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>], identified by the media type (formerly known as MIME type), <code>text/n3</code>, conforming to the following structural constraints:</p>
```

RubenVerborgh: Thanks @TallTed, edited in bad5f15b5f05e8ef98094e41a18672c905c5c45c.

protocol.html
```diff
@@ -717,6 +719,51 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+
+                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>], identified by the media type <code>text/n3</code>, conforming to the following structural constraints:</p>
+
+                    <ul>
+                      <li>The patch document can contain patch resources identified by a URI <code>?patch</code> which can be chosen freely.</li>
```
kjetilk, line 731: Should we specify that it should be identified by a URI? It does seem to me that this URI doesn't really have that much value, so it might be reasonable to use a blank node for it?

protocol.html
```diff
@@ -717,6 +719,51 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                      <li>A patch resource may contain a triple <code>?patch rdf:type solid:Patch</code>.</li>
+                      <li>A patch resource contains at most one triple of the form <code>?patch solid:deletes ?deletions</code>.</li>
+                      <li>A patch resource contains at most one triple of the form <code>?patch solid:inserts ?insertions</code>.</li>
+                      <li>A patch resource contains at most one triple of the form <code>?patch solid:where ?conditions</code>.</li>
+                      <li>When present, <code>?deletions</code>, <code>?insertions</code>, and <code>?conditions</code> must be non-nested <em>cited formulae</em> [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>] consisting only of triples and/or triple patterns. When not present, they are presumed to be the empty formula <code>{}</code>.</li>
```
kjetilk, line 736: In general, I think the language should have stronger imperatives here, and also use the capitalization for MAY, MUST, etc, that we have in the rest of the document. So, perhaps e.g. "A patch resource MUST NOT contain more than..."

protocol.html
```diff
@@ -717,6 +719,51 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                    </ul>
+
+                    <p>While other specifications might provide a structure and interpretation for a wider class of N3 Patch documents, the present specification only governs the application of N3 Patch documents that additionally adhere to the following structural constraints:</p>
+
+                    <ul>
+                      <li>The patch document contains exactly one patch resource, identified by one or more of the triple patterns described above, which should all share the same <code>?patch</code> subject.</li>
+                      <li>The patch resource may optionally contain one triple of the form <code>?patch solid:matchingStrategy solid:SingleMatch</code>.</li>
+                      <li>The <code>?insertions</code> and <code>?deletions</code> formulae must not contain variables that do not occur in the <code>?conditions</code> formula.</li>
+                      <li>The <code>?insertions</code> and <code>?deletions</code> formulae must not contain blank nodes.</li>
+                      <li>The patch document does not contain any other triples.</li>
```
kjetilk, line 746: Some of the same here, cap the compliance keywords.

kjetilk, line 746: Is the final requirement needed? I would think we might as well ignore any other triples in there? I guess some might add an `rdfs:comment` to describe what the patch is doing to devs, and that should be OK, I guess?

protocol.html
```diff
@@ -717,6 +719,51 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                  </div>
+
+                  <figure id="n3-patch-example" class="example listing" rel="schema:hasPart" resource="#n3-patch-example">
+                    <p class="example-h"><span>Example</span>: Applying an N3 patch.</p>
+                    <pre about="#n3-patch-example" property="schema:description" typeof="fabio:Script">
+<code>@prefix solid: &lt;http://www.w3.org/ns/solid/terms#&gt;.
+@prefix ex: &lt;http://www.example.org/terms#&gt;.
+
+&lt;&gt; a solid:Patch;
```
kjetilk, line 760: Not sure that `<>` is a good idea. That means "this resource", and that would resolve to the target resource, right? So, that would result in a URI collision between the target resource and the patch document. I think a blank node is actually a reasonable thing to have there...

kjetilk (changes requested):
I think this is going in a direction that will be workable very soon. My comments are mostly editorial, but I think needed to get into line with the language of the rest of the spec.

kjetilk (changes requested):
One more thing, it should mention what kind of operations these are (i.e. append, read, write), so that authz systems can hook into that, also in line with #220 . I guess you can adopt some of the language I used for this #320.

protocol.html
justinwb, line 736: This section introduces the first occurrence of "triple pattern" in the protocol document. I think it would be useful to provide a reference link, or include it as a definition (might also be worth adding a definition for triple while we're at it).

protocol.html
```diff
@@ -717,6 +719,51 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
                 </div>
+
+                <section id="n3-patch" inlist="" rel="schema:hasPart" resource="#n3-patch">
+                  <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
+                  <div datatype="rdf:HTML" property="schema:description">
+                    <p><span about="" id="server-patch-n3" rel="spec:requirement" resource="#server-patch-n3"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support a <code>PATCH</code> request with an <em>N3 Patch</em> body when its target is an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>], and indicate such support by listing <code>text/n3</code> as a value of the <code>Accept-Patch</code> header [<cite><a class="bibref" href="#bib-rfc5789">RFC5789</a></cite>] of relevant responses.</span></span> [<a href="https://github.com/solid/specification/issues/125#issuecomment-959518598" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
justinwb, line 726:
```suggestion
                    <p><span about="" id="server-patch-n3" rel="spec:requirement" resource="#server-patch-n3"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support a <code>PATCH</code> request with an <em>N3 Patch</em> body when the target of the request is an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>], and indicate such support by listing <code>text/n3</code> as a value of the <code>Accept-Patch</code> header [<cite><a class="bibref" href="#bib-rfc5789">RFC5789</a></cite>] of relevant responses.</span></span> [<a href="https://github.com/solid/specification/issues/125#issuecomment-959518598" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

protocol.html
```diff
@@ -717,6 +719,51 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
...s, creating an empty RDF dataset if the target resource does not exist yet. If <code>?conditions</code> is non-empty, then the dataset must contain exactly one match of its contents after variable substitution; if not, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code. The resulting (possibly empty) variable mapping must be propagated to the <code>?deletions</code> and <code>?insertions</code> formulae. Then, the resulting triples in <code>?deletions</code> must be removed from the RDF dataset if there is exactly one match; if not, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code. Finally, the resulting triples in <code>?insertions</code> must be added to the RDF dataset, where any blank nodes must result in newly created blank nodes. The combination of the deletions and insertions then forms the new resource state of the RDF document, and the server responds with the appropriate status code.</span></span></p>
```
justinwb, line 751: This might benefit from being formatted as an ordered list after "as follows"

justinwb (commented):
Overall looks good - just a few comments

protocol.html
TallTed: Unless I'm misunderstanding badly, I believe `<>` here resolves to the patch document.

protocol.html
RubenVerborgh: Done in cb80918902b2e546f0313766db083f5919fb9520

protocol.html
RubenVerborgh: It is needed for future compatibility. Suppose that other clauses like `solid:filter` are added, or that `solid:matchingStrategy solid:MultipleMatch` is used. So the server needs to reject what it does not understand.

protocol.html
RubenVerborgh: We assumed indeed that `<>` is the patch document; when @timbl and myself designed this, we wanted the thing to describe itself (hence also the `solid:patches`). So we wanted to convey "I am a patch that inserts/deletes …".
But now that a patch document can contain multiple patches, we probably want a different identifier altogether. Replaced in cb80918902b2e546f0313766db083f5919fb9520.

protocol.html
RubenVerborgh: True, in 1d336ff11f56b9689c1a325114a43e94702c38e1

protocol.html
RubenVerborgh: In 1d336ff11f56b9689c1a325114a43e94702c38e1

protocol.html
RubenVerborgh: Thanks, in c4613d24da2038c40b5c35f36a4382e9412c1962

protocol.html
RubenVerborgh: @csarven What is the best place for this?

protocol.html
RubenVerborgh: I'm cautious with using explicitly numbered steps, as this gets close to the description of an imperative algorithm, which has caused trouble with specs in the past. Deferring to @csarven?

protocol.html
RubenVerborgh: (I had hoped the Notation3 spec would define this, but it does not, because of quantification. SPARQL does, but it might not be entirely correct to cross-reference that here.)

protocol.html
kjetilk: Right, OK!

protocol.html
TallTed: I'm confused by there apparently being one action after "as follows:" (the list is ended by the next period; if there are meant to be multiple things that "follow", that period [and some subsequent] should instead be a semicolon).

If the (must be multiple) steps are meant to be algorithmic (though not imperatively so), an ordered list does make sense.  The implied imperative can be can be relieved by simply changing "as follows" to "roughly as follows".

If it's not so much an ordered route to conclusion, as a list of disjoint states of which only one will match the situation in play, then an unordered list would be better.

Either ordered or unordered list would be clearer than the streaming paragraph.

protocol.html
```diff
@@ -723,27 +723,27 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                      <li id="server-patch-n3-identifier" rel="spec:requirement" resource="#server-patch-n3-deletes"><span property="spec:statement">A patch resource <span rel="spec:requirementLevel" resource="spec:MAY">MAY</span> contain at most one triple of the form <code>?patch solid:deletes ?deletions</code>.</span></li>
+                      <li id="server-patch-n3-inserts" rel="spec:requirement" resource="#server-patch-n3-inserts">A patch resource <span rel="spec:requirementLevel" resource="spec:MAY">MAY</span> contain at most one triple of the form <code>?patch solid:inserts ?insertions</code>.</li>
+                      <li id="server-patch-n3-where" rel="spec:requirement" resource="#server-patch-n3-where"><span property="spec:statement">A patch resource <span rel="spec:requirementLevel" resource="spec:MAY">MAY</span> contain at most one triple of the form <code>?patch solid:where ?conditions</code>.</span></li>
```
TallTed, line 735: RFC `MAY` does not have the same meaning as appears to be intended by these statements.  `MAY contain at most` does not restrict the resource being described at all.

protocol.html
RubenVerborgh: Done in f9893d4156756f1d2a1388decb18bf4be527ad46

protocol.html
RubenVerborgh: Thanks, clarified in 03d8df7e6d224fd7d3dfaa2e857c9c8bcc15a119

RubenVerborgh:
> One more thing, it should mention what kind of operations these are (i.e. append, read, write), so that authz systems can hook into that, also in line with #220 . I guess you can adopt some of the language I used for this #320.

In 76398fe029d936f025ad9f90c08043e1166bc474

protocol.html
RubenVerborgh: I'll need too leave this one open now, as "triple" is also used but not referenced in the document at the moment, so that's a dependency.

protocol.html
```diff
@@ -748,7 +748,15 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
+                    <p><span about="" id="server-patch-n3-semantics" rel="spec:requirement" resource="#server-patch-n3-semantics"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> process a patch resource against the target document as follows:</span></span></p>
+                    <ol>
+                      <li>Start from the RDF dataset in the target document, or an empty RDF dataset if the target resources does not exist yet.</li>
+                      <li>If <code>?conditions</code> is non-empty, then the dataset must contain exactly one match of its contents after variable substitution; if not, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.</li>
```
TallTed, line 756:
```suggestion
                      <li>If <code>?conditions</code> is non-empty, and the dataset contains exactly one match of its contents after variable substitution, proceed to the next step; if not, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.</li>
```

protocol.html
```diff
@@ -748,7 +748,15 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
+                      <li>The resulting (possibly empty) variable mapping is propagated to the <code>?deletions</code> and <code>?insertions</code> formulae.</li>
+                      <li>Then, the resulting triples in <code>?deletions</code> are removed from the RDF dataset if there is exactly one match; if not, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.</li>
```
TallTed, line 758: I'm struggling with this one. Is the `409` required for the whole patch if there is not exactly one match for any triple in `?deletions`; or is there to be a `409` for *each* triple that has not exactly one match; or something else? The phrasing I'm suggesting here is probably not correct, but I hope it's closer than the phrasing I'm suggesting be changed...
```suggestion
                      <li>Then, for each resulting triple in <code>?deletions</code> which has exactly one match in the RDF dataset, that match is removed; if there is not exactly one match in the RDF dataset, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.</li>
```

RubenVerborgh: Indeed; clarified in 99fcc3219789169e459d7078d42df5e2af9eb64c.

protocol.html
RubenVerborgh: Clarified in 99fcc3219789169e459d7078d42df5e2af9eb64c.

protocol.html
RubenVerborgh: Following up in https://github.com/solid/specification/issues/348

protocol.html
```diff
@@ -753,11 +753,12 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
+                      <li>The resulting variable mapping is propagated to the <code>?deletions</code> and <code>?insertions</code> formulae to obtain two sets of resulting triples.</li>
+                      <li>If the set of triples resulting from <code>?deletions</code> is non-empty and the dataset does not contain <em>all</em> of these triples, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.
+                      <li>The triples resulting from <code>?deletions</code> are to be removed from the RDF dataset.</li>
+                      <li>The triples resulting from <code>?insertions</code> are to be added to the RDF dataset, where any blank nodes result in newly created blank nodes.</li>
```
TallTed, line 760:
```suggestion
                      <li>The triples resulting from <code>?insertions</code> are to be added to the RDF dataset, with each blank node from <code>?insertions</code> resulting in a newly created blank node.</li>
```

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-other-triples" rel="spec:requirement" resource="#server-patch-n3-identifier"><span property="spec:statement">The patch document <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain any other triples.</span></li>
+                    </ul>
+
+                    <p><span about="" id="server-patch-n3-invalid" rel="spec:requirement" resource="#server-patch-n3-invalid"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>400</code> status code if a patch document does not satisfy all of the above constraints.</span></span></p>
```
kjetilk, line 749: May I suggest [`422`](https://datatracker.ietf.org/doc/html/rfc4918#section-11.2) for this purpose, as in this case, the patch is syntactically valid, but not semantically valid?

kjetilk (approved):
My main concerns have been adequately addressed, so I think this can go in now. :+1: 

I still think `422` for semantic errors is advisable, but we could also add that later.

justinwb (approved):
Happy with adjustments based on feedback ✅

protocol.html
RubenVerborgh: Oh absolutely.

RubenVerborgh: …although this is a WebDAV extension (https://datatracker.ietf.org/doc/html/rfc4918#section-11.2) and not part of the HTTP spec (https://datatracker.ietf.org/doc/html/rfc7231). Not sure if/how that matters.

kjetilk (approved):
Yup, which is basically an RDF Merge operation from the RDF MT spec, but yeah, I guess it could be said that simply :-)

RubenVerborgh:
All comments resolved; handover to @csarven.

kjetilk (approved):
Yay! Now definitely GTG!

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+
+                    <p><span about="" id="server-patch-n3-semantics" rel="spec:requirement" resource="#server-patch-n3-semantics"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> process a patch resource against the target document as follows:</span></span></p>
+                    <ol>
+                      <li>Start from the RDF dataset in the target document, or an empty RDF dataset if the target resources does not exist yet.</li>
+                      <li>If <code>?conditions</code> is non-empty, find all (possibly empty) variable mappings such that all of the resulting triples occur in the dataset. If no such mapping exists, or if multiple mappings exist, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>409</code> status code.
```
csarven, line 756: I'm curious to understand why multiple mappings results in 409. (I'm not challenging the idea.) I would've assumed that multiple mappings will be removed.

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                    <p class="example-h"><span>Example</span>: Applying an N3 patch.</p>
+                    <pre about="#n3-patch-example" property="schema:description" typeof="fabio:Script">
+<code>@prefix solid: &lt;http://www.w3.org/ns/solid/terms#&gt;.
+@prefix ex: &lt;http://www.example.org/terms#&gt;.
+
+_:rename a solid:Patch;
+   solid:where   { ?person ex:familyName "Anchor". };
+   solid:inserts { ?person ex:givenName "An". };
+   solid:deletes { ?person ex:givenName "Bob". }.</code></pre>
+                    <figcaption property="schema:name">This N3 Patch instructs to rename <em>An Anchor</em> into <em>Bob Anchor</em>, on the condition that no other Anchor family members are present in the target RDF document.</figcaption>
```
csarven, line 775: I admit that it took me a couple of eyeballing to get "An". Can we use Kathryn Janeway instead or whatever you prefer? (Related: https://github.com/solid/specification/issues/244 )

If I'm not mistaken, the intention is to rename "Bob Anchor" to "An Anchor":

```suggestion
                    <figcaption property="schema:name">This N3 Patch instructs to rename <em>Bob Anchor</em> into <em>An Anchor</em>, on the condition that no other Anchor family members are present in the target RDF document.</figcaption>
```

csarven (changes requested):
Mostly good to go. Just some considerations. Requesting your feedback/change. I'll go ahead with the markup stuff (which I didn't spend time on but generally seemed fine to me) after your response.

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+
+                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>], identified by the media type <code>text/n3</code>, conforming to the following structural constraints:</p>
+
+                    <ul>
+                      <li id="server-patch-n3-identifier" rel="spec:requirement" resource="#server-patch-n3-identifier"><span property="spec:statement">The patch document <span rel="spec:requirementLevel" resource="spec:MAY">MAY</span> contain patch resources identified by a URI or blank node <code>?patch</code> which can be chosen freely.</span></li>
+                      <li id="server-patch-n3-type" rel="spec:requirement" resource="#server-patch-n3-identifier">A patch resource <span rel="spec:requirementLevel" resource="spec:MAY">MAY</span> contain a triple [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>] <code><span property="spec:statement">?patch rdf:type solid:Patch</span></code>.</li>
```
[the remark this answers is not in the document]
RubenVerborgh: Do you have a pointer to those discussions? I have seen the conclusion (and we have implemented it as such), but I didn't agree at first sight. So would be useful to understand where that came from.

Note that `solid:Patch` was not mandatory in the past (whereas I believe `acl:Authorization` was).

protocol.html
RubenVerborgh: https://github.com/solid/specification/issues/322
https://github.com/solid/solid-spec/pull/193
https://github.com/solid-archive/query-panel/issues/3

Will add source.

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-other-triples" rel="spec:requirement" resource="#server-patch-n3-identifier"><span property="spec:statement">The patch document <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain any other triples.</span></li>
+                    </ul>
+
+                    <p><span about="" id="server-patch-n3-invalid" rel="spec:requirement" resource="#server-patch-n3-invalid"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond with a <code>422</code> status code [<cite><a class="bibref" href="#bib-rfc4918">RFC4918</a></cite>] if a patch document does not satisfy all of the above constraints.</span></span></p>
```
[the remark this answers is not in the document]
RubenVerborgh: I wouldn't mention to avoid duplication / inconsistencies.

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                      <li id="server-patch-n3-variables" rel="spec:requirement" resource="#server-patch-n3-variables"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain variables that do not occur in the <code>?conditions</code> formula.</span></li>
+                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-other-triples" rel="spec:requirement" resource="#server-patch-n3-identifier"><span property="spec:statement">The patch document <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain any other triples.</span></li>
```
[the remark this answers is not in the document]
RubenVerborgh: It is necessary for forward compatibility: what if patch functionality is altered by such statements? Please advise.

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+
+                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>], identified by the media type <code>text/n3</code>, conforming to the following structural constraints:</p>
+
+                    <ul>
+                      <li id="server-patch-n3-identifier" rel="spec:requirement" resource="#server-patch-n3-identifier"><span property="spec:statement">The patch document <span rel="spec:requirementLevel" resource="spec:MAY">MAY</span> contain patch resources identified by a URI or blank node <code>?patch</code> which can be chosen freely.</span></li>
```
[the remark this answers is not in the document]
kjetilk: What's the point with a patch document that doesn't contain patch resources?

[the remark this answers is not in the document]
RubenVerborgh: Future compat; later down, we explain that the current spec only governs cases where there is exactly one.

protocol.html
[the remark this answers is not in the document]
kjetilk: Wasn't that the discusssion in #57 ?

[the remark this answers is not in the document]
RubenVerborgh: It was; but I don't see how we've come to the conclusion there, because there were disagreements above.

protocol.html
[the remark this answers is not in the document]
kjetilk: Sure, but I don't understand the point of the case where there is none...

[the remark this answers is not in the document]
RubenVerborgh: It might be a different kind of patch resource than the one we describe here. But happy to change it into "at least one" when I make the above edit.

protocol.html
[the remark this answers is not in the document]
csarven: I don't have a hard stance on this. Just want to make sure we cover what we really want out of this.

---

Perhaps I'm not understanding how other statements can modify the patch functionality. Isn't that out of spec behaviour? Or do you mean stuff getting added that's unspecified.. and later on a spec comes up and uses a pattern which may or may not conflict with existing N3 Patch documents? (New extensions shouldn't spec stuff that conflicts with existing use) Or do you mean another specification could make it so? My understanding is that only what's specified will be processed and the rest will be ignored. That aside, it is an RDF document.

---

Another way to look at this may from the point of extensions basically saying that new stuff is allowed as long as they don't contradict (fudge up) what's defined in this spec. Perhaps a much shorter version of the following will cover (from the Notifications Protocol / WAC):

>Extensions incorporate additional features beyond what is defined in this specification. Extensions MUST NOT contradict nor cause the non-conformance of functionality defined in this specification.

Not necessarily in terms of "extension" but along those lines. Only look at solid:Patch stuff..

protocol.html
TallTed: Ah.  Might I request the more common name spellings of "Ann" or "Anne" for this use?

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+                  <div datatype="rdf:HTML" property="schema:description">
+                    <p><span about="" id="server-patch-n3" rel="spec:requirement" resource="#server-patch-n3"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support a <code>PATCH</code> request with an <em>N3 Patch</em> body when the target of the request is an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>], and indicate such support by listing <code>text/n3</code> as a value of the <code>Accept-Patch</code> header [<cite><a class="bibref" href="#bib-rfc5789">RFC5789</a></cite>] of relevant responses.</span></span> [<a href="https://github.com/solid/specification/issues/125#issuecomment-959518598" rel="cito:citesAsSourceDocument">Source</a>]</p>
+
+                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>], identified by the media type <code>text/n3</code>, conforming to the following structural constraints:</p>
```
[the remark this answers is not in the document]
RubenVerborgh: I wrote "structural" as opposed to "semantic" constraints as in https://datatracker.ietf.org/doc/html/draft-svensson-profiled-representations-01#section-1.1.
I've removed it in f95ca698273e40d8d157bcd85895fc3974d4ec2f.

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
                 </div>
+
+                <section id="n3-patch" inlist="" rel="schema:hasPart" resource="#n3-patch">
+                  <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
+                  <div datatype="rdf:HTML" property="schema:description">
+                    <p><span about="" id="server-patch-n3" rel="spec:requirement" resource="#server-patch-n3"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support a <code>PATCH</code> request with an <em>N3 Patch</em> body when the target of the request is an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>], and indicate such support by listing <code>text/n3</code> as a value of the <code>Accept-Patch</code> header [<cite><a class="bibref" href="#bib-rfc5789">RFC5789</a></cite>] of relevant responses.</span></span> [<a href="https://github.com/solid/specification/issues/125#issuecomment-959518598" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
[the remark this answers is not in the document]
RubenVerborgh: Done in 183b03b46380503a4d6dcae3236a636a4e3ff5f4.

protocol.html
```diff
@@ -717,6 +719,62 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
...ec:requirement" resource="#server-n3-patch-where"><span property="spec:statement">When <code>?conditions</code> is non-empty, <span rel="spec:requirementSubject" resource="spec:Server">servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> treat the request as a <a href="#read-operation">Read operation</a>.</span></span> <span about="" id="server-n3-patch-insert" rel="spec:requirement" resource="#server-n3-patch-insert"><span property="spec:statement">When <code>?insertions</code> is non-empty, <span rel="spec:requirementSubject" resource="spec:Server">servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> treat the request as an <a href="#append-operation">Append operation</a>.</span></span> <span about="" id="server-n3-patch-delete" rel="spec:requirement" resource="#server-n3-patch-delete"><span property="spec:statement">When <code>?deletions</code> is non-empty, <span rel="spec:requirementSubject" resource="spec:Server">servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> treat the request as a <a href="#read-operation">Read</a> and <a href="#write-operation">Write operation</a>.</span></span></p>
```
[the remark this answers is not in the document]
RubenVerborgh: Tried to do this with dcb2fce13a467af2f15d8b7f196b2d832665fb77; anything more seemed to make it more complex instead of more clear.

protocol.html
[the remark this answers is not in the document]
RubenVerborgh: > Or do you mean stuff getting added that's unspecified.. and later on a spec comes up

Exactly; see https://github.com/solid/specification/pull/346#discussion_r758183032

> (New extensions shouldn't spec stuff that conflicts with existing use)

Agreed.

> Or do you mean another specification could make it so?

Yeah.

Example:

```
_:rename a solid:Patch;
   ex:condition { ?person ex:startWith "http://example.org" };
   solid:where   { ?person ex:familyName "Anchor". };
   solid:inserts { ?person ex:givenName "An". };
   solid:deletes { ?person ex:givenName "Bob". }.
```

where `ex:condition` adds a stronger constraint.

> Another way to look at this may from the point of extensions basically saying that new stuff is allowed as long as they don't contradict (fudge up) what's defined in this spec.

It does not, because of https://github.com/solid/specification/pull/346#discussion_r758183032, where we say that this spec only governs very specify patch documents.

Happy to consider other solutions; for instance, mandating another `rdf:type` in such cases.

protocol.html
RubenVerborgh: Done in 34020da6df2600c33996a71209aaec0c323de16e.

protocol.html
RubenVerborgh: Done in fef2ac79e99f497c769df416aa7be968e254c349 (meant to use traditional Alice/Bob example).

RubenVerborgh:
@csarven Thanks, I have resolved all comments except these that need input:
- https://github.com/solid/specification/pull/346#discussion_r758154409
- https://github.com/solid/specification/pull/346#discussion_r758188977

protocol.html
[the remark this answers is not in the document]
kjetilk: I don't really have a strong opinion on this, but in the interest of progress, will just say that the key to the requirement with `acl:Authorization` is extensibility. That is then balanced against that it takes longer to look up an authorization with the additional triple.

Conceivably, we could have more detailed patch classes in the ontology, so the same could apply there, but OTOH, it seems quite unlikely. My feeling is that the difference part of sameSameButDifferent should prevail, and that a MAY is OK.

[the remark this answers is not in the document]
RubenVerborgh: > will just say that the key to the requirement with acl:Authorization is extensibility

I don't understand how exactly, but happy to drop the discussion for now as to not get distracted.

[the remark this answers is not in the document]
csarven: solid:Patch can be subclassed and used to indicate the specialisations such as #server-patch-n3-default or others in the future. I think this is similar to authorizations eg. time-based https://github.com/solid/specification/issues/56#issuecomment-872020854

I think the designs should align here. That aside, as Kjetil, "I don't have a strong open on this". It is worthwhile to record the decision here however it may be.

protocol.html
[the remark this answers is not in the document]
csarven: They're essentially the same. It entails that extensions can be subclassed.

Can no other statement be part of this document e.g. `_:rename rdfs:label "...".`, `<> dcterms:creator <...>.` ?

csarven (approved):
It'd be good to clarify extensibility for spec 1.0.

protocol.html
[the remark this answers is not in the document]
timbl: Here is the logic: If you don't demand the `a :Patch` triple, then your hands are tied introducing any other functions which use the same properties as PATCH but have different semantics.  there will be nothing you can do so stop the PATCH engine erroneously taking them as input.  For example suppose we make a sort of `-n` version `DontActuallyPatch` which checks the input but doesn;t  write the data.   The analalogy was adding `TemporaryAuthroization` to teh acl system, would have been impossible were we not insisting that current ACL engines require `Authorisation`.  Is that clear? 
So by the way that means also do not mak e Patch the range of its properties or we will indirectly get the same problem.

protocol.html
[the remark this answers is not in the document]
timbl: I agree it is important to remove the requirement that there are no other triples.  That's not the way we work with RDF.  We wok by ignoring triples which we don't understand.   And when we add triples, we make designs which will be fine if the triples are ignored.   (See the discussion on whether the Patch rdf:type is reuired, which allows us to make new things with different semantics).

protocol.html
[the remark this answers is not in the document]
RubenVerborgh: I follow that argument; should we introduce a subclass then? Like `:InsertWherePatch` or `:SimplePatch`?

protocol.html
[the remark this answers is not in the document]
RubenVerborgh: I agree. I initially wanted to take the shortcut of allowing no other triples for 0.9, but we can bypass this if we introduce a specific type. Resuming in https://github.com/solid/specification/pull/346#discussion_r768611736

kjetilk:
Merging this will close #332 and #220 . @timbl 's suggestion to require class should be incorporated first.

RubenVerborgh:
Will implement and re-request review.

RubenVerborgh:
Rebased; requested change in cd44b697341e7cd6568e2ee5029037477f9717d6 /CC @timbl

protocol.html
```diff
@@ -1102,6 +1165,8 @@ <h3 property="schema:name">Normative References</h3>
                     <dd><a href="https://httpwg.org/specs/rfc8288.html" rel="cito:citesAsAuthority"><cite>Web Linking</cite></a>. M. Nottingham.  IETF. October 2017. Proposed Standard. URL: <a href="https://httpwg.org/specs/rfc8288.html">https://httpwg.org/specs/rfc8288.html</a></dd>
                     <dt id="bib-solid-oidc">[SOLID-OIDC]</dt>
                     <dd><a href="https://solid.github.io/solid-oidc/" rel="cito:citesAsAuthority"><cite>SOLID-OIDC</cite></a>. Aaron Coburn; elf Pavlik; Dmitri Zagidulin.  W3C Solid Community Group. W3C Editor's Draft. URL: <a href="https://solid.github.io/solid-oidc/">https://solid.github.io/solid-oidc/</a></dd>
+		    <dt id="bib-sparql11-query">[SPARQL11-QUERY]</dt>
```
csarven, line 1168:
```suggestion
                    <dt id="bib-sparql11-query">[SPARQL11-QUERY]</dt>
```

protocol.html
```diff
@@ -719,6 +721,63 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+
+                    <p id="server-patch-n3-default" rel="schema:hasPart" resource="#server-patch-n3-default"><span property="schema:description">While other specifications might provide a structure and interpretation for a wider class of N3 Patch documents, the present specification only governs the application of N3 Patch documents that additionally adhere to the following constraints:</span></p>
+
+                    <ul about="#server-patch-n3-default">
+                      <li id="server-patch-n3-single" rel="spec:requirement" resource="#server-patch-n3-single"><span property="spec:statement">The patch document <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> contain exactly one patch resource, identified by one or more of the triple patterns described above, which all share the same <code>?patch</code> subject.</span></li>
+                      <li id="server-patch-n3-simple-type" rel="spec:requirement" resource="#server-patch-n3-type">A patch resource <span rel="spec:requirementLevel" resource="spec:MAY">MUST</span> contain a triple <code><span property="spec:statement">?patch rdf:type solid:InsertDeletePatch</span></code>.</li>
```
csarven, line 746:
```suggestion
                      <li id="server-patch-n3-simple-type" rel="spec:requirement" resource="#server-patch-n3-type">A patch resource <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> contain a triple <code><span property="spec:statement">?patch rdf:type solid:InsertDeletePatch</span></code>.</li>
```

protocol.html
```diff
@@ -719,6 +721,63 @@ <h4 property="schema:name"><span>Note</span>: Conditional Update</h4>
...
+
+                    <p>An <em>N3 Patch</em> is a document in the <em>Notation3 (N3)</em> format [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>], identified by the media type <code>text/n3</code>, conforming to the following constraints:</p>
+
+                    <ul>
+                      <li id="server-patch-n3-patches" rel="spec:requirement" resource="#server-patch-n3-patches"><span property="spec:statement">A patch document <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> contain one or more patch resources.</span></li>
+                      <li id="server-patch-n3-patch-identifier" rel="spec:requirement" resource="#server-patch-n3-patch-identifier"><span property="spec:statement">A patch resource <span rel="spec:requirementLevel" resource="spec:MAY">MUST</span> be identified by a URI or blank node, which we refer to as <code>?patch</code> in the remainder of this section.</span></li>
```
csarven, line 734:
```suggestion
                      <li id="server-patch-n3-patch-identifier" rel="spec:requirement" resource="#server-patch-n3-patch-identifier"><span property="spec:statement">A patch resource <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> be identified by a URI or blank node, which we refer to as <code>?patch</code> in the remainder of this section.</span></li>
```

RubenVerborgh:
@csarven Thanks, amended 2a99d6a13b0e86a88a926b33b1d1e5459f9b5e97

kjetilk (approved):
OK, great! I think we should think economy when requiring new triples, but indeed, the extensibility that this affords is important.

timbl (approved):

csarven:
dokieli implements clients side of N3 Patch: https://github.com/linkeddata/dokieli/commit/10214271232c2f88e27c97ac74a7cc8111c1e6ae

Currently used for patching a Memento TimeMap resource:

timemap-n3-patch

Issue #348, opened 2021-11-12
Add references for "triple" (and "triple pattern") to the protocol document


RubenVerborgh:
These terms are currently not defined/referenced.
See https://github.com/solid/specification/pull/346#discussion_r747915497

kjetilk:
Apart from SPARQL, I don't really know which reference that should be... I guess we could be looking back to the early days of semweb data access to see if there is something that predates SPARQL that we could refer to. I think @ericprud is the one who tends to hang around here who were [present at that time](https://lists.w3.org/Archives/Public/public-rdf-dawg/2004AprJun/0385.html). Do you have any suggestions, @ericprud ?

If not, I think we would either refer SPARQL as the normative reference (as I suspect that is actually the earliest spec even though the idea had been around), or we add a simple definition in the preamble of the protocol spec.

ericprud:
I guess [RDF Concepts:triple](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-triple) and [SPARQL:Triple pattern](https://www.w3.org/TR/sparql11-query/#defn_TriplePattern) are the most foundational you'll find in relatively formalized specs. SPARQL experience came from other languages like [RDQL](https://www.w3.org/Submission/2004/SUBM-RDQL-20040109/) and [Algae](https://www.w3.org/2004/05/06-Algae/) but SPARQL didn't reference them.

csarven:
The Solid Protocol should take care to not (re?)define these terms. Makes sense to reference the specs Eric mentions - and we do refer to RDF 1.1 already in different context. We can include the terms under Terminology with external references.

If we still can't capture this properly in context of N3 Patch, we (Solid CG) should coordinate with the RDF-DEV CG for guidance. It should be defined on a lower level spec as opposed to the Solid Protocol.

kjetilk:
Thanks a lot @ericprud ! Yeah, I agree with @csarven , we shouldn't depart from these existing normative references. So, even though we aren't taking in SPARQL now, I think we would need to use its definition of triple pattern.

justinwb:
> Makes sense to reference the specs Eric mentions - and we do refer to RDF 1.1 already in different context. We can include the terms under Terminology with external references.

Agree with this as well 👍

kjetilk:
There's no assignee, have anybody started writing a PR? If not, I can look at it tonight.

kjetilk:
Fixed in #351.

Issue #347, opened 2021-11-10
Use IRIs for resource identification


jakubklimek:
Is there a reason IRIs (Internationalized Resource Identifiers) (RFC3987) are not used for [resource identification](https://solidproject.org/TR/protocol#uri) even though they are part of the RDF 1.1 specification and, in fact, they are already widely used?

I understand that on the level of HTTP, which may be the base of the Solid protocol, only URIs are used. However, I have the feeling the relationship of URIs and IRIs should be addressed somehow. In RFC3987 there is a mapping (percent-encoding) of IRIs to URIs and back. However, e.g. in RDF Turtle, IRI and its percent-encoded URI counterpart, are [treated as two different IRIs](https://www.w3.org/TR/turtle/#sec-escapes), therefore some issues could arise.

Maybe this was already discussed, but I cannot see any mention of IRIs in [the specification](https://solidproject.org/TR/protocol).

damooo:
Currently CSS also assume identifiers as URIs.
in `.acl` resources for a resource, CSS uses percent-encoded URI of that resource.

joachimvh:
> Currently CSS also assume identifiers as URIs. in `.acl` resources for a resource, CSS uses percent-encoded URI of that resource.

There is an open issue on what to do with this at https://github.com/solid/web-access-control-spec/issues/93

Also related: https://github.com/solid/specification/issues/22

kjetilk:
I don't know if it is actually the concern, but IRIs expand the attack surface for [homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack) quite significantly. It makes sense to tread carefully in this area.

damooo:
> I don't know if it is actually the concern, but IRIs expand the attack surface for [homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack) quite significantly. It makes sense to tread carefully in this area.

that is concern only for case of domain-names. not w.r.t resource identifiers in given domain. Even that case is not a technical vulnerability per se, but of naive usage. This need not prevent us in embracing iris, which enable international users name resources in their scripts.

elf-pavlik:
> which enable international users name resources in their scripts

Human-readable labels should be part of the resource description. IRI preferably stays machine-readable, especially everything after the domain name. Most mobile browsers will not even show anything else to the user.

kjetilk:
> that is concern only for case of domain-names. not w.r.t resource identifiers in given domain. 

[...]

PR #350, opened 2021-11-16
Add definitions for triples and triple patterns

kjetilk:
Addressing #348 . 

I decided to paraphrase, as the formal definitions in particular for SPARQL, did not seem appropriate to put our document. Please check that out. I also decided to constrain to URIs, as we generally have done, but see #347 . 

I'm not particularly confident this is the right thing to put in there, but the normative references are there for those who need them.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -369,6 +369,9 @@ left:4.5em;
                     <dt about="#root-container" property="skos:prefLabel" typeof="skos:Concept"><dfn id="root-container">root container</dfn></dt>
                     <dd about="#root-container" property="skos:definition">A root container is a container resource that is at the highest level of the collection hierarchy.</dd>
 
+                    <dt about="#triple" property="skos:prefLabel" typeof="skos:Concept"><dfn id="triple">triple</dfn></dt>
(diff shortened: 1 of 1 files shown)


damooo:
URI restriction must be only about identifiers assigned to information-resources managed by solid server, **independent of triple definition**.
It must no way restrict general definition of triple to have it's terms to be only uris. Thus it just bars away much of linked-data.
According to spec we there cannot be triple `<<dbpedia:rama> dbpedia:wikipage wikipedia:राम>>.` in a resource `<exampl.org/about-rama.ttl>`.
It is contrary to rdf-spec. `subject`, `predicate`, `object` identifiers can be any `iri` in a triple. 
Only identifiers we assign to solid-resources is constrainable, whether they are used in a triple or in any position or not.

kjetilk:
Yes, indeed. But if we defined it in terms of a IRI, we'd have to define IRI too. Perhaps we shouldn't introduce definitions in our document at all, but rather reference the relevant specifications in the places we use those terms?

kjetilk:
Solid Editors resolved to close this and instead merge #351 .

PR #351, opened 2021-11-17
Alternative references for triple and triple pattern

### LINKED ISSUE #348: Add references for "triple" (and "triple pattern") to the protocol document

kjetilk:
So, perhaps better to simply add references like this does to resolve #348 than #350 , which may have been too much and would introduce new problems.


dmitrizagidulin (approved):
👍

justinwb (approved):

Issue #343, opened 2021-11-07
Specify existing practice for Container data about contents


timbl:
The data about each contained resource which is given by the containing Container has been stable for at least 6 years, in NSS and so it is appropriate to take existing well-established practice as the current Solid spec, and as the default for any future versions.

This issue requires that the current situation be documented in the solid protocol before any additions or subtractions are considered .

RubenVerborgh:
The situation might be a little more complex; NSS operates only in the subspace of the ecosystem where the backend is filesystem. In that world, things like "file size" and "content type" are well defined; in other worlds, they are —and hence not in the wider Solid ecosystem.

@timbl Would you be okay with SHOULDs rather than MUSTs?

RubenVerborgh:
See https://github.com/solid/specification/issues/227

jeff-zucker:
It seems to me that a blob of a .png stored in a database still has the media-type "image/png".  And that if the user wants to download it, its size is knowable.

timbl:
The data about contained resource `?x` seems to currently 

|  Subject | Predicate | Object | Object type |
| --- | :-- | :-- |  :-- | 
| `?x` | rdfs:type | A class whose URI  formed by concatenating `'http://www.w3.org/ns/iana/media-types/`, the Internet Content Type of the resource, and `#Resource` | rdfs:Class |
| `?x` | stat:size | An integer  giving the size of the resource in bytes | xsd:integer |
| `?x` | dct:modified | The time when the resource was lastmodified | xsd:dateTime |
| `?x` | stat:mtime | The  unix time when the resource was lastmodified | xsd:integer |

using the [usual prefix conventions](https://github.com/solid/solid-namespace/blob/main/index.js).

RubenVerborgh:
> It seems to me that a blob of a .png stored in a database still has the media-type "image/png". And that if the user wants to download it, its size is knowable.

All fine; the problem are RDF resources stored in a triple-based backend.
Which of the dozen RDF serializations is their content type? (none/all)
What size do they have? (none/all)

I.e., content type and size are a function of content negotiation parameters.

timbl:
@RubenVerborgh I'd not be happy just making this a SHOULD, as basically that makes it something which a client could never depend on.    Maybe something like "MUST unless that information does not exist is really expensive to find out"

[...]

Issue #154, opened 2016-01-13
Add support for Last-Modified header


dmitrizagidulin:
If a resource is readable, the `Last-Modified` header should be returned.

(Originally opened by @csarven)
(Moved from https://github.com/linkeddata/gold/issues/64)

From RFC 7232:

> An origin server SHOULD send Last-Modified for any selected
>    representation for which a last modification date can be reasonably
>    and consistently determined, since its use in conditional requests
>    and evaluating cache freshness ([RFC7234]) results in a substantial
>    reduction of HTTP traffic on the Internet and can be a significant
>    factor in improving service scalability and reliability.
> https://tools.ietf.org/html/rfc7232#section-2.2.1

csarven:
The current spec sets MUST for 7232 (and actually SHOULD for Last-Modified in there) for servers, and MAY for clients.

For 7234, server SHOULD, client MAY.

As it stands, there is no particular functionality that needs Last-Modified as MUST for server. It can however come from elsewhere and that would be orthogonal.

The Best Practices and Guidelines document can make suggestions.

csarven:
Given the changes in the Solid Protocol involving the `Last-Modified` header field ( https://solidproject.org/ED/protocol , https://solidproject.org/TR/2024/protocol-20240512 ), this issue can be revisited to make the requirement explicit. A PR to resolve this issue is available at: https://github.com/solid/specification/pull/678.

PR #352, opened 2021-11-17
Authoritative Contained Resource Data

csarven:
Issues concerning:
* https://github.com/solid/specification/issues/227
* https://github.com/solid/specification/issues/343

---

The concept #authoritative-information describes an RDF constraint that's analogous to "authoritative metadata" as per https://www.w3.org/2001/tag/doc/mime-respect - the sender's HTTP message is considered to be authoritative.

Server receiving the message can either respect sender's intentions or responds with redirect or error message. For example, in #server-protect-authoritative-resource-data , when a client requests to update a container's description including authoritative data about the contained resources, the server will reject.

The concept #authoritative-resource-data specialises #authoritative-information in that it is used in context of resource descriptions. #authoritative-contained-resource-data is one application of authoritative information in the container description.


protocol.html
```diff
@@ -369,6 +369,9 @@ <h3 property="schema:name skos:prefLabel">Terminology</h3>
                     <dt about="#root-container" property="skos:prefLabel" typeof="skos:Concept"><dfn id="root-container">root container</dfn></dt>
                     <dd about="#root-container" property="skos:definition">A root container is a container resource that is at the highest level of the collection hierarchy.</dd>
 
+                    <dt about="#authoritative-information" property="skos:prefLabel" typeof="skos:Concept"><dfn id="authoritative-information">authoritative information</dfn></dt>
+                    <dd about="#authoritative-information" property="skos:definition">Authoritative information encompasses data about resources described by means of RDF statements [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>] supplied by clients, servers, and intermediaries</dd>
```
RubenVerborgh, line 373: I'm not sure I can find counterexamples here; to me it seems this is _all_ RDF data. Is that correct? Why is it authoritative?

protocol.html
```diff
@@ -559,6 +572,40 @@ <h3 property="schema:name">Resource Containment</h3>
...
+
+                  <section id="authoritative-resource-data" inlist="" rel="schema:hasPart" resource="#authoritative-resource-data">
+                    <h4 property="schema:name">Authoritative Resource Data</h4>
+                    <div datatype="rdf:HTML" property="schema:description">
+                      <p>Container descriptions are not limited to containment triples. To further support client navigation and application interaction, servers can include <a href="#authoritative-information">authoritative information</a> about contained resources as part of the container description.</p>
+
+                      <p><span about="" id="authoritative-contained-resource-data" rel="spec:requirement" resource="authoritative-contained-resource-data"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> include the following information about contained resources as part of the container description, unless that information does not exist or is expensive to determine:</span></span></p>
```
RubenVerborgh, line 581: Agreed, but how are we going to test this? The "unless" basically means a flag for the test suite then?

protocol.html
```diff
@@ -559,6 +572,40 @@ <h3 property="schema:name">Resource Containment</h3>
...
+
+                      <dl>
+                        <dt><code>rdf:type</code></dt>
+                        <dd>A class whose URI formed by concatenating <code>http://www.w3.org/ns/iana/media-types/</code>, the IANA Media Type [<cite><a class="bibref" href="#bib-iana-media-types">IANA-MEDIA-TYPES</a></cite>] of the resource, and <code>#Resource</code>.</dd>
```
RubenVerborgh, line 585: We need to mention that no escaping is performed, so the slash remains as-is.

protocol.html
```diff
@@ -559,6 +572,40 @@ <h3 property="schema:name">Resource Containment</h3>
...
+                        <dt><code>stat:size</code></dt>
+                        <dd>A non-negative integer giving the size of the resource in bytes.</dd>
+                        <dt><code>dcterms:modified</code></dt>
+                        <dd>The time when the resource was last modified.</dd>
+                        <dt><code>stat:mtime</code></dt>
+                        <dd>The Unix time when the resource was last modified.</dd>
+                      </dl>
```
RubenVerborgh, line 592: The above: accurate to what unit? Or a general statement about required accuracy?

protocol.html
```diff
@@ -559,6 +572,40 @@ <h3 property="schema:name">Resource Containment</h3>
...
+
+                      <div class="note" id="authoritative-resource-data-logically-part-of-container-description" inlist="" rel="schema:hasPart" resource="#authoritative-resource-data-logically-part-of-container-description">
+                        <h5 property="schema:name"><span>Note</span>: Authoritative Resource Data Logically Part of Container Description</h5>
+                        <div datatype="rdf:HTML" property="schema:description">
+                          <p>Authoritative data about contained resources is logically part of the container. Servers are encouraged to consider omitting authoritative data about a contained resource when agent is unauthorized to read the contained resource.</p>
```
RubenVerborgh, line 597: MAY? SHOULD?

protocol.html
```diff
@@ -559,6 +572,40 @@ <h3 property="schema:name">Resource Containment</h3>
...
+                        </div>
+                      </div>
+
+                      <p><span about="" id="contained-resource-state-cascading" rel="spec:requirement" resource="#contained-resource-state-cascading"><span property="spec:statement">When the state of a contained resource changes, the <span rel="spec:requirementSubject" resource="spec:Server">servers</span> <span rel="spec:requirementLevel" resource="spec:MUST-NOT">MUST NOT</span> expose any authoritative information that has a cascading potential to its immediate container.</span></span></p>
```
RubenVerborgh, line 601: I do not (fully) understand.

RubenVerborgh (approved):
Aligns in spirit with what was decided.

protocol.html
```diff
@@ -559,6 +572,40 @@ <h3 property="schema:name">Resource Containment</h3>
...
+
+                      <dl>
+                        <dt><code>rdf:type</code></dt>
+                        <dd>A class whose URI formed by concatenating <code>http://www.w3.org/ns/iana/media-types/</code>, the IANA Media Type [<cite><a class="bibref" href="#bib-iana-media-types">IANA-MEDIA-TYPES</a></cite>] of the resource, and <code>#Resource</code>.</dd>
+                        <dt><code>stat:size</code></dt>
+                        <dd>A non-negative integer giving the size of the resource in bytes.</dd>
+                        <dt><code>dcterms:modified</code></dt>
+                        <dd>The time when the resource was last modified.</dd>
+                        <dt><code>stat:mtime</code></dt>
+                        <dd>The Unix time when the resource was last modified.</dd>
```
kjetilk, line 591: This is incorrect. 

NSS does not say when the resource was last modified, as can be seen for example at https://kjetilktest.solidcommunity.net/public/ , prefixes omitted for brevity:

```turtle
@prefix : <#>.
@prefix dct: <http://purl.org/dc/terms/>.
@prefix ldp: <http://www.w3.org/ns/ldp#>.
@prefix stat: <http://www.w3.org/ns/posix/stat#>.
@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
@prefix pub: <>.
@prefix test: <test/>.

pub:
    a ldp:BasicContainer, ldp:Container;
    dct:modified "2021-10-26T09:56:26Z"^^xsd:dateTime;
    ldp:contains test:;
    stat:mtime 1635242186.156;
    stat:size 4096.
test:
    a ldp:BasicContainer, ldp:Container, ldp:Resource;
    dct:modified "2021-10-26T10:04:40Z"^^xsd:dateTime;
    stat:mtime 1635242680.624;
    stat:size 4096.
```

As can be seen from the most recent timestamp, this resource was updated at the latest at 2021-10-26T10:04:40Z. This resource is given by the `@prefix pub: <>.`, which has an older timestamp.

In reality, NSS sets the last modification to the time determined by inspecting the inode of the underlying filesystem as defined by POSIX. This is *not* the last modification time of the resource, as evident from the above. This can also be verified from the NSS source code: https://github.com/solid/node-solid-server/blob/693e7a15d5cae3d4b4f328ef1810ee6c4ecb7efd/lib/ldp-container.js#L117-L126

The same thing can be seen by various other methods, but this is the simplest illustration I know of. 

So, `stat:mtime` and `dct:modified` needs to be defined in different terms. I guess we could bring in POSIX as a normative reference, but it is not an open standard. I'm not sure how we'd resolve this, actually, that's why I have tried to point out this problem numerous times...

protocol.html
kjetilk: @csarven is just trying to accommodate my concern from https://github.com/solid/specification/issues/227#issuecomment-931202286 but he can't. It is trying to reconcile two mutually exclusive ideas. If certain properties like mtime is defined as the mtime of the resource, it will be cascading. There's no way to avoid that. So, NSS doesn't do that, it defines it based on the mtime of the filesystem inode POSIX. So, the exercise we have to go through is, if we don't want to do like POSIX and use an auxiliary resource, to define mtime in a way that is consistent with Web principles, but not in terms of modification time for a resource. I've shouted as loud as I possibly can about this problem...

kjetilk (changes requested):
I'm just going on about this: `dct:modified` and `stat:mtime` isn't defined as the relevant times for the resource, as detailed in the comments. It can't be. It has to be defined in other terms.

protocol.html
acoburn: If the representation contents may change based on who is viewing the resource, what are the implications for caching?

protocol.html
acoburn: Does this mean that an agent cannot otherwise use these predicate in the RDF description of a container resource?

protocol.html
```diff
@@ -708,6 +755,8 @@ <h3 property="schema:name">Writing Resources</h3>
...
 
+                  <p><span about="" id="server-protect-authoritative-resource-data" rel="spec:requirement" resource="#server-protect-authoritative-resource-data"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> NOT allow HTTP <code>POST</code>, <code>PUT</code> and <code>PATCH</code> to update the <a href="#authoritative-resource-data">authoritative resource data</a>; if the server receives such a request, it MUST respond with a <code>409</code> status code.</span></span> [<a href="https://github.com/solid/specification/issues/227#issuecomment-919312592" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
acoburn, line 758: How does one distinguish between the `dcterms:modified` triples that are "authoritative metadata" and the `dcterms:modified` triples that a client supplies? Or are all attempts (malicious or benign) to use `dcterms:modified` as a predicate to be rejected by a server?

protocol.html
kjetilk: To try to be a little more constructive here, there is a possibility to say something like 

"The time where the non-authoritative parts of the representation of the resource was modified."

That covers a bit more than it needs to, but there should be no harm to that.

protocol.html
csarven: I've updated the comment to include an overview which was originally intended as a response to your question here. Hope that helps.

protocol.html
csarven: Servers can use their own precision.

protocol.html
csarven: The condition is an exit strategy for servers that essentially cannot or do not wish to provide the authoritative information. Inapplicable.

The test suite would require a flag if the reports wish to distinguish the outcome of the test for subjects that implement the required feature from those subjects that it is "inapplicable". (EARL takes this into account.)

I've generalised the condition using "inapplicable" and then expanded on it in #authoritative-resource-data-logically-part-of-container-description .

When the group reaches consensus on a more precise requirement (with an exception or otherwise), we can update the specification.

protocol.html
csarven: If the following is not agreeable, the advisement need not be in the Note:

I wanted to have an initial acknowledgement of the ongoing considerations until opposing stances are reconciled or a compromise is reached. It is an advisement for servers indicating the case in which they can consider omitting information - this is a legitimate concern for "mission critical systems". The advisement entails that omitting information based on authorization would require the server to do more work and handle the implications - however, this is currently not described in the specification.

protocol.html
csarven: Reject all attempts. Authoritative information about contained resources in the container description are strictly claimed and protected by the server. In the same way containment statements are protected by the server. Moreover, thus far there is no compelling case for clients to update these statements about contained resources in the container description. There are no additional restrictions on clients attempting to claim such statements anywhere else in the storage.

protocol.html
csarven: I've updated the text in the meantime that hopefully better communicates this:
https://github.com/solid/specification/pull/352/commits/c0d8d4cea9b66a2914814714b134e9b1cc6042fa

protocol.html
csarven: I've updated to be more specific about the URI that's being constructed.. which is actually a URI Template:

https://github.com/solid/specification/pull/352/commits/5ea32a4e440eb952b9e11ec0ec97cc27c4ca843e

I thought the current language would imply that no encoding transformation is applied on the value, but if you think that it'd be better to use something like:

"..expands to a value <ins>without an encoding transformation</ins> that is.."

we can do that.

protocol.html
kjetilk: Yeah, it is probably better left out and dealt with a different mechanism.

protocol.html
kjetilk: I have a general problem with this type of formulation, "When the server... it MUST" or "MUST ... unless". I find that it mostly serves to obscure the actual requirement level and complicate the text. It makes testing more complicated, and it does not help interoperability. Either, it is something you can rely on, or you can't.

I think it would be much clearer to use SHOULD in this case, and explain in a non-normative note in which cases it is OK to consider not doing it.

protocol.html
kjetilk: I think the definition needs to be more precise. For it to have value as a definition, it needs to concisely describe what is meant in this context. In this context, we are talking about server-supplied metadata, right?

csarven: ? No. As said, it is a broader concept covering the sender's intention. What can be improved in the definition is what constitutes authoritative. I've expressed that elsewhere with references, but will look into revising.

protocol.html
csarven: I don't think it complicates testing but I'll take your word for it.

Inapplicable is a valid test outcome. It is not always straight forward to have a requirement with a specific requirement level without additional context. "MUST, unless" or "SHOULD" still generally end up in the same category I think.

(MUST unless is not unheard of in specs. LDP has several "MUST (NOT), unless", if that means anything - probably not)

The current text reflects the discussion in issue 343.

kjetilk: Yeah, I know it is being formulated that way. It may be OK in a contexts such as HTTP, where specific applications like Solid needs to establish additional constraints anyway, but I feel it gets very awkward with Solid and LDP.

I think that with a MUST, a test failure is clearly an indication of non-compliance, so you need to code into the tests further intentions of the spec as special cases. In the case of a SHOULD, it is fair to assume that a test failure is probably a case of inapplicable. There's a nuance there, even though the end result is indeed the same, the test is inapplicable.

csarven: I don't think a failure on a SHOULD is necessarily "inapplicable". That should be a "failed". "inapplicable" or "cantTell" (again from EARL) are possible outcome values provided that there is some context thrown in. And then there is the "info" as part of the test result which would probably explain what/why..

Any way, we are discussing two things here 1) readability/correctness of the requirement 2) validity/usefulness of the test reports. I generally agree with you on both accounts. I'm just saying that for "inapplicable" to happen, there needs to be some context given and that's not due to the requirement level - putting aside the fact why MUST or a SHOULD would be preferable for that particular requirement.

protocol.html
kjetilk: Yes, but the broad concept doesn't seem useful to us, AFAICT. Anyway, looking forward to clarifications.

protocol.html
kjetilk: It shouldn't be a "cantTell", that's for when the outcome of the test is not certain, and you can't fail a SHOULD or MAY test, because it might just be that you decided to not do it, which is OK. We may add something to the harness where implementers may say that they intend to support a SHOULD or MAY, and therefore fail would be OK. But for now, among the EARL options, it leaves inapplicable. Arguably, a "optional" outcome value or something like that might be desireable.

Nevertheless, I acknowledge I might be conflating things, I guess I should hold my fire on those things until we can have a video chat about it, I feel such issues should not hold up 0.9, but I also don't want to delay the test suite over it.

protocol.html
```diff
@@ -755,7 +755,7 @@ <h3 property="schema:name">Writing Resources</h3>
...
+                  <p><span about="" id="server-protect-authoritative-resource-data" rel="spec:requirement" resource="#server-protect-authoritative-resource-data"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST-NOT">MUST NOT</span> allow HTTP <code>POST</code>, <code>PUT</code> and <code>PATCH</code> to update the <a href="#authoritative-resource-data">authoritative resource data</a>; if the server receives such a request, it MUST respond with a <code>409</code> status code.</span></span> [<a href="https://github.com/solid/specification/issues/227#issuecomment-919312592" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
kjetilk, line 758: I couldn't find a namespace document, so I've been using `spec:MUSTNOT`. Only one of those have been merged, so it isn't a big change, but we need to be consistent throughout, so I just want to raise it.

csarven: It is `MUST-NOT`: https://github.com/solid/vocab/blob/specification-terms/spec.ttl

protocol.html
csarven: I'd like to update the requirement at some point specifying the particular circumstances in which the server may not honour the requirement. Right now it is a bit vague (and not exciting). Some of the mentions currently in the Note may need to be more precise and move into normative language.

protocol.html
```diff
@@ -369,6 +369,9 @@ <h3 property="schema:name skos:prefLabel">Terminology</h3>
                     <dt about="#root-container" property="skos:prefLabel" typeof="skos:Concept"><dfn id="root-container">root container</dfn></dt>
                     <dd about="#root-container" property="skos:definition">A root container is a container resource that is at the highest level of the collection hierarchy.</dd>
 
+                    <dt about="#authoritative-information" property="skos:prefLabel" typeof="skos:Concept"><dfn id="authoritative-information">authoritative information</dfn></dt>
+                    <dd about="#authoritative-information" property="skos:definition">Authoritative information encompasses data about resources described by means of RDF statements [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>] that can be supplied by clients, servers, and intermediaries.</dd>
```
justinwb, line 373: The definition feels like it could use a bit more detail on what it means for this information to be authoritative. Does it mean that consumers should be able to rely on its accuracy? Can any data about a resource provided by anyone be considered authoritative since clients, servers, and intermediaries are mentioned?

protocol.html
justinwb: Acknowledging the issue raised by @kjetilk, but at a minimum it could a be bit more clear for the reader what the text is trying to accomplish.

protocol.html
csarven: @kjetilk (partial response):
In your example with the NSS behaviour, the description of public/ including authoritative data about itself (the container) is not covered by the requirement in this PR.

What's requested is a requirement on authoritative data about contained resources in the description of a container (this PR). If I overlooked the call to have authoritative data about the container as well in the container description, can someone point me to it?

csarven: (partial response continued..)

Although this is not explicit (but we can make it so..), recipient observing `dcterms:modified` may interpret/expect that the value would correspond with the `Last-Modified` header value in the response of the contained resource, if the recipient were to request it at that time. Perhaps of minor importance but the `Last-Modified` header remains as a SHOULD (as is from RFC 7232): https://github.com/solid/specification/issues/154 .

csarven: NSS behaviour seems to match POSIX. `mkdir foo; mkdir foo/bar; mkdir foo/bar/baz; touch foo/bar/baz/qux;` (with a second in between each command just to spot time diffs with `stat FILE`). NSS is just using `stat.mtime` after all.

timbl: So you are saying that what NSS and CSS presumably do  works because it gives a sense of when the directory was changed in terms o the set of things contained, which does not propagate, which is important.  But it has  formally the problem that if we look at when the content you get from a GET has changed, then for example a client would assume that it does not have to refresh its cache of the container resource because it has a copy of that date.    When in fact the GET contents would have changed because the size or date of a contained resource have changed.  Yes, Kjetil, we must explain that it is the last modified dat of the structure which does not change with the metadata about the contents.  To avoid change propagation all the way to the top with every patch anywhere in the tree.  It is the Structure Last Modified time.

kjetilk: First, @csarven , it is not just the self-description that is the issue here, that `dct:modified` does not reflect the resource modification time and thus is different from the value of `Last-Modified` can also be verified with for example
```sh
curl -D - -H "Accept: text/turtle" https://kjetilktest.solidcommunity.net/
```

As I believe caching is extremely important to the success of Solid, anything that can get in conflict with that is a red flag for me. I would much prefer that whenever we define what the value of `dct:modified` should be that we reference the definition of `Last-Modified` in [Section 2.2 of RFC7232](https://datatracker.ietf.org/doc/html/rfc7232#section-2.2) (but allow for greater precision).

I have a proposal that I'd like to implement that would show how this can be done, that would enable last-modified tracking for all resources, and wouldn't have the problem of cascading. I don't think I should do that right now, because I believe we urgently need to publish 0.9, but I would be sleeping better at night (literally!) if we could leave the option open to do that later. 

For 0.9, I believe we can get where we need to be if we just make it clear that for now, `dct:modified` and `stat:mtime` does not reflect the last modification for the resource, just some of the data. Calling that "Structure Last Modified time" works for me.

csarven: Good that we share the same interests/care on caching (e.g., "Caching, caching, caching" in the first comment in issue 227 on 2021-02.)

And tracking down TAG findings to communicate/apply ideas:

>For example, Last-Modified is used for representation "metadata". dcterms:modified is for RDF "data".

or what I wrote just two comments ago above about the relationship between the two.

>it is not just the self-description that is the issue here

Well aware. The examples in 227 with the headers served just fine. The self-description example here wasn't necessary or at least seemed confusing. I had to rule it out.

---

I wanted to get to the relevant 7232 bit:

> How that value is
   determined for any given resource is an implementation detail beyond
   the scope of this specification.

We have more freedom in explaining how the Last-Modified value is determined for the container than redefining dcterms:modified (or am I misunderstanding you? or is this effectively the same in the end?)

---

I think we should drop stat:mtime for 1.0.

(I don't see an issue with a server basing Last-Modified off mtime. That's fine by 7232.)

protocol.html
csarven: I may not be interpreting GitHub's UI properly. At first I though you're referring to mtime which is why I said "precision" thinking that you meant whether subseconds can be expected or not but now I'm thinking that you wanted to know about dcterms:modified and whether that should be a "date and time" (perhaps even implying xsd:dateTime). So, I updated that any way in https://github.com/solid/specification/pull/352/commits/4734bbf53e3dea52ea7d537660dacf1e7d298814 but it'd be good if you can clarify further.

RubenVerborgh: Both are fine, thanks.

protocol.html
csarven: I've expressed it around the concept of "resource metadata" in place of "authoritative information" in https://github.com/solid/specification/pull/352/commits/1727bb80ac6e8bf026d69d16045ed217b11b6cf8 . We can come back to "authoritative information" later e.g., after "server description" (Issue #355 ).

protocol.html
csarven: Clarified the Note in https://github.com/solid/specification/pull/352/commits/f599dd5d9df818ecaa1217359c45ba66a88d5e04 and https://github.com/solid/specification/pull/352/commits/98d2db4ec1d0438cd455163748df413f967cdb36

Deferring to https://github.com/solid/specification/issues/343 for changes to requirement level and applicability.

protocol.html
csarven: Clarified the Note in https://github.com/solid/specification/pull/352/commits/f599dd5d9df818ecaa1217359c45ba66a88d5e04 and https://github.com/solid/specification/pull/352/commits/98d2db4ec1d0438cd455163748df413f967cdb36

protocol.html
csarven: Removed https://github.com/solid/specification/pull/352/commits/f2c8c8b82e2c40a1371f73c60671dfc612b2cbbd in favour of https://github.com/solid/specification/pull/352/commits/4eaf3456c4a8d90a2fb9fd931a8135733914da91 and https://github.com/solid/specification/pull/352/commits/30eee02aafb2077ee48c7ddbe1c596554184f3b8

protocol.html
csarven: Clarified container's Last-Modified and how dcterms:modified corresponds with the Last-Modified of a contained resource: https://github.com/solid/specification/pull/352/commits/4eaf3456c4a8d90a2fb9fd931a8135733914da91 , https://github.com/solid/specification/pull/352/commits/30eee02aafb2077ee48c7ddbe1c596554184f3b8

Deferring to https://github.com/solid/specification/issues/343 for further changes to these requirements.

protocol.html
csarven: https://github.com/solid/specification/pull/352#discussion_r762578541

NoelDeMartin:
I haven't read all the comments in this thread, so apologies if this has already been addressed. But I just have a couple of comments/questions.

Reading this part is confusing for me:

> Servers MUST include resource metadata about contained resources as part of the container description, unless that information is inapplicable to the server.

It says "MUST", but then it says "unless that information is inapplicable to the server". Wouldn't that be an implementation detail of the server, and thus something irrelevant in the protocol description? Also, as an app developer reading this I could be mislead to think that all servers provide the data, because it says MUST. But as I understand it, that's not true, depending on some implementation details of the server, some containers won't return this data.

Keeping that in mind, why not just use MAY in that sentence?

Also, I'm not sure why there are two values indicating what seems to be the same information: `dcterms:modified` and `stat:mtime`. Is it just for compatibility with current implementations? Or is there an important difference between the two?

As an app developer, it was confusing to know which one to use, so I think I just picked one at random. Now it mentions that `dcterms:modified` has to match the `Last-Modified` header, so I suppose that's the "cannonical" one?

csarven:
@NoelDeMartin Thank you. The discussion on "MUST, unless" happened here: https://github.com/solid/specification/pull/352#discussion_r751682478 . "unless" has to do with inapplicability. We may change to "SHOULD" but this is to be worked out in https://github.com/solid/specification/issues/343 or next release of the spec. It has to do with specifying inapplicable cases. Unfortunately MAY is not strong enough for a requirement to enable and have this this feature.

stat:mtime may be deprecated in a later version of the spec ( https://github.com/solid/specification/pull/352#discussion_r759724771 ). Indeed rely on dcterms:modified for the foreseeable future.

protocol.html
```diff
@@ -559,6 +572,44 @@ <h3 property="schema:name">Resource Containment</h3>
...
 
                   <p><span about="" id="server-basic-container" rel="spec:requirement" resource="#server-basic-container"><span property="spec:statement">The representation and behaviour of containers in Solid corresponds to LDP Basic Container and <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> be supported by <span rel="spec:requirementSubject" resource="spec:Server">server</span>.</span></span> [<a href="https://github.com/solid/specification/issues/47#issuecomment-561675764" rel="cito:citesAsSourceDocument">Source</a>]</p>
+
+                  <p id="server-container-last-modified">Servers can determine the value of the HTTP <code>Last-Modified</code> header field in response to <code>HEAD</code> and <code>GET</code> requests targeting a container based on modifications to the selected representation or changes to containment triples.</p>
```
RubenVerborgh, line 576: Non-normative?

protocol.html
```diff
@@ -559,6 +572,44 @@ <h3 property="schema:name">Resource Containment</h3>
...
+
+                  <section id="contained-resource-metadata" inlist="" rel="schema:hasPart" resource="#contained-resource-metadata">
+                    <h4 property="schema:name">Contained Resource Metadata</h4>
+                    <div datatype="rdf:HTML" property="schema:description">
+                      <p>Container descriptions are not limited to containment triples. To further support client navigation and application interaction, servers can include <a href="#resource-metadata">resource metadata</a> about contained resources as part of the container description.</p>
```
RubenVerborgh, line 581: Non-normative; should this be a MAY?

RubenVerborgh: Or perhaps be merged with the statement below; the MUST do xyz and MAY add others?

protocol.html
```diff
@@ -559,6 +572,44 @@ <h3 property="schema:name">Resource Containment</h3>
...
+
+                      <p id="contained-resource-metadata-statements" about="#contained-resource-metadata-statements" typeof="skos:Collection"><span property="skos:prefLabel">Contained resource metadata statements</span> include the properties:</p>
+
+                      <dl about="#contained-resource-metadata-statements" rel="skos:member">
+                        <dt about="#contained-resource-metadata-rdf-type" id="contained-resource-metadata-rdf-type" property="skos:prefLabel"><code>rdf:type</code></dt>
+                        <dd about="#contained-resource-metadata-rdf-type" property="skos:definition">A class whose URI is based on the <em>URI Template</em> [<cite><a class="bibref" href="#bib-rfc6570">RFC6570</a></cite>] syntax: <code>http://www.w3.org/ns/iana/media-types/{iana-media-type}#Resource</code>, where <code>iana-media-type</code> expands to a value that is part of IANA Media Types [<cite><a class="bibref" href="#bib-iana-media-types">IANA-MEDIA-TYPES</a></cite>].</dd>
```
RubenVerborgh, line 589: This is incorrect, because the slashes would be escaped. Also the phrasing is off.
```suggestion
                        <dd about="#contained-resource-metadata-rdf-type" property="skos:definition">A class whose URI is the expansion of the <em>URI Template</em> [<cite><a class="bibref" href="#bib-rfc6570">RFC6570</a></cite>] <code>http://www.w3.org/ns/iana/media-types/{+iana-media-type}#Resource</code>, where <code>iana-media-type</code> corresponds to a value from the IANA Media Types [<cite><a class="bibref" href="#bib-iana-media-types">IANA-MEDIA-TYPES</a></cite>].</dd>
```

RubenVerborgh: Additionally, what if a media type is not listed with IANA?

protocol.html
```diff
@@ -559,6 +572,44 @@ <h3 property="schema:name">Resource Containment</h3>
...
+                        <dd about="#contained-resource-metadata-dcterms-modified" property="skos:definition">The date and time when the resource was last modified.</dd>
+                        <dt about="#contained-resource-metadata-stat-mtime" id="contained-resource-metadata-stat-mtime" property="skos:prefLabel"><code>stat:mtime</code></dt>
+                        <dd about="#contained-resource-metadata-stat-mtime" property="skos:definition">The Unix time when the resource was last modified.</dd>
+                      </dl>
+
+                      <p id="dcterms-modified-corresponds-last-modified">The <code>dcterms:modified</code> value of a contained resource corresponds with the <code>Last-Modified</code> header value of the contained resource. If one were to perform <code>HEAD</code> or <code>GET</code> requests on the URI of the contained resource at the time of the HTTP message’s generation, then a response with the <code>200</code> status code including the <code>Last-Modified</code> header would indicate the same date and time.</p>
```
RubenVerborgh, line 598: Non-normative?

protocol.html
RubenVerborgh: Also perhaps best positioned below https://github.com/solid/specification/pull/352/files#diff-dcde422fcd5ad628eda53177df32d7ed1c10ac7ec86e62606b1d8979c812bdfcR598 ?

RubenVerborgh (changes requested):
All good, except for the wrong RFC6570 template, to which I propose a correction.

RubenVerborgh:
Implementer's note: the Community Solid Server supports the text of this PR via configuration: https://github.com/solid/community-server-recipes/blob/feat/spec-352/metadata/config-metadata.json.
Live version for testing at https://drive.verborgh.org/.

protocol.html
csarven: It is a brief statement about the section.

RubenVerborgh:
```suggestion
                      <p>Container descriptions are not limited to containment triples. To further support client navigation and application interaction, servers can include <a href="#resource-metadata">resource metadata</a> about contained resources as part of the container description, as described below.</p>
```

protocol.html
csarven: This statement sits closer to the containment. I thought about refactoring this section - container description - but not in this PR.

protocol.html
csarven: I don't think this is necessary. Merging your suggestion.

protocol.html
csarven: Servers are not required to output other values. If they do, clients may not recognise or dereference the value listed under `http://www.w3.org/ns/iana/media-types/`

RubenVerborgh: Okay; just want to note this means that servers either need to have a (continuously updating) list of IANA media types, or clients might not recognize a value.

kjetilk (approved):
I have bad feelings about this, as we are now saying that it is OK for a server to not track the modification of representations of a resource and give the modification times of just a part of it. It seems likely to me that this will cause many problems for caching in the future, and that servers will serve stale data, and that it will result in difficulty for decentralization efforts.
 
Nevertheless, I acknowledge that it is now an internally consistent structure in line with NSS behavior and that RFC7232 is very liberal on this point, as it says:
> The last-modified time would usually be the most recent time that any of those parts were changed. 

and, what this does is to say "usually, yes, but not for containers in Solid". 

There ought to be a :nose: reaction

protocol.html
csarven: Surely it can be written as normative. I think it also works by describing the intention that's related to how the last-modified header value is determined. Hesitant to over-specify this for 0.9. But I'm all ears.

protocol.html
```diff
@@ -559,6 +572,44 @@ <h3 property="schema:name">Resource Containment</h3>
...
+
+                      <dl about="#contained-resource-metadata-statements" rel="skos:member">
+                        <dt about="#contained-resource-metadata-rdf-type" id="contained-resource-metadata-rdf-type" property="skos:prefLabel"><code>rdf:type</code></dt>
+                        <dd about="#contained-resource-metadata-rdf-type" property="skos:definition">A class whose URI is the expansion of the <em>URI Template</em> [<cite><a class="bibref" href="#bib-rfc6570">RFC6570</a></cite>] <code>http://www.w3.org/ns/iana/media-types/{+iana-media-type}#Resource</code>, where <code>iana-media-type</code> corresponds to a value from the IANA Media Types [<cite><a class="bibref" href="#bib-iana-media-types">IANA-MEDIA-TYPES</a></cite>].</dd>
+                        <dt about="#contained-resource-metadata-stat-size" id="contained-resource-metadata-stat-size" property="skos:prefLabel"><code>stat:size</code></dt>
+                        <dd about="#contained-resource-metadata-stat-size" property="skos:definition">A non-negative integer giving the size of the resource in bytes.</dd>
```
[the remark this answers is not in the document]
csarven: Do you consider stat:size (I assume that's what your comment is referring to - GitHub UI trips me up sometimes) to more of a concern than the others?

Had this text earlier in the Note #contained-resource-metadata-considerations :

>Servers are encouraged to consider omitting authoritative data about a contained resource when an agent is unauthorized to read the contained resource.

Removed it as per Kjetil's suggestion: https://github.com/solid/specification/pull/352#discussion_r754951534

Can revive and put it under #security-considerations but note that it may be equivalent / already captured by:

>Servers are strongly discouraged from exposing information beyond the minimum amount necessary to enable a feature.

---

If you'd like more specific considerations that should be mentioned in there, we can do that as well. Have something in mind?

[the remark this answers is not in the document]
acoburn: Provided that a server can choose to omit this data (under the "inapplicability" clause), this is fine as-is.

Knowing the size of a file that one may not otherwise have access to read, can be extremely useful data when looking to exploit weaknesses in a server. I would have serious reservations about including that information on a Storage that holds sensitive personal data.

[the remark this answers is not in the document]
kjetilk: Yes, I believe that since we do know that we cannot have a server check the authz of every child resource, then is inappropriate to leave it as an implementation detail, we'd need to specify a concrete mechanism. I think it would be better for this to be a SHOULD, as per RFC2119, but for now, it seems that it could be left out under the inapplicability clause.

protocol.html
```diff
@@ -559,6 +572,44 @@ <h3 property="schema:name">Resource Containment</h3>
...
+                        <dt about="#contained-resource-metadata-dcterms-modified" id="contained-resource-metadata-dcterms-modified" property="skos:prefLabel"><code>dcterms:modified</code></dt>
+                        <dd about="#contained-resource-metadata-dcterms-modified" property="skos:definition">The date and time when the resource was last modified.</dd>
```
acoburn, line 593: Assuming that a server provides a `Last-Modified` header and assuming that we do not want to break how caching works for browsers, I do not see how this (`dcterms:modified` or `stat:mtime`) can be implemented without causing a cascade of updates to the root of a Storage.

For example, consider a resource at `/lvl-1/lvl-2/foo.ttl` with the following response headers:

```http
Last-Modified: Sun, 05 Dec 2021 17:45:02 GMT
ETag:  "abcd"
```

The container at `/lvl-1/lvl-2/` has the following headers:

```http
Last-Modified: Tue, 07 Dec 2021 12:23:15 GMT
ETag: "1234"
```

That container would include the following triples:

```turtle
</lvl-1/lvl-2/> 
    dcterms:modified "2021-12-07T12:23:15Z"^^xsd:dateTime ;
    ldp:contains </lvl-1/lvl-2/foo.ttl> .
</lvl-1/lvl-2/foo.ttl>
    dcterms:modified "2021-12-05T17:45:02Z"^^xsd:dateTime .
```

The container at `/lvl-1/` has the following headers:

```http
Last-Modified: Tue, 07 Dec 2021 19:37:22 GMT
ETag: "zyxw"
```

And the following body:

```turtle
</lvl-1/> 
    dcterms:modified "2021-12-07T19:37:22Z"^^xsd:dateTime ;
    ldp:contains </lvl-1/lvl-2/> .
</lvl-1/lvl-2/>
    dcterms:modified "2021-12-07T17:45:02Z"^^xsd:dateTime .
```

Now consider that a client adds a triple to `/lvl-1/lvl-2/foo.ttl`. The result will lead to the `Last-Modified` being updated on `/lvl-1/lvl-2/foo.ttl`. The content of that resource has changed, so the ETag header also changes.

Because the `Last-Modified` header changes, the representation of `/lvl-1/lvl-2/` now also changes to reflect the new status of the contained resource. As a consequence, and in order to ensure that clients receive the latest version of `/lvl-1/lvl-2/`, the `Last-Modified` and `ETag` headers of this container resource must change.

Because the `Last-Modified` header for `/lvl-1/lvl-2/` changes, the representation of `/lvl-1/` also needs to change, which leads to a further cascade of changes to the root of the storage.

There are three ways around this, as I see it:

* Do not include the `dcterms:modified` or `stat:mtime` triples (i.e. ignore this part of the spec)
* Don't worry about breaking how browser caching works (i.e. ignoring app/client needs)
* Put this data in a separate (auxiliary) resource that doesn't lead to a cascade of change.

I prefer the third option because it makes it possible to include this data without breaking how browsers interact with HTTP resources, but absent that approach, the only route forward I see (while staying in line with this spec) is to simply _not_ include the modified time of contained resources.

kjetilk: Yes, @acoburn , I have been arguing the same, an auxiliary resource solves all problems, it can represent the entirety of the resource (and it makes sense to add data for each representation too if needed), avoids cascading problems, can be subject to separate authorization, can be used to track any changes and should be at least as easy to implement. 

However, we haven't been able to find consensus around that for 0.9, so indeed, the result is that these data may be stale when conditional requests are used. I disagree with the conclusion, but as the RFC7232 is fairly vague at this point, I have chosen to accept it for 0.9, sincerely hoping we can revisit for 1.0.

csarven: >Because the Last-Modified header changes, the representation of `/lvl-1/lvl-2/` now also changes to reflect the new status of the contained resource.

While the representation data of `/lvl-1/lvl-2/` changes because of the `dcterms:modified` value of `/lvl-1/lvl-2/foo.ttl` , the `Last-Modified` header of `/lvl-1/lvl-2/` need not change - nor prohibited - as there were no changes to the containment triples ( #server-container-last-modified ).

A recommendation (or advisement) along the following lines may be necessary:

When the resource metadata of an existing contained resource changes, the server MUST send a weak `entity-tag` when responding to container’s request URI.

---

I want to mention again that the requirements introduced by this PR does not expect a container (`/lvl-1/lvl-2/`) to include resource metadata about itself (`</lvl-1/lvl-2/> dcterms:modified "2021-12-07T12:23:15Z"^^xsd:dateTime .`). (I don't care if that's not the point - it easily introduces complications/assumptions to the discussion that we are better off without.)   And again, if resource metadata about the container itself is desired - I may have missed the discussion that calls for it - the specification should needs to say so because both the generation and protection of those resource metadata needs to apply. On a related note:

When a server plans to include resource metadata about the container in the response of the same container, the server should determine the `Last-Modified` header value as a regular change to the representation data. (It can send a strong `entity-tag` as usual.)

acoburn: This does mean that apps/users will see stale data. From an impl perspective (given the choice), I would rather not provide the data than provide it in a way that confuses users.

I am really trying to be constructive here, but I don't see how this is implementable, given this text

csarven: As it stands, the Protocol uses the same requirement levels as in the RFC for Last-Modified and ETag headers. Servers might not include either Last-Modified or ETag headers, and so clients can't absolutely rely on them for being there.

That aside,

https://datatracker.ietf.org/doc/html/rfc7232#section-2.4 and https://datatracker.ietf.org/doc/html/rfc7232#section-6 suggests that headers with entity-tags will have higher precedence.

Server needs to enable the use of conditional requests so that client can eventually get a 200/304/412 or whatever.

Perhaps I'm missing something here but as mentioned in previous comment, I don't see a staleness issue around entity-tags.

If a client is only making decisions with Last-Modified/If-*-Since (when Last-Modified is provided by the server), then the representation metadata in the response will not help them to differentiate between a change to containment triples or any change to the representation. In this particular scenario, client has to make a GET request on the container.

We can certainly revisit that in 1.0.

For now, I suggest to add the following and go ahead with 0.9:

>When the resource metadata of an existing contained resource changes, the server MUST send a weak entity-tag when responding to container’s request URI.

acoburn: > When the resource metadata of an existing contained resource changes, the server MUST send a weak entity-tag when responding to container’s request URI.

I would like to just register that sending a weak ETag in this case does not address my concerns about browser caching at all. Any client sending a GET with an If-None-Match header will still receive a 304 (regardless of the weakness indicator) and therefore will present stale data to users

csarven: The suggestion is derived directly from the RFCs and is not required to be said but figured it can help.

There will be a different entity-tag for all changes to the container, so I don't quite see how a client using entity-tag x meanwhile the current representation's entity-tag is y will return a 304.

acoburn: If the entity tag changes from x to y, there is no issue, but that is orthogonal to the weakness indicator

csarven: Good. Then in what case would it not change from x to y?

acoburn: Any time the representation changes, the ETag changes. There is no argument about that. I am only responding to the proposed requirement to use weak ETags. And to be clear, I am not opposed to using weak ETags, but we can't claim  that weak ETags will solve this issue

csarven: The weak entity-tag is suggested because strong entity-tag wouldn't be correct when only the resource metadata changes. All meanwhile allowing the client to make conditional requests. Right at this second, I don't have a strong opinion on adding that requirement because I do think it comes directly from the RFC but it may be reasonable to mention.

That aside, I acknowledge that there is (always) room to discuss/work this out further but it doesn't need to be a blocker for 0.9. (We mark handful of stuff to be revisited for 1.0, and I don't see why it can't be done here as well.)

justinwb (approved):
✅ following discussion and agreements in 12/15 editor meeting

protocol.html
csarven: https://github.com/solid/specification/pull/352/commits/746520d614e1631334879e5f55b010665d707a1f

timbl (approved):

PR #357, opened 2021-11-28
Remove optional advisement relying on delegation of authority

csarven:
The URI owner may not have delegated authority to the server.


kjetilk (approved):
Indeed, since there is no corresponding server requirement.

RubenVerborgh (approved):

PR #358, opened 2021-12-13
Refer to WAC. Remove duplicate requirements

csarven:


matthieubosquet (approved):
Ah that feeling :grin: of removing duplicates! :tada:

kjetilk (approved):
Yes, this change makes sense. It simply references the WAC spec for both client and server requirements, which is good and also prepares the ground for later inclusion of several access control systems.

RubenVerborgh (approved):
Yes, great!

Issue #359, opened 2021-12-14
Complete Self-Review Questionnaire: Societal Impact


csarven:
Complete https://w3ctag.github.io/societal-impact-questionnaire/ and include as subsection.

PR #360, opened 2021-12-14
Add societal-impact-questionnaire stub

csarven:
In response to https://github.com/solid/specification/issues/359


timbl (approved):

kjetilk (approved):
Yes, we can have a stub in there, even though I think I would tend to answer all these questions with a 100 page dissertation if I were to answer them :-)

VirginiaBalseiro (approved):

jeff-zucker (approved):

Hindia (approved):

protocol.html
```diff
@@ -1168,6 +1169,47 @@ <h3 property="schema:name">Security and Privacy Review</h3>
...
+                    <dd about="#societal-impact-review-what-points-of-centralization-does-your-feature-bring-to-the-web-platform" datatype="rdf:HTML" property="schema:description">..</dd>
+
+                    <dt about="#societal-impact-review-to-what-extent-do-the-features-in-your-specification-result-in-increased-power-consumption-or-emissions" id="societal-impact-review-to-what-extent-do-the-features-in-your-specification-result-in-increased-power-consumption-or-emissions"><a href="https://w3ctag.github.io/societal-impact-questionnaire/#to-what-extent-do-the-features-in-your-specification-result-in-increased-power-consumption-or-emissions" rel="cito:repliesTo">To what extent do the features in your specification result in increased power consumption or emissions?</a></dt>
+                    <dd about="#societal-impact-review-to-what-extent-do-the-features-in-your-specification-result-in-increased-power-consumption-or-emissions" datatype="rdf:HTML" property="schema:description">..</dd>
```
TallTed, line 1203: @csarven -- This reads as `To what extent do the features in your specification result in increased power consumption or [unspecified] emissions?`  The most obvious, `power emissions`, would seem to be a positive.  Emissions of H2O or O2 would also likely be fine, so it's not `any emissions`.  Therefore, assuming greenhouse gases are the actual concern, this should change from `power consumption or emissions` to **`power consumption or greenhouse gas emissions`**.

bourgeoa:
I'm not sure to understand what this questionnaire has to do here ?
And not happy with the idea to see this in solid-protocol.

jeff-zucker:
@bourgeoa - one of the main goals of Solid and the web itself is "The web should be a platform that helps people and provides a net positive social benefit."   So it behooves this project, like all projects to think about and clearly state what the project's impact will be on society.  Part of the reason we need Solid now is that we didn't fully think through all of the societal impacts of the web the first time around.  So I do believe this belongs in the protocol,

jeff-zucker:
The quote above is from https://www.w3.org/2001/tag/doc/ethical-web-principles/

bourgeoa:
Ethic concerns people and not a protocol.
I went to the text and find that net social benefit cannot be defined other than with a political background.
Societal impact background is not better.
As to energy consumption this can also not be considered in a protocol either. This a much too complex study depending on a system approach of implementation.

jeff-zucker:
I disagree, the specification has a social impact apart from the social impact of implementations.  So there needs to be an impact statement for the specification itself.  

As for how "net social benefit" is defined - that is a moral question and only secondarily a political question.  The days when software can pretend to have no moral impact are over.  Also, I am working on an ontology of social benefit.

PR #361, opened 2021-12-14
Create solid-webid-profiles.md

jeff-zucker:
A repo for the new webid-profiles spec.

### DIFF
--- a/profile/solid-webid-profiles.md
+++ b/profile/solid-webid-profiles.md
@@ -1,0 +1,295 @@
+# Solid WebID Profiles Spec
+
+**Note:** This is currently the old document.  We can use it as a basis for creating a new one.
+
+## Table of Contents
+
+* [Overview](#overview)
+* [Profile Representation Formats](#profile-representation-formats)
+* [Required Profile Information](#required-profile-information)
+* [Minimum Recommended Profile
+      Information](#minimum-recommended-profile-information)
+    * [Recommendation for User Names in
+      Profiles](#recommendation-for-user-names-in-profiles)
+* [Public and Private Profiles](#public-and-private-profiles)
+    * [Extended Profile](#extended-profile)
+* [Public Key Certificates](#public-key-certificates)
+* [Account Resource Discovery](#account-resource-discovery)
+    * [Storage Discovery](#storage-discovery)
+    * [Inbox Discovery](#inbox-discovery)
+    * [Type Registry Index Discovery](#type-registry-index-discovery)
+
+## Overview
+
+Solid uses WebID Profile Documents for management of user identity and security
+credentials (such as public keys), and user preferences discovery.
+
+## Profile Representation Formats
+
+From the
+[WebID Profile](https://www.w3.org/2005/Incubator/webid/spec/identity/#publishing-the-webid-profile-document)
+spec:
+
+> A WebID Profile Document is an RDF Web resource that MUST be available as
+> [text/turtle](http://www.w3.org/TR/turtle/), but MAY be available in other RDF
+> serialization formats (such as JSON-LD or
+> [HTML+RDFa](http://www.w3.org/TR/rdfa-core/)) if requested through content
+> negotiation.
+
+## Required Profile Information
+
+There are only 3 statements required for a valid (though not very useful)
+WebID Profile Document:
+
+1. Identifying the document as a `foaf:PersonalProfileDocument` instance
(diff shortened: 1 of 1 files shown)


profile/solid-webid-profiles.md
```diff
@@ -0,0 +1,295 @@
...
+
+```ttl
+@prefix foaf: <http://xmlns.com/foaf/0.1/>.
+<https://alice.databox.com/profile/card>
+    a foaf:PersonalProfileDocument ;
+    foaf:primaryTopic <#me> .
+
+<#me>
+    a foaf:Person .
+```
```
matthieubosquet, line 58: This is incorrect.

The only requirements for a WebID Profile Document are the following:

1. It must be available as `text/turtle`
    >  A WebID Profile Document is a Web resource that MUST be available as text/turtle...
    > -- https://www.w3.org/2005/Incubator/webid/spec/identity/#introduction
2. It must conform to one of the [Cool URIs solution for identifying real world objects](https://www.w3.org/TR/cooluris/#solutions) 
    > For WebIDs with fragment identifiers (e.g. #me), the URI without the fragment denotes the Profile Document. For WebIDs without fragment identifiers an HTTP request on the WebID MUST return a 303 with a Location header URI referring to the Profile Document.
    > -- https://www.w3.org/2005/Incubator/webid/spec/identity/#terminology
3. The WebID referent must control the Profile Document
    > The WebID HTTP URI must be one that dereferences to a document the user controls.
    > -- https://www.w3.org/2005/Incubator/webid/spec/identity/#the-webid-http-uri


If we want to create a recommendation as to the shape of a `foaf:PersonalProfileDocument` linked to a WebID, then we should be clear about that intent.

In other words, a WebID Profile Document may or not be a foaf:PersonalProfileDocument, which is a good thing since a WebID is extremely useful as a unique identifier for decentralised authorization and authentication. Optimising for authentication and authorization is not necessarily compatible with presenting someone's person in different public and private or semi private documents. Solid is a great tool to do such things but not if we confuse FOAF Personal Profile Documents and WebID Profile Documents.

matthieubosquet (changes requested):
My comment applies to the whole document really.

I would be happy to discuss or propose a more adequate text if others find it useful.

profile/solid-webid-profiles.md
matthieubosquet: PS: The logical and interesting conclusion one can reach from the above is that an empty document is a perfectly valid WebID Profile Document.

Note that the one kind of statement that is required to make a WebID profile document useful in Solid is one stipulating a trusted OIDC issuer:

```turtle
ex:ahmed solid:oidcIssuer ex:issuer1 .
```

Note also that if the Solid-OIDC spec was slightly ammended to allow either/or header/representation instead of a MUST on the RDF representation bearing the trusted issuer, then we could get away with simply a header, rendering WebID Profile Document representation requirements void (which might not be a bad thing). See also: https://solid.github.io/solid-oidc/#webid-issuer-discovery.

matthieubosquet: @TallTed, as far as I understand it is the most authoritative document on the subject of WebIDs, is there anything superseding it?

csarven (changes requested):
I think this PR is not necessary - solid-webid-profile.md need not be introduced to this repo or even reviewed. (For those reviewing, there is a Note at the top of the document implying that it doesn't need to be reviewed.)

Use protocol.html's structure/semantics as template for the profile spec. PR that HTML.

We can have a dedicated repository for the profile specification (as we tend to do with other specs) e.g., solid/{shortname} where it will be potentially published at /TR/{shortname}. It can be an Editor's Draft in that space and snapshots of it intended to be published under /TR/ can be processed through this repo ( solid/specification ).

The editors of that spec can somewhat work quietly in the new repo, in public, for the time being. Just don't draw additional attention to it (by linking to it) until you have a version PR'd here. That repo can be the home for profile work. We can gradually bring back the identity panel and that can be one of its work items.

The scope of that repo can be worked out there... but keeping the initial recommendation/draft focus on WebID. So, something the editors of that spec can work out is the shortname and then we can follow up with the repo creation etc.. I suggest solid-profiles or webid-profile... depending on your aims. This distinction (re scope) is something we can also pick up in an editors meeting in parallel.

csarven:
okie dokie. let's continue work at https://github.com/solid/webid-profiles/

PR #362, opened 2021-12-15
Auxiliary resource for container metadata as alternative

kjetilk:
Given @acoburn 's objections in https://github.com/solid/specification/pull/352/files#r764244049 and onwards, I decided to write down an alternative to #352 that introduces an auxiliary resource (hereafter known colloquially as a stat resource), as proposed by @acoburn , myself and others. We can hold these two PRs up against each other.

I believe this PR solves the following problems:

* It ensures that the entire resource is taken into consideration when computing last modified, and thus avoids the staleness problem of #352.
* Since the data goes into an auxiliary resource, the container itself is not modified when a child is modified, and thus avoids the cascade-to-root problem.
* I allow the aux resource to have its own access control, which should alleviate the concern that metadata may be exploited to find vulnerabilities somewhat, as it can be used to restrict access to these data further than the container permissions indicate.

The drawbacks of this approach is, AFAICS only that NSS would have to be modified slightly and thus clients currently using these data would need to `GET` another resource. I believe this is a small cost that should be taken as soon as possible, but it is also possible that this proposal should be considered for 1.0 and not 0.9.

I have in the present proposal required that this stat resource exists, and I have registered opposition to that. I believe that having this aux resource would make a fine implementation detail too, you'd record any changes on the aux resource on the backend, and so it should ease implementation too. If it is still insisted that this should be optional if it is too expensive, then I suggest that we make the discovery sentence a SHOULD, as if there is no discovery, there's no stat resource. 

I think that further work on this should include making data for representations explicit, things like size and media type shouldn't be on the resource, but on the representations.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -585,7 +585,8 @@ left:4.5em;
                   <ul>
                     <li><a href="#auxiliary-resources-web-access-control">Web Access Control</a></li>
                     <li><a href="#auxiliary-resources-description-resource">Resource Description</a></li>
-                  </ul>
+                    <li><a href="#auxiliary-resources-stat-resource">Server Metadata Description Resource</a></li>
+		  </ul>
 
                   <p><span about="" id="client-link-auxiliary-type" rel="spec:requirement" resource="#client-link-auxiliary-type">Clients can discover auxiliary resources associated with a subject resource by making an HTTP <code>HEAD</code> or <code>GET</code> request on the target URL, and checking the HTTP <code>Link</code> header with the <code>rel</code> parameter [<cite><a class="bibref" href="#bib-rfc8288">RFC8288</a></cite>].</span></p>
 
@@ -607,6 +608,11 @@ left:4.5em;
                         <td><a href="#auxiliary-resources-description-resource">Description Resource</a></td>
                         <td><code>describedby</code></td>
                         <td>[<cite><a class="bibref" href="#bib-ldp">LDP</a></cite>]</td>
+                      </tr>
+		      <tr>
+                        <td><a href="#auxiliary-resources-stat-resource">Server Metadata Description Resource</a></td>
+                        <td><code>http://www.w3.org/ns/solid/terms#statResource</code></td>
+                        <td></td>
                       </tr>
                     </tbody>
                     <tfoot>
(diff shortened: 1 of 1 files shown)


acoburn (approved):
Thank you @kjetilk, this is great.

My only suggestion (which is definitely in the bikeshedding category, so feel free to ignore) is to call this a Container Metadata Resource. To me, a Server Metadata Resource describes the server as a whole, rather than being server-managed metadata about a particular container.

kjetilk:
Thank you very much @acoburn . I am happy to change the name. I wasn't sure what to call it, but decided to include "server" to allude to that it is server managed, but that is also made clear in the text, so I am happy to change it.

protocol.html
```diff
@@ -648,6 +654,34 @@ <h4 property="schema:name">Description Resource</h4>
                       <p><span about="" id="client-link-describes" rel="spec:requirement" resource="#client-link-describes">Clients can discover resources that are described by description resources by making an HTTP <code>HEAD</code> or <code>GET</code> request on the target URL, and checking the HTTP <code>Link</code> header with a <code>rel</code> value of <code>describes</code> (inverse of the <code>describedby</code> relation) [<cite><a class="bibref" href="#bib-rfc6892">RFC6892</a></cite>].</span></p>
                     </div>
                   </section>
+
+                  <section id="auxiliary-resources-stat-resource" inlist="" rel="schema:hasPart" resource="#auxiliary-resources-stat-resource">
+                    <h4 property="schema:name">Container Metadata Resource</h4>
+                    <div datatype="rdf:HTML" property="schema:description">
+                      <p>
+			An auxiliary resource of type <em>Container Metadata Resource</em> provides server-maintained metadata about contained resources.
```
TallTed, line 662:
```suggestion
			An auxiliary resource of type <em>Container Metadata Resource</em> provides server-maintained metadata about contained resources.
```
Shouldn't the type be a URI (presumably `#auxiliary-resources-stat-resource` with a base I'm not quickly seeing), rather than a literal?

kjetilk:
The Solid Editors meeting decided to merge #352 today, but indicate that the mechanism may change, so we may return to something like this later.

PR #365, opened 2021-12-17
Use storage or server in place of data pod

csarven:
Removes use of the term data pod in requirements in favour of server or storage. Leaves the the term "data pod" as is for now. It should be revisited for 1.0.


protocol.html
```diff
@@ -469,13 +469,13 @@ <h2 property="schema:name">Hypertext Transfer Protocol</h2>
...
+                  <p><span about="" id="server-authentication" rel="spec:requirement" resource="#server-authentication"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> conform to <cite>HTTP/1.1 Authentication</cite> [<cite><a class="bibref" href="#bib-rfc7235">RFC7235</a></cite>].</span></span> <span about="" id="server-unauthenticated" rel="spec:requirement" resource="#server-unauthenticated"><span property="spec:statement">When a client does not provide valid credentials when requesting a resource that requires it (see <a href="#webid">WebID</a>), the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> send a response with a <code>401</code> status code (unless <code>404</code> is preferred for security reasons).</span></span></p>
```
matthieubosquet, line 478:
```suggestion
                  <p><span about="" id="server-authentication" rel="spec:requirement" resource="#server-authentication"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> conform to <cite>HTTP/1.1 Authentication</cite> [<cite><a class="bibref" href="#bib-rfc7235">RFC7235</a></cite>].</span></span> <span about="" id="server-unauthenticated" rel="spec:requirement" resource="#server-unauthenticated"><span property="spec:statement">When a client does not provide valid credentials when requesting a resource that requires it (see <a href="#webid">WebID</a>), <span rel="spec:requirementSubject" resource="spec:Server">servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> send a response with a <code>401</code> status code (unless <code>404</code> is preferred for security reasons).</span></span></p>
```

Use the pluralized form for the term "Server" for consistency.

matthieubosquet (approved):
I like the simplification and distinction between server and storage.

I am wondering if the term "Solid Server" would be better than simply "Server". I know it is derived from context and I'm not sure what it means for the use of spec:Server (do we need to use solid:Server subclassOf spec:Server instead?) but it seems potentially nice to be more specific.

elf-pavlik (approved):
That sounds like a good change 👍 
I didn't read the full diff due to ridiculous line length.

kjetilk (approved):
Yes, I think it is good to use the term pod as an informal term and storage as the corresponding formal term.

RubenVerborgh (approved):
All good!

justinwb (approved):
Much better ✅

acoburn (approved):
:+1:

Issue #356, opened 2021-11-26
Servers responding to request targets using the asterisk-form or the origin-form


csarven:
While the Solid Protocol requires servers to conform to RFC 7230 and RFC 7231, there may exist Solid server instances that cannot receive requests using the `asterisk-form` or a particular `absolute-path` (and path segments in lower hierarchy) as target.

It is important to consider whether any of the following requirements should be included in the Solid Protocol:

* Servers MUST respond to request targets using the `asterisk-form` (RFC 7230).

* Servers MUST respond to request targets using the `origin-form` (RFC 7230).

The first requirement is relevant for communication options: `OPTIONS *`. If servers cannot respond to requests using the `asterisk-form` as target, the following requirement either needs to be removed:

>Servers MAY include the HTTP Accept-Patch, Accept-Post and Accept-Put headers in the response of a OPTIONS * request.

or the requirement for those Accept- headers should be in the response of `OPTIONS` for authorized requests.

The second requirement is relevant when responding to specific resources, and perhaps most importantly whether server can respond to request target: `/`.

acoburn:
It may be worth noting that any Solid server running behind an nginx proxy will have difficulty responding to `OPTIONS *` requests.

https://trac.nginx.org/nginx/ticket/1882

In my experience it simply doesn't work. It would be unfortunate if a specification requirement made it impossible to use one of the most widely deployed reverse proxy servers on the web, especially one that is built into most kubernetes deployments.

I have no concerns about including those `Allow-*` headers in authenticated `OPTIONS /resource-path` responses. I also have no concerns about `OPTIONS /` requests.

kjetilk:
> It may be worth noting that any Solid server running behind an nginx proxy will have difficulty responding to `OPTIONS *` requests

Oh no! Indeed, nginx is an extremely important piece of software, so if they decided this, we can't use that. :-(

RubenVerborgh:
> Oh no! Indeed, nginx is an extremely important piece of software, so if they decided this, we can't use that. :-(

We might need this in an overall checklist somewhere; bit me at https://github.com/solid/specification/issues/340 as well.

acoburn:
Related to the `origin-form` request (e.g., `GET / HTTP/1.1`), the "get-off-my-lawn" [RFC 8820, section 2.3](https://datatracker.ietf.org/doc/html/rfc8820#section-2.3) is relevant.

The key point made there is that a specification should not require or expect a certain path for applications. 

[...]

PR #369, opened 2021-12-27
Remove server-options-asterisk-accept-headers. Add OPTIONS to server-accept-headers

csarven:
Related issue: https://github.com/solid/specification/issues/356

Not all servers will be configured to receive/respond to `OPTIONS *` (`asterisk-form`). Clients can't rely on all servers (be instructed by their respective URI owners) allowing the request.

Checking for `Accept-*` header options should be possible with `OPTIONS` targeting a resource (`origin-form` or `absolute-form`).

### DIFF
--- a/ED/protocol.html
+++ b/ED/protocol.html
@@ -131,7 +131,7 @@ content: "";
     <main>
       <article about="" typeof="schema:Article doap:Specification">
         <h1 property="schema:name">Solid Protocol</h1>
-        <h2>Editor’s Draft, 2021-12-22</h2>
+        <h2>Editor’s Draft, 2021-12-27</h2>
 
         <details open="">
           <summary>More details about this document</summary>
@@ -178,7 +178,7 @@ content: "";
 
           <dl id="document-modified">
             <dt>Modified</dt>
-            <dd><time content="2021-12-22T00:00:00Z" datatype="xsd:dateTime" datetime="2021-12-22T00:00:00Z" property="schema:dateModified">2021-12-22</time></dd>
+            <dd><time content="2021-12-27T00:00:00Z" datatype="xsd:dateTime" datetime="2021-12-27T00:00:00Z" property="schema:dateModified">2021-12-27</time></dd>
           </dl>
 
           <dl id="document-repository">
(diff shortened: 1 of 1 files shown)


RubenVerborgh (approved):

kjetilk (approved):
I think `OPTIONS *` is the most promising candidate for a sane, general discovery mechanism for HATEOAS, but given that nginx doesn't support it, it is indeed not possible to rely on :sob:

TallTed:
*sighs* Might be worth [logging an issue on `nginx`](http://trac.nginx.org/nginx?_ga=2.245932484.1583906896.1640630753-250270040.1640630753), as `OPTIONS *` is becoming ever more important, as web servers continue to deliver more than basic web services, but not always the same set of extensions, features, etc.

acoburn:
This _was_ raised with [nginx](https://trac.nginx.org/nginx/ticket/1882) and subsequently closed as "won't fix".

TallTed:
Thanks for finding that, @acoburn.  I don't see a way to comment on a closed issue there; might be because I haven't logged in to their `trac` system, or because I'm not a contributor to their project.  I think that providing them with some new (to them, anyway) information -- i.e., that while `nginx` doesn't expect `OPTIONS *` and so won't handle it directly, many are deploying `nginx` as a proxy to load balance and otherwise protect actual back-ends, which *may* expect `OPTIONS *` as they support more (and more complex) services, and users (or their apps/agents) *may* have reason to ask what the back-end server supports, distinctly from what it supports for any specific resource -- and relaying such HTTP/S API calls to the back-end server in such proxy configurations would be worthwhile.

acoburn (approved):

acoburn:
@TallTed even if nginx were to support such an operation, it is currently _not possible_ to issue such a request via the JavaScript `fetch` function in any browser. As such, the feature would still not be usable for most of Solid. Either way, removing this requirement from the Solid protocol is a good direction.

csarven:
Server and client implementations in the wild is an important signal but it is perhaps worth repeating that the primary reason for this PR is about removing the assumption that the owner of a URI delegates authority to a server to receive/respond to request targets in asterisk-form.

Even if software (the servers, proxies, clients, SDKs, applications, .. at any level..) have the capability to request and respond to `OPTIONS *` (eg from RFC 7230):

>OPTIONS * HTTP/1.1
>Host: www.example.org:8001

a Solid server instance may never have been configured to even receive such request.

As it stands, the Protocol requires that servers must make a storage (root container), at any URI path ending with a slash, but nothing further. Which is why "[c]lients can't rely on all servers (be instructed by their respective URI owners) allowing the request."

The same reasoning was applied to removing any expectation that a Solid server can receive/respond to requests at root path `/` ( https://github.com/solid/specification/pull/357 ) or any specific URI for that matter (e.g., `.well-known/solid`, `/profile/card#me`, ...)

---

Yes, the Protocol can be changed to require all of that but until there is a need to force all server instances to receive/respond to specific requests, we should continue to tighten the spec.

csarven:
Double-checked: neither fetch and xhr specs describe how `OPTIONS *` can be requested. Also did quick tests from dev console in Firefox Nightly and Chromium. No go. Asked in whatwg in case I missed a spec: https://matrixlogs.bakkot.com/WHATWG/2022-01-03#L11-L12 (FWIW):

>11:26 | \<csarven> | Is it possible to request target in asterisk-form (OPTIONS *) with fetch, xhr (or another)? Any pointers on history/discussion would be great.
>12:35 | \<annevk> | csarven: it's not possible, don't recall discussion

---

`OPTIONS *` is neat but doesn't appear to be usable for the Solid ecosystem at this time.

PR #371, opened 2022-01-05
Require Description Resources to be supported

kjetilk:
Currently, there is no server requirement to support `describedby` resources, only a description about how clients may discover them. I believe this to be a simple omission, it has been in Solid since the dawn of ages, and even though we want to [generalize aux resources](https://github.com/solid/specification/issues?q=is%3Aissue+is%3Aopen+label%3A%22topic%3A+auxiliary+resources%22+), this basic functionality should be required in 0.9.x.

There are also tests https://github.com/solid/specification-tests/pull/65 in this area, and I believe it is appropriate to make this a requirement rather than introduce more variability.

### DIFF
--- a/ED/protocol.html
+++ b/ED/protocol.html
@@ -753,7 +753,8 @@ content: "";
                   <section id="auxiliary-resources-description-resource" inlist="" rel="schema:hasPart" resource="#auxiliary-resources-description-resource">
                     <h4 property="schema:name">Description Resource</h4>
                     <div datatype="rdf:HTML" property="schema:description">
-                      <p>An auxiliary resource of type <em>Description Resource</em> provides a description of a subject resource ([<cite><a class="bibref" href="#bib-ldp">LDP</a></cite>]).</p>
+                      <p>An auxiliary resource of type <em>Description Resource</em> provides a description of a subject resource ([<cite><a class="bibref" href="#bib-ldp">LDP</a></cite>]).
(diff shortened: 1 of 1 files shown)


csarven (changes requested):
I propose PR #372 instead of this PR. It clarifies the actual advertisement of the link relations for auxiliary resources described by the Solid Protocol. It also complements/accompanies #client-link-auxiliary-type.

ED/protocol.html
```diff
@@ -753,7 +753,8 @@ <h4 property="schema:name">Web Access Control</h4>
                   <section id="auxiliary-resources-description-resource" inlist="" rel="schema:hasPart" resource="#auxiliary-resources-description-resource">
                     <h4 property="schema:name">Description Resource</h4>
                     <div datatype="rdf:HTML" property="schema:description">
-                      <p>An auxiliary resource of type <em>Description Resource</em> provides a description of a subject resource ([<cite><a class="bibref" href="#bib-ldp">LDP</a></cite>]).</p>
+                      <p>An auxiliary resource of type <em>Description Resource</em> provides a description of a subject resource ([<cite><a class="bibref" href="#bib-ldp">LDP</a></cite>]).
+			<span about="" id="server-description-required" rel="spec:requirement" resource="#server-description-required"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support this auxiliary resource type.</span></span></p>
```
RubenVerborgh, line 757: And how is it identified? What does supporting it mean?

kjetilk: So, this is language pointing at the more generic language in the spec, which also needs work. The `rel` value is defined in a table, the protocol specific things are there too, the lifecycle assumptions also. I recognize it is in many contexts, including this, a bit problematic to just say "support", but I also tend to think that the tests is a venue to provide authoritative interpretations.

This is mostly to fix the immediate omission that `describedby` wasn't required in 0.9.

PR #374, opened 2022-01-24
Make the requirement only to support all

kjetilk:
As mentioned in #372, the language is still too fluffy for me, so I suggest just making the requirement to support them all. It doesn't clarify, but then at least it is clear what we can test.

I couldn't see how I just make suggestions, so did it this way.


csarven:
Updated re https://github.com/solid/specification/pull/372 . Let's move on.

PR #378, opened 2022-02-08
server-allow-methods in successful responses

csarven:
This PR is a continuation of #353 with the difference being that the Allow header is included in successful responses.

Merging this PR closes #353 - There was no point in bringing that PR up to speed and changing it to modify ED/protocol.html.


ED/protocol.html
```diff
@@ -795,11 +795,9 @@ <h3 property="schema:name">Reading Resources</h3>
...
+                  <p><span about="" id="server-accept-headers" rel="spec:requirement" resource="#server-accept-headers"><span property="spec:statement">When responding to authorized requests, <span rel="spec:requirementSubject" resource="spec:Server">servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> indicate supported media types in the HTTP <code>Accept-Patch</code> [<cite><a class="bibref" href="#bib-rfc5789">RFC5789</a></cite>], <code>Accept-Post</code> [<cite><a class="bibref" href="#bib-ldp">LDP</a></cite>] and <code>Accept-Put</code> [<cite><a href="#accept-put">The Accept-Put Response Header</a></cite>] response headers that correspond to acceptable HTTP methods listed in <code>Allow</code> header value in response to HTTP <code>GET</code>, <code>HEAD</code> and <code>OPTIONS</code> requests.</span></span></p>
```
acoburn, line 800: Do you feel that this text is sufficiently clear that CORS preflight requests (which contain no auth headers) would be exempt from the requirement to provide these `Allow-*` headers?

csarven: (I think you mean Accept-* headers) I believe / hope so. It was intended to be exempt since no auth will take place for CORS-preflight. If this is unclear to implementers, happy to revise.

kjetilk (approved):
Alright, sounds good!

ED/protocol.html
```diff
@@ -795,11 +795,9 @@ <h3 property="schema:name">Reading Resources</h3>
...
                   <p><span about="" id="server-safe-methods" rel="spec:requirement" resource="#server-safe-methods"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support the HTTP <code>GET</code>, <code>HEAD</code> and <code>OPTIONS</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] for clients to read resources or to determine communication options.</span></span> [<a href="https://github.com/solid/specification/issues/39#issuecomment-538017667" rel="cito:citesAsSourceDocument">Source</a>]</p>
 
-                  <p>When responding to authorized requests:</p>
+                  <p><span about="" id="server-allow-methods" rel="spec:requirement" resource="#server-allow-methods"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> indicate the HTTP methods supported by the target resource by generating an <code>Allow</code> header field in successful responses.</span></span></p>
```
RubenVerborgh, line 798: But does that depend on auth? So 401/403 replies are ok?

RubenVerborgh (approved):

ED/protocol.html
csarven: "successful responses" as in 2xx. 401/403 shouldn't include Allow. Essentially 2xx and 405 (as per RFC) is recommended.

csarven: Yes. If POST (and other methods) would be a "valid request" supported by the target resource, include it in Allow. See also last three comments at https://github.com/solid/specification/pull/353 .

csarven:
Reading between the lines (RFC 7231):

>In general, it is assumed that the origin
   server will only allow DELETE on resources for which it has a
   prescribed mechanism for accomplishing the deletion.
>[..]
>For example, a resource that was
   previously created using a PUT request, or identified via the
   Location header field after a 201 (Created) response to a POST
   request, might allow a corresponding DELETE request to undo those
   actions.

PR #391, opened 2022-04-04
Add TR/2022/notifications-protocol-20220401

csarven:
Editor's Draft: https://solid.github.io/notifications/protocol

The https://github.com/solid/notifications-panel considers this document to be Version 0.1.0 (equivalent to W3C FPWD).


kjetilk (commented):
My comment is primarily around the discussion on RESTfulness. I'm not sure why that has entered, nor its relevance. Both the Solid Protocol and the Notifications Protocol, we have chosen to clarify certain stuff out of band, rather than inbound, since we're writing specs about it. That's fine, you have to do that sometimes. However, they both define their protocols in terms of the usual connectors, the resources and resource representations, etc. 

AFAICT, they can both be RESTful, but they can also not be, if a certain deployment chooses to just drive the interaction with out-of-band information, or a certain subscription type having a stateful protocol, etc.

Or is it something that I'm missing that prevents the notifications protocol from being RESTful?

Otherwise, I think it makes sense to remove the discussion, it doesn't seem to be important to the protocol.

CxRes:
It is my understanding that, a notifications protocol, at least as notifications technologies currently exist, cannot be RESTful. REST was designed to explicitly allow clients to “follow their nose” (Roy Fielding's language) when navigating information, with the requirement that servers do not persist any information about the client (Statelessness), allowing the web to have a number of nice network properties.

Our goal is to allow clients to follow their nose not just spatially but temporally as well. However, before the advent of notifications, the only way to do so was polling, which is extremely inefficient. By allowing servers to just remember who the clients are, the latter can be informed of temporal changes to a resource. But our system is no longer RESTful as the server is not stateless!

Now, the question that arises is how far do we want to stray from the REST ideal? Each feature requires the server to remember more information about the client (beyond just knowing which client is connected), as a result of which we lose even more of the advantageous properties that REST allowed us to have. For example, with rate limiting as a feature, servers, instead of having to broadcast changes to all subscribed clients, have to tailor messages to each client, with the result that use of intermediaries (a REST benefit), even in theory, is not possible. I believe that each feature of the Solid notifications protocol should be weighed against the departure from REST, even if it is not relevant to the current crop of notification technologies. 

This also ties into the issue of simplicity that TimBL has been insistent upon in one of the meeting. Features add complexity, that for client developers is more work. I want my program to be able to "just connect" and "start receiving notifications" without worrying about extra features. This can be of immediate benefit to both client and server developers. On the flip side, I can see that a protocol must make room for features (or the possibility for features) that can be made available through it.

I have raised this concern in a chat a long time ago. I can see that this is not an immediate concern with the use of, say, websockets where we have bi-directional persistent connections between a server and clients. However, protocols are long-term concerns which, in all probability, will outlive the technology stack on which they are implemented. Therefore, there must be a principled basis (beyond utility) for what goes in!

RubenVerborgh:
REST is not a goal but a means to an end; not is it an ideal.
It is indeed an architectural style that, when followed, gives a system certain properties.

Depending on what properties we want, other architectures can be desirable for other parts of systems. As far as notifications are concerned, my team has performed evaluations in the past, and similar evaluations can be performed for the notifications protocol to make sure it shows the desired behavior.

kjetilk:
Lets not overanalyze this, the spec is fine without the REST stuff, lets just say that.

In the interest of further understanding, note that [REST discusses](https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm#sec_5_2)  these things in terms of *components*, and it is important to note that "a component may include both client and server connectors."

As I understand the proposal, it will actually be up to each subscription type to fully define the components and, it is therefore not precise to say that something is a server or a client, they may play the role of both, or either, depending on the subscription type extension. Some subscription types may be RESTful, some may not, and that's all fine, no problem at all.

In the case of a subscription type that maintains a persistent connection between the subscriber and an notification source, no server state is required, so it is a good chance that will be RESTful.

In the case of a subscription type that needs to record the subscriber, that could be implemented with a component that has both a client and server connector. The server connector records a subscriber, but it does not persist it, it merely passes it to the client connector of the component, and so the visibility and scalability is preserved, and it isn't terrible for reliability either. Then, the client connector publishes a notification as needed. This would also not violate REST, as far as I can see.

Given that notifications tend to be rather ephemeral, I would think that the scalability quality is quite easily satisfied and so subscription types are unlikely to go out of their way to design for statelessness, and so REST isn't an architectural style they are likely to aim for, I just wanted to say that it is also not something that doesn't seem to be unachievable like the intro seemed to imply.

CxRes:
I am in concurrence that this spec is fine for a FPWD! 

@RubenVerborgh 

1. I think we are much closer in agreement than your comment suggests. I agree that REST is not "the" or even "a" a goal/ideal. However, we are using REST as the de facto means of communication with the Solid Pod. IMHO, the desire for notifications architecture being close to REST stems from the desire to benefit from the network properties that REST ensures. I am not contesting the choice of a completely different architectural style for different components, with the caveat that this brings additional complexity, which must be balanced against the value it provides.

2. What I think is more important is that, irrespective of the choice of architecture or features, the reasons for/principles underlying those decisions need to be documented, not in the standard itself but in some adjacent document. This becomes important for a component like notifications, where there is little canonical guidance. And it is very helpful for people like me, who are not from the field.

3. I would love to see your evaluation and learn from it.

@kjetilk 

As for the just academic discussion about REST, I am unfortunately not able to follow your reasoning. Most probably, I even disagree with it. If you desire to continue, we can take this discussion out of band, say, on Gitter, as it is not relevant to the merging of this PR.

PR #392, opened 2022-04-04
Add TR/2022/websocket-subscription-2021-20220401

csarven:
Editor's Draft: https://solid.github.io/notifications/websocket-subscription-2021

The https://github.com/solid/notifications-panel considers this document to be Version 1.0.0.


websocket-subscription-2021.html
```diff
@@ -0,0 +1,498 @@
...
+
+                    <figcaption property="schema:name">Response to the <code>POST</code> request including <code>type</code> and <code>source</code>.</figcaption>
+                  </figure>
+
+                  <p>In JavaScript, a client can use the data in the response to establish a connection to the WebSocket hypermedia API.</p>
+
+                  <figure id="websocket-new" class="example listing" rel="schema:hasPart" resource="#websocket-new">
+                    <p class="example-h"><span>Example</span>: Establishing a new WebSocket connection.</p>
+
+                    <pre about="#websocket-new" property="schema:description" typeof="fabio:Script"><code>const ws = new WebSocket(url, type);</code></pre>
```
kjetilk, line 439: If the `url` in the code is understood as the `source` from the response above, which is how I understand it, then perhaps the `source` should be used in the code too?

Or, if my understanding is wrong, then perhaps it should be explained?

kjetilk (approved):
I have just one minor inline comment, other than that this seems to address the concerns that were raised in a manner consistent with the larger notifications document, and so may be published RSN.

csarven:
I've updated the snapshot based on the the review here. Created a new PR: https://github.com/solid/specification/pull/405

PR #407, opened 2022-05-12
Add version scheme to Contributing Guide

csarven:


kjetilk (approved):
I would prefer if panels could use version <1 for their initial releases of documents that have not yet been submitted to the Solid Editors.

Also, I note that the Solid Protocol itself is not ready to follow this requirement.

But in the interest of this not becoming a permathread, I approve.

justinwb (approved):
Considered whether or not it's worth specifying that early / in-progress can use 0.X.X, but I think it's implicit with semver so I'm fine with it as it is.

csarven:
I think publications at spaces other than https://solidproject.org/TR/ can use what they see fit, but I do agree that they might want to consider using <1 to make things clear. Currently they use ED. I'm open to adding a guideline along those lines if that's helpful - yay/nay?

timbl (approved):

RubenVerborgh (approved):
Happy with the proposal; just want to raise that the specific notation `Version 1.0.0-WD.20221231` might cause confusion. It might be read as "version 1.0.0 with date xxx" or something. Perhaps DRAFT can be labeled more explicitly, but that's just usability on the document.

elf-pavlik:
During the Solid Notification call, @csarven mentioned that the next release of Solid Protocol will add new features including a new Solid Notifications Protocol. As I understand semver, that next release should be `0.10.0` not `0.9.1`.

I believe we shouldn't use `pre-release` until we consider the spec *CR*/*PR* equivalent.

PR #409, opened 2022-05-23
Add requirement for Solid Notification Protocol

acoburn:
This adjusts the Solid Protocol ED to require the new [Solid Notification Protocol](https://solidproject.org/TR/notifications-protocol).

As part of this change, all references to the old WebSocket mechanism move to non-normative statements, encouraging existing implementations to continue to support backwards compatibility, as appropriate.

This change also modifies the existing "Notifications" section title to "Linked Data Notifications" in order to avoid confusion between sections 6 and 7.

PR #448, opened 2022-08-07
ED/protocol.html: remove doubled dot at the end of a sentence

almereyda:
There is a dot doubled at the end of one sentence.

This commit removes one of the redundant dots.


csarven:
Thank you for catching this and creating the PR. I believe https://github.com/solid/specification/pull/409/ supersedes this PR in that the correction as proposed in this PR is no longer necessary.

almereyda:
Well then. I didn't see and/or review it. Many thanks for the heads up, though.

Issue #355, opened 2021-11-25
Server Description


csarven:
<span id="background">Background</span>: there are a wide-range of use cases that needs to get a hold of *authoritative information* about the server. To date, there is no uniform way for agents and applications to find all information about a server.

<span id="general-use-case">General use case</span>:
Find authoritative information about a server.

<span id="use-cases">Use cases</span>:
* Guinan wants to read the **description** of the server and understand its purpose.
* Janeway wants to find the server's **policies** to review its data privacy and pledge for resource persistence.
* Uhura wants to check the server's **communication options** in order to use a preferred notification protocol.
* Seven of Nine wants to determine a suitable location to keep astrometric data from the server's **available storages**.
* Picard wants to find the server's **contact information** to request a diplomatic meeting about security audits.
* Torres wants her devices to use server's URI **templates** when naming particular kinds of resources.

There are many others.. which may be low-priority or already have existing standards, but could be captured in this data nevertheless, e.g., sitemap, Web syndication, robots control.

<span id="general-requirement">General requirement</span>:
Discovery mechanism and data model to express authoritative information about the server.

<span id="specific-requirement">Specific requirement</span>:
* Server advertises the location of the authoritative information.
* Server generates the authoritative information.
* Server controls the authoritative information.

<span id="considerations">Considerations</span>:
* Accommodating various use cases that needs to get a hold of authoritative information about the server.
* Should the authoritative information about the server be publicly discoverable?
* Should the authoritative information about the server be publicly readable or access controlled?
* Can specific class of agents (e.g., owners) claim/update authoritative information about the server?
* Should the authoritative information about the server be discoverable from root path, storage, any resource, and/or somewhere else?
* Caching.

<span id="related-issues">Related issues</span> (sample):
* https://github.com/solid/specification/issues/61
* https://github.com/solid/specification/issues/65
* https://github.com/solid/specification/issues/66
* https://github.com/solid/specification/issues/151#issuecomment-596192613
* https://github.com/solid/specification/issues/152
* https://github.com/solid/specification/issues/180
* https://github.com/solid/specification/issues/206
* https://github.com/solid/specification/issues/222
* https://github.com/solid/specification/issues/227
* https://github.com/solid/specification/issues/266
* https://github.com/solid/specification/issues/286
* https://github.com/solid/specification/issues/310
* https://github.com/solid/notifications/issues/14

<span id="notes">Notes</span>:
[...]

Issue #310, opened 2021-09-16
Discovery of storage


kjetilk:
While writing tests for [Section 4.1](https://solidproject.org/TR/protocol#storage), I found no reliable way to discover a storage, and therefore no reliable way to test this section's requirements. 

Now, since we generally have a storage on the server's `/`, for the most part, you could just start with that, and the tests would pass, but that's not the requirement. A server may support several storages, but parts of the URI space doesn't need to be a storage. 

In the current protocol, you can discover the storage by traversing towards it's root container. So, in principle, you could `GET /foo/bar` to discover that the root container is `/foo/`. However, this can't be used for discovery, because it would require the entire URI space to be covered by storages, and that may not be the case. If I have understood it correctly, you could for example have that `/pods/users/foo/` is the root container for the user foo, and similarly for other users. Then, just `GET /bar/baz` will not yield a pointer to a storage. 

To remedy this problem, I suggest that we have a way that clients can discover the pods as a SHOULD level requirement (so that a server may support anonymous pods, but otherwise should make them discoverable).

kjetilk:
There are basically two different mechanisms that I have in mind.

One is that we have a resource `.well-known/solid/storages` that provides an RDF resources listing the storages.

The other is that we define that the response payload to `OPTIONS *` has it.

Personally, `.well-known` has always been itching my aeastheticle (that much out-of-band knowledge of resource identifiers just feel wrong), so I would prefer the latter if possible.

Orthogonal to that is the content of that response. I see two options, one is for example:

```turtle
<> pim:storage </pods/users/foo/> , 
               </pods/users/bar/> .
```

However, there's no value to this being self-describing. Also, it appears to stretch the definition of `pim:storage`. Even though its domain is undefined, the vocabs says rdfs:comment "The storage in which this workspace is", and this resource is certainly no workspace.

Therefore, I rather suggest we do for example:

```turtle
</pods/users/foo/> a pim:Storage . 
</pods/users/bar/> a pim:Storage .
```

That tells you exactly what you need to know in concise terms, and without stretching the definition of anything, it describes the storages exactly the way they describe themselves.

csarven:
>However, this can't be used for discovery, because it would require the entire URI space to be covered by storages and that may not be the case.

I don't quite understand this. Any given resource is part of only one storage and there are no overlapping storages.

[...]

PR #476, opened 2022-10-31
Add definition storage

csarven:
This PR is intended to add the notion of "storage" that is commonly understood and in use. It includes other minor editorial for clarity.

Below is a non-exhaustive list of information pertaining to "storage" that I've taken into account for your reference/consideration.

```turtle
<http://www.w3.org/ns/pim/space#storage> <http://www.w3.org/2000/01/rdf-schema#comment> "The storage in which this workspace is, or the storage which\ncontains this resource, or a storage available to this agent to use." .

<http://www.w3.org/ns/auth/acl#Authorization> <http://www.w3.org/2000/01/rdf-schema#comment> "An element of access control,\n    allowing agent to agents access of some kind to resources or classes of resources" .

<http://www.w3.org/ns/auth/acl#owner> <http://www.w3.org/2000/01/rdf-schema#comment> "The person or other agent which owns this.\n    For example, the owner of a file in a filesystem.\n    There is a sense of right to control.   Typically defaults to the agent who craeted\n    something but can be changed." .

<http://www.w3.org/ns/auth/acl#delegates> <http://www.w3.org/2000/01/rdf-schema#comment> "Delegates a person or another agent to act on behalf of the agent.\n    For example, Alice delegates Bob to act on behalf of Alice for ACL purposes." .
```

https://www.w3.org/DesignIssues/CloudStorage
>There is an architecture in which a few existing or Web protocols are gathered together with some glue to make a world wide system in which applications (desktop or Web Application) can work on top of a layer of commodity read-write storage. Crucial design issues are that principals (users) and groups are identifies by URIs, and so are global in scope, and that elements of storage are access controlled using those global identifiers. The result is that storage becomes a commodity, independent of the application running on it.

>This can be called "socially-aware" storage, because the access control within the storage layer is just powerful enough to implement the social requirements of the social network applications.

https://solidproject.org/TR/2021/protocol-20211217#data-pod
>data pod
>    A data pod is a place for storing documents, with mechanisms for controlling who can access what.

https://solidproject.org/TR/2021/protocol-20211217#storage
>Servers MUST provide one or more storages (pim:Storage) – a space of URIs in which data can be accessed. A storage is the root container for all of its contained resources (see Resource Containment).

https://solidproject.org/ED/protocol#solid-app
>Solid app
>    A Solid app is an application that reads or writes data from one or more storages.

https://solidproject.org/ED/protocol#owner
>owner
>    An owner is a person or a social entity that is considered to have the rights and responsibilities of a data storage. An owner is identified by a URI, and implicitly has control over all data in a storage. An owner is first set at storage provisioning time and can be changed.

https://solid.github.io/solid-oidc/
>Hosting Client ID Document on Solid Storage

https://www.w3.org/wiki/WebAccessControl
>The design goal is that the WebAccessControl storage should be a creative medium in which

>without the intervention of administrative humans running the storage server.

https://solid.github.io/authorization-panel/acp-specification/
> Effective Policies are the Policies controlling access to a resource.

>This creates natural data boundaries that make data storage and authorization more intuitive.

https://github.com/solid/specification/issues/355#issuecomment-979503292
>we should be careful to name stuff that infringes on the authority of a storage (being roughly the same as a "pod", but I use the term storage, since that is what is specified) to control its URI space. This seems to turn into two distinct classes of things though: Those things that are within the URI space controlled by a storage and those that aren't. Those things that are within the space controlled by a storage should be discovered by interrogating the storage, but that leaves us with the requirement that storages must be really easy to discover, which they aren't now (#310). There should be a list of storages hosted by a server somewhere.

https://www.rfc-editor.org/rfc/rfc7231#section-4.3.4
>It does not define how
>   resource state is "stored", nor how such storage might change as a
>   result of a change in resource state

https://www.rfc-editor.org/rfc/rfc7231#section-4.3.5
> associated storage might or might not be reclaimed

https://www.rfc-editor.org/rfc/rfc7231#section-9.1
>Similar naming conventions might exist
   within other types of storage systems.

https://solidproject.org/
>Solid is a specification that lets people store their data securely in decentralized data stores called Pods. Pods are like secure personal web servers for data. When data is stored in someone's Pod, they control which people and applications can access it. 

https://solid.github.io/authorization-panel/authorization-ucr/#uc-inheritance
>The group uses a resource server for storing their information at https://research.example/, which Bob administers as the resource controller.

https://github.com/solid/authentication-panel/blob/main/proposals/HttpSignature.md
>Alice makes a request to a resource &lt;/comments/&gt; on her Personal Online Data Store (POD) at &lt;https://alice.name&gt;

https://solid.github.io/notifications/protocol#notification-channel-discovery
>When a server wants to enable applications to discover Notification Channels available to a storage in which a given resource is in

### DIFF
--- a/ED/protocol.html
+++ b/ED/protocol.html
@@ -488,9 +488,12 @@ margin-top:1em;
 
                   <p property="skos:definition">The Solid Protocol specification defines the following terms. These terms are referenced throughout this specification.</p>
 
-                  <span rel="skos:hasTopConcept"><span resource="#solid-app"></span><span resource="#uniform-resource-identifier"></span><span resource="#resource"></span><span resource="#container-resource"></span><span resource="#root-container"></span><span resource="#resource-metadata"></span><span resource="#agent"></span><span resource="#owner"></span><span resource="#origin"></span><span resource="#read-operation"></span><span resource="#write-operation"></span><span resource="#append-operation"></span></span>
+                  <span rel="skos:hasTopConcept"><span resource="#storage"></span><span resource="#solid-app"></span><span resource="#uniform-resource-identifier"></span><span resource="#resource"></span><span resource="#container-resource"></span><span resource="#root-container"></span><span resource="#resource-metadata"></span><span resource="#agent"></span><span resource="#owner"></span><span resource="#origin"></span><span resource="#read-operation"></span><span resource="#write-operation"></span><span resource="#append-operation"></span></span>
 
                   <dl>
+                    <dt about="#storage" property="skos:prefLabel" typeof="skos:Concept"><dfn id="storage">storage</dfn></dt>
+                    <dd about="#storage" property="skos:definition">A storage is a space of URIs that affords agents controlled access to resources.</dd>
+
                     <dt about="#solid-app" property="skos:prefLabel" typeof="skos:Concept"><dfn id="solid-app">Solid app</dfn></dt>
                     <dd about="#solid-app" property="skos:definition">A Solid app is an application that reads or writes data from one or more <a href="#storage">storages</a>.</dd>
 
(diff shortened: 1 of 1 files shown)


ED/protocol.html
```diff
@@ -513,7 +516,7 @@ <h3 property="schema:name skos:prefLabel">Terminology</h3>
                     <dd about="#agent" property="skos:definition">An agent is a person, social entity, or software identified by a URI; e.g., a WebID denotes an agent [<cite><a class="bibref" href="#bib-webid">WEBID</a></cite>].</dd>
 
                     <dt about="#owner" property="skos:prefLabel" typeof="skos:Concept"><dfn id="owner">owner</dfn></dt>
-                    <dd about="#owner" property="skos:definition">An owner is a person or a social entity that is considered to have the rights and responsibilities of a data storage. An owner is identified by a URI, and implicitly has control over all data in a storage. An owner is first set at storage provisioning time and can be changed.</dd>
+                    <dd about="#owner" property="skos:definition">An owner is a person or a social entity that is considered to have the rights and responsibilities of a storage. An owner is identified by a URI, and implicitly has control over all URIs in a storage. An owner is first set at storage provisioning time and can be changed.</dd>
```
RubenVerborgh, line 519: Nit that it is unclear what "control over all URIs" means. Is it control over the content identified by those URIs? Or something more tricky?

RubenVerborgh (approved):

ED/protocol.html
csarven: I see. How about:

* "implicitly has control over all resources in a storage" - a bit more readable I think.
* "implicitly controls a storage" - simple but hopefully not conflated with the storage resource (root container).

RubenVerborgh: First one works well for me. Second is similarly ambiguous.

timbl (approved):
lgtm

kjetilk (approved):

Issue #147, opened 2020-02-18
Complete Self-Review Questionnaire: Security and Privacy


csarven:
https://www.w3.org/TR/security-privacy-questionnaire/

PR #490, opened 2022-12-26
Update security-privacy-review

csarven:
In response to https://github.com/solid/specification/issues/147 and https://github.com/solid/specification/blob/main/CONTRIBUTING.md#publication-rules

It is encouraged that improvements are made in subsequent PRs.

### DIFF
--- a/ED/protocol.html
+++ b/ED/protocol.html
@@ -1302,58 +1302,58 @@ margin-top:1em;
 
                   <dl rel="schema:hasPart">
                     <dt about="#security-privacy-review-purpose" id="security-privacy-review-purpose"><a href="https://www.w3.org/TR/security-privacy-questionnaire/#purpose" rel="cito:repliesTo">What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?</a></dt>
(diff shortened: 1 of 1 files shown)


TallTed:
*sighs* It would have been good to let this PR sit for at least a few hours, better a few days, before merging.  It's all but pointless to use a PR if there's no opportunity for other eyes to review it; you might as well merge directly.

Why I'm commenting: in line 1323, `resources, (e.g., the` should be `resources (e.g., the` (delete the comma preceding the open-paren).

csarven:
Good enough to create awareness. As mentioned, any significant change can be followed up with a PR. I've introduced this non-normative section long ago and no one even remotely approached to authoring, discussing, or working on it properly in the open. I wrote something down FWIW. I don't mean to debate whether this PR should've sit around longer for editorial and non-editorial changes. Yes, it could've. I decided to merge it - it served its purpose IMO.

I appreciate your attention to detail! So, I've updated the document in https://github.com/solid/specification/commit/616a0aecb883104646fa12c980af0289fae7584d which makes the style consistent with the rest of the document.

PR #516, opened 2023-03-28
Add 2023-03-29 agenda and minutes

csarven:


meetings/2023-03-29.md
```diff
@@ -38,41 +41,145 @@
...
+* SC: Noticed the background color. I can update the PR. 
+* PAC: One quote has been inserted in the PR, but that's from a different PR. 
+* SC: There were two PRs. The other one got merged. 
+
+ACTION: SC to update stylesheet and merge.
 
 
 ### Clarify and reference Solid Protocol, CG, use cases
 URL: https://github.com/solid/solid-wg-charter/pull/8
+* SC: We have one approval (PAC). This is just changing Solid specification to Protocol, and the other one is reference to Solid Community Group, and the other one where it says Pods I think it means to say server. Fourth is repository with use cases and user stories. We want to put there as much as possible of what the group has done.
```
TallTed, line 70:
```suggestion
* SC: We have one approval (PAC). This is just changing Solid specification to Protocol; a reference to Solid Community Group; and another where it says Pods, I think it means to say server. Fourth is repository with use cases and user stories. We want to put as much as possible of what the group has done there.
```

meetings/2023-03-29.md
```diff
@@ -38,41 +41,145 @@
...
+* PAC: I see a sentence you did not copy from the template: 
+  https://w3c.github.io/charter-drafts/charter-template.html#success-criteria (second to last sentence of this section)
+* PAC: Was wondering if it was deliberate to omit the sentence, now I see it was not. Since it's part of the template we might as well include.
+* SC: I noticed in the source code, when you clone this Charter template and look at it I see a different version from the one on the GitHub pages one. Which may explain why we missed each other there (in this PR's comments)
```
TallTed, line 82:
```suggestion
* SC: This is trying to align content with the latest Charter template W3C is using. This PR links to an issue mentioning that. And then responding/filling in content based on that. Key areas: testing and interop, horizontal reviews. Main additions are clarifying what will count as interoperable (can be verified with open test suites) and committing to the fact we're willing to work on tests/have a testing plan as early as possible, and how to promote interoperability in testing. We have work in the CG around testing, QA. Some of the specs have already been moving in that direction, and we already have test suites that are testing that. Building on the W3C QA activities which go back over two decades of work. This is important for any group to properly show how interoperability can be verified and how implementation reports are obtained, etc. The last part is not too different from previous template.
* PAC: I see a sentence you did not copy from the template (second to last sentence of [this section](https://w3c.github.io/charter-drafts/charter-template.html#success-criteria))
* PAC: I was wondering if it was deliberate to omit the sentence; now I see it was not. Since it's part of the template, we might as well include it.
* SC: I noticed in the source code, when you clone this Charter template and look at it, I see a different version than is on GitHub pages. Which may explain why we missed each other there (in this PR's comments).
```

meetings/2023-03-29.md
```diff
@@ -38,41 +41,145 @@
...
+* PAC: Maybe GH pages is lagging behind?
+* SC: The repo says you should clone the GH pages branch, and that's where changes will happen. I can investigate and if you feel we should do what we see now on the website then happy to do that.
```
TallTed, line 84:
```suggestion
* SC: The repo says you should clone the GH pages branch, and that's where changes will happen. I can investigate, and if you feel we should do what we see now on the website, then happy to do that.
```

meetings/2023-03-29.md
```diff
@@ -38,41 +41,145 @@
...
+* SC: No objection to leaving it in. Just was looking at from the perspective of what the charter communicates that's coming from a CG and its incubation. Let's see how it goes with the Members.
+* SC: Integration is specific, not a blanket statement about whatever is available from a particular Group, Charter, Homepage, or Whatever - we don't "integrate" specification details with arbitrary documents but instead do it with actual Specifications. (There is a whole thing on Coordination with other W3C Groups which seems more suitable as a general interest.) An exhaustive list of TRs is not required either as there is the notion of normative references that will pick them up.
+* SC: I don't think we should link to a group or charter. 
+* eP: In that case I think we could link to the VC 1.1 (current) and then coordinations link to the VC WG, I think it's valid that by the time 2.0 is ready, there won't be a problem with charter linking to 1.1. 
+* SC: Isn't the TR latest, and that'd be covered? I doubt anyone would care about the versions.
+* PAC: latest: https://www.w3.org/TR/vc-data-model/ , specific version: https://www.w3.org/TR/2022/REC-vc-data-model-20220303/
```
TallTed, line 117:
```suggestion
* SC: This is one where there were different views (in the comments). My main point is the difference between CG and WG. CG is one of the groups that could propose work towards a WG, and it is intended to be where work is incubated, implementation experienced, and adoption. This is something that the Members/reviewers look for. 
* SC: Two things in here. One is about acknowledging notifications about resource changes and linking to that. PAC already agreed. Any other objections to including that part (linking to Notifications protocol)? 
* eP: No objections. 
* SC: The other part is about the Verifiable Credentials. This is about access grants using VCs. The issue was that there is no spec about access grants. And it is not sufficiently incubated in that we don't have multiple implementations interoping that demonstrate this. This type of thing is being equated on the same grounds as something like WAC, Notification Protocol, ACP, OIDC, and so on. There's no spec for access grants, it is not implemented, and Solid Protocol does not mention anything about access grants. All the other things suggesting things that will be detailed are fine, because there are specs, work, etc., but that's not the case for access grants. AGs is not checking a bunch of boxes, and I think that will be easily caught by reviewers. 
* eP: In interop spec, we have access grants and data grants and they are not exactly the same Inrupt came up with. ??? meeting last month about access grants. I suggested the authorization-to-access should be represented with a VC. There is work that does not talk about VC directly, but will require VC to complete. 
* SC: There's no interop between the notion of AG and the one that's been worked on by another group?
* eP: Yes, and they cover different parts of the problem. If we reconcile, they can provide different parts of the ???. In interop there's discovery covered by those. In some ways, those parts (scribe note: which parts???) are complementary.
* SC: There is nothing that is reconciled in the sense that is spec'd. For example, Solid Protocol refers to many things (WAC, ACP, WebID, etc.), but does not refer to any notion of AG in any of the versions you described. There's no single spec/work that is there. I understand there are considerations on how they can work out, but the question is, are there enough implementations that are interoping? I understand there are servers that are working, and we have enough implementation feedback, but there's no spec, and we need to do that as an indication of well incubated work. I don't see that here. There's hardly one implementation that is also not open source. That's the perspective that I'm looking at. 
* PAC: There are in fact some implementations, some documentation about Inrupt's ESS, but as you said, it is one implementation that is not open source and might be too restricted to be considered proper incubation. 
* SC: To be fair, it's not that the closed-source is not sufficient for incubation, but that version/implementation differs from the other incubated work that has been done in the interop spec/panel. Not sure what the implementations on that are either.
* SC: Bottom line is, we could throw all the random things that we've been experimenting with or thinking about into this charter. The notion of requesting access is not new in Solid. I even proposed one based on Inbox, and have implemented that work. But I will not ask the group that we should put this in the charter, even though conceptually it is along those lines. There's a list of things we have been experimenting with, but we need to be careful not to throw everything into the charter and expect that it'll actually be taken up.
* PAC: I agree with you that we should not include any random idea in the charter. That said, some people have put that here, I hear there are different people working on the idea of including VCs. I think it'd be a shame to close that door, if several people are working on it. Granted, this is not as mature and incubated as the other points on the list, but we don;'t have too many of those either. Also this list of things are points that MAY be addressed by the group. Not too committing. Group might still decide in the middle that this is not ready and will not be included. Success criteria says we need two independent implementations so I'm in favor of keeping it here, as there are not too many random ideas in this charter that it'll be too risky. 
* eP: Just to emphasize, I don't think it's a random idea. We have an AuthZ use case about user authorizing a client, which is not currently addressed completely. Combined with Inrupt's AG implementation, this could be addressed. There are some issues, but I don't think it's just about VC, but other problems with implementations. I don't see it as a random thing. One of the reasons why Solid Protocol doesn't address it is because it is incomplete. It doesn't address this use case. "Trusted apps" is not a proper solution to this problem. It is definitely not random. 
* SC: I am aware. You may remember that the Solid CG has a joint meeting with Credentials CG, and one of the reasons why I set that up was because work in that group could be useful to Solid. Prior to that we had not incorporated any of the CCG group's work into Solid, more of less after those meetings was where the group started taking more of their work items into consideration. I'm well aware and in support of using the specs out there that other groups have developed. I was just looking at it from what the Charter is supposed to signal. No judgement about the technology itself. I'm glad we are doing that and we should continue. 
* PAC: In terms of signal, we could just put some link in the Coordination section. Not having any mention of VCs in a project that emphasized privacy will be a mistake, but I take your point and maybe it should be better to have it in the Coordination section. I'm in favor of keeping it also for the signal that we send that we should coordinate with VC people.
* SC: No objection to leaving it in. Just was looking at from the perspective of what the charter communicates that's coming from a CG and its incubation. Let's see how it goes with the Members.
* SC: Integration is specific, not a blanket statement about whatever is available from a particular Group, Charter, Homepage, or Whatever — we don't "integrate" specification details with arbitrary documents, but instead do it with actual Specifications. (There is a whole thing on Coordination with other W3C Groups which seems more suitable as a general interest.) An exhaustive list of TRs is not required either, as there is the notion of normative references that will pick them up.
* SC: I don't think we should link to a group or charter. 
* eP: In that case I think we could link to the VC 1.1 (current) and then coordinations link to the VC WG. I think it's valid that by the time 2.0 is ready, there won't be a problem with charter linking to 1.1. 
* SC: Isn't the TR latest, and that'd be covered? I doubt anyone would care about the versions.
* PAC: [latest](https://www.w3.org/TR/vc-data-model/) is [specific version](https://www.w3.org/TR/2022/REC-vc-data-model-20220303/)
```

PR #518, opened 2023-04-05
Add 2023-04-05 agenda and minutes

csarven:

### DIFF
--- a/meetings/2023-04-05.md
+++ b/meetings/2023-04-05.md
@@ -1,0 +1,78 @@
+# W3C Solid Community Group: Weekly
+
+* Date: 2023-04-05T14:00:00Z
+* Call: https://meet.jit.si/solid-cg
+* Chat: https://gitter.im/solid/specification
+* Repository: https://github.com/solid/specification
+* Status: Draft
+
+## Present
+* [Sarven Capadisli](https://csarven.ca/#i)
+
+---
+
+## Announcements
+
+### Meeting Guidelines
+* [W3C Solid Community Group Calendar](https://www.w3.org/groups/cg/solid/calendar).
+* [W3C Solid Community Group Meeting Guidelines](https://github.com/solid/specification/blob/main/meetings/README.md).
+* No audio or video recording, or automated transcripts without consent. Meetings are transcribed and made public. If consent is withheld by anyone, recording/retention must not occur.
+* Join queue to talk.
+* Topics can be proposed at the bottom of the agenda to be discussed as time allows. Make it known if a topic is urgent or cannot be postponed.
+
+### Participation and Code of Conduct
+* [Join the W3C Solid Community Group](https://www.w3.org/community/solid/join), [W3C Account Request](http://www.w3.org/accounts/request), [W3C Community Contributor License Agreement](https://www.w3.org/community/about/agreements/cla/).
+* [Solid Code of Conduct](https://github.com/solid/process/blob/main/code-of-conduct.md), [Positive Work Environment at W3C: Code of Ethics and Professional Conduct](https://www.w3.org/Consortium/cepc/)
+* Operating principle for effective participation is to allow access across disabilities, across country borders, and across time. Feedback on tooling and meeting timing is welcome.
+* If this is your first time, welcome! please introduce yourself.
+
+
+### Scribes
+* name
+
+### Introductions
+* name: text
+
+---
+
+
+## Topics
+
+### Overview of today's agenda
+
+* SC: Today's meeting will focus on the Solid WG Charter that follows 2023-03-22 meeting topic for reviews: https://github.com/solid/specification/blob/main/meetings/2023-03-22.md#solid-wg and continues on topics in 2023-03-29 meeting: https://github.com/solid/specification/blob/main/meetings/2023-03-29.md
+* SC: First open PRs then issues:
+
+### Use w3c/charter-draft's template
+URL: https://github.com/solid/solid-wg-charter/pull/24
+
+* SC: Resolves https://github.com/solid/solid-wg-charter/issues/2 . Already ack'd by PAC in https://github.com/solid/solid-wg-charter/issues/2#issuecomment-1479836619 . Assigned to PAC.
+
+
+### Use technical terms - storage instead of pod
+URL: https://github.com/solid/solid-wg-charter/pull/25
+
+* SC: Resolves https://github.com/solid/solid-wg-charter/issues/3
+
+
+### Update scope - remove client-client
+URL: https://github.com/solid/solid-wg-charter/pull/26
+
+* SC: Resolves https://github.com/solid/solid-wg-charter/issues/23
+* SC: Part of https://github.com/solid/solid-wg-charter/issues/9
+
+
+### Scope needs to be tightly defined with narrow focus
+URL: https://github.com/solid/solid-wg-charter/issues/9
+
+
+### Clarification on Solid and comparisons
+URL: https://github.com/solid/solid-wg-charter/issues/5
+
+
+### Update mission
+URL: https://github.com/solid/solid-wg-charter/issues/7
+
+
+### Review from TAG should be requested prior to Member review
+URL: https://github.com/solid/solid-wg-charter/issues/15


meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* SC: I read "recommend a set of practices" as "best practices or guidelines". If it were requirements it'd be a deliverable, but this seems to be BP&G. 
+* SC: In the Charter "out of scope" has two parts: definition of identify mechanisms and definition of linked data formats. The part about set of practices is a very broad thing being thrown in. I'm not sure if we need to recommend best practices around AuthN or identity where, for example, we're also thinking those are out of scope. If we do anything about identity mechanisms, we could do best practices, but this seems to be reaching to far.
+* eP: I'm not sure if the intention wasn't to somehow mention that we'd try to build on other standards. I guess it might have been formulated in this way to have flexibility around normative requirements or recommendations but not committing to requirements.
```
TallTed, line 126:
```suggestion
* SC: In the Charter, "out of scope" has two parts: definition of identify mechanisms, and definition of linked data formats. The part about set of practices is a very broad thing being thrown in. I'm not sure if we need to recommend best practices around AuthN or identity where, for example, we're also thinking those are out of scope. If we do anything about identity mechanisms, we could do best practices, but this seems to be reaching too far.
* eP: I'm not sure whether the intention wasn't to somehow mention that we'd try to build on other standards. I guess it might have been formulated in this way to have flexibility around normative requirements or recommendations, but not committing to requirements. 
```

meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* SC: In the Charter "out of scope" has two parts: definition of identify mechanisms and definition of linked data formats. The part about set of practices is a very broad thing being thrown in. I'm not sure if we need to recommend best practices around AuthN or identity where, for example, we're also thinking those are out of scope. If we do anything about identity mechanisms, we could do best practices, but this seems to be reaching to far.
+* eP: I'm not sure if the intention wasn't to somehow mention that we'd try to build on other standards. I guess it might have been formulated in this way to have flexibility around normative requirements or recommendations but not committing to requirements. 
+* SC: May I ask eP to chime in in this issue? For us to translate this issue into a PR we need to remove some ambiguities. If you could say in what ways something would be within a practice and which ways out of scope, we can update the language. Right now it is virtually everything we do in Solid and I don't think that's the intention of the Charter.
```
TallTed, line 127:
```suggestion
* SC: May I ask eP to chime in in this issue? For us to translate this issue into a PR, we need to remove some ambiguities. If you could say in what ways something would be within a practice and which ways out of scope, we can update the language. Right now, it is virtually everything we do in Solid, and I don't think that's the intention of the Charter.
```

meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* JZ: I'm thinking of a conversation I had recently with someone who has big ideas on infrastructure and the relationship between an URI and identity. Not client-client and not even server-client, but something that Solid needs to be thinking about.
+* SC: That's under the CG. New material or things that needs to be conceptually worked out, spec'ed, incubated, or needs adoption falls under CG, in general. For example, if we had a lot of implementations with a lot of user adoption out there for WebID-Profile spec, or Type Indexes, we could make an argument that we have a good sense of how it works but needs to be spec'ed, so we'd need it to be part of the Charter, but we're not there yet. The purpose of CG is to move us closer to that so that it can be picked up by WG. Hope that clarifies.
+* JZ: Yes. WG has a specific charter and CG does everything else.
```
TallTed, line 132:
```suggestion
* JZ: I'm thinking of a conversation I had recently with someone who has big ideas on infrastructure and the relationship between an URI and identity. Not client-client and not even server-client, but something that Solid needs to be thinking about.
* SC: That's under the CG. New material or anything that needs to be conceptually worked out, spec'ed, incubated, or needs adoption falls under CG, in general. For example, if we had a lot of implementations with a lot of user adoption out there for WebID-Profile spec, or Type Indexes, we could make an argument that we have a good sense of how it works but it needs to be spec'ed, so we'd need it to be part of the Charter, but we're not there yet. The purpose of CG is to move things closer to that so that they can be picked up by WG. Hope that clarifies.
* JZ: Yes. WG has a specific charter and CG does everything else.
```

meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* SC: That's under the CG. New material or things that needs to be conceptually worked out, spec'ed, incubated, or needs adoption falls under CG, in general. For example, if we had a lot of implementations with a lot of user adoption out there for WebID-Profile spec, or Type Indexes, we could make an argument that we have a good sense of how it works but needs to be spec'ed, so we'd need it to be part of the Charter, but we're not there yet. The purpose of CG is to move us closer to that so that it can be picked up by WG. Hope that clarifies.
+* JZ: Yes. WG has a specific charter and CG does everything else. 
+* SC: Yes. If it's mature enough to be further worked on, it should be carried out by WG towards a recommendation. CG basically graduates specs towards that. If you look at TR homepage/index of work items, we know not all of those are as mature as they can be. There's a lot of activity, but some items are not hitting all those things sufficiently. It's not about favoritism of solutions, but makes it easier for the Charter to be accepted if there is sufficient incubation.
```
TallTed, line 133:
```suggestion
* SC: Yes. If something's mature enough to be further worked on, it should be carried out by WG towards a recommendation. CG basically graduates specs towards that. If you look at TR homepage/index of work items, we know not all of those are as mature as they can be. There's a lot of activity, but some items are not hitting all those things sufficiently. It's not about favoritism of solutions, but making it easier for the Charter to be accepted following sufficient incubation.
```

meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* JZ: For example, WebID-Profile and Type Index, those are client-client, so if those are out of scope they will never graduate to WG.
```
TallTed, line 134:
```suggestion
* JZ: For example, WebID-Profile and Type Index, those are client-client, so if those are out of scope, they will never graduate to WG.
```

meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* SC: In this particular charter. That's one criterion for this charter (server-client) but there's other criteria for incubation.
+* JZ: That clarifies, thanks.
+* eP: I have [a good example (solid-oidc/issue#219)](https://github.com/solid/solid-oidc/issues/219). Talks about security considerations regarding Solid-OIDC issuer. This is very important because it is the foundation of security. It's an example where I think there can't be one normative statement that makes sure this is taken care of, more of a best practice informing not allowing this statement to be altered. That could be a concrete example. 
+* SC: Comment on the issue, but do not go too deep on the issue. We shouldn't get too into the considerations around that particular issue. Just have a sense of the set of issues that are intended. You can use an example in brackets, but I wouldn't highlight it as a discussion point - not the center of this issue. 
+* SC: JZ, if you could also chime in with your feedback in the issue that'd be great.
```
TallTed, line 139:
```suggestion
* eP: I have [a good example (solid-oidc/issue#219)](https://github.com/solid/solid-oidc/issues/219). Talks about security considerations regarding Solid-OIDC issuer. This is very important because it is the foundation of security. It's an example where I think there can't be one normative statement that makes sure this is taken care of, more of a best practice informing not allowing this statement to be altered. That could be a concrete example. 
* SC: Comment on the issue, but do not go too deep there. We shouldn't get too into the considerations around any particular issue. Just have a sense of the set of issues that are intended. You can use an example in brackets, but I wouldn't highlight it as a discussion point — it's not the center of this issue. 
* SC: JZ, if you could also chime in with your feedback in the issue, that'd be great.
```

meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* SC: There's 2 items here, one about best practices to request and grant access. We covered that as something we can potentially detail in another PR. But the Best Practices of that is not something mature enough. The next item is about state synchronization regarding changes to resources in Solid Pods. I have no memory of any sort of state sync work done in the CG, whether it is being spec'ed or any implementations, or even applications demonstrating that. It's hard to say it's in scope when it has not been incubated.
```
TallTed, line 142:
```suggestion
* JZ: Generally it sounds good to me. I think eP is right that there needs to be a place to say something along the lines of that issue. That seems to me to be a server responsibility; who else could prevent changes to the OIDC issuer? That seems to fall within the scope of WG. Who else would implement such a security policy other than. a server?
* SC: If it's about security/privacy consideration, we already have sections for it in Solid Protocol, and those details can be mentioned there. That doesn't need to be mentioned as in scope. What's intended in the scope here is what other high-level material that could be potentially spec'ed, that could be a new deliverable/non-normative document that could be spun out. If it's a general consideration, that does not need to be mentioned as in scope. If it's a new document, I am not sure we have enough material to tackle that. Item 3 lists a lot of things, like security considerations for all those documents. Verification, for example, has not been incubated or spec'ed in Solid Protocol. We do not have that material, and we're not in a position to recommend BP&G as to how verification should be. If we had a lot of material spec'ed, we'd be in a better position to make those recommendations. 
* SC: There are two items here. One about best practices to request and grant access. We covered that as something we can potentially detail in another PR. But the Best Practices of that is not something mature enough. The other item is about state synchronization regarding changes to resources in Solid Pods. I have no memory of any sort of state sync work done in the CG, whether it is being spec'ed or any implementations, or even applications demonstrating that. It's hard to say it's in scope when it has not been incubated. 
```

meetings/2023-04-05.md
```diff
@@ -47,23 +54,99 @@
...
+* eP: We should check with the author of this sentence if it means to talk about Notifications, as it follows the state.
+* SC: Yes. We already said Notifications protocol is a part of the protocol we can expand on. If it's hinting at notifications, then it's already covered and doesn't need to be mentioned.
+* eP: There's duplication in deliverables/scope. 
+* SC: I think I'd like to ask PAC to assist with scope and deliverables. The deliverables are concrete things that we will give. Scope will be things that are still within consideration of the work. For example, we use WebID, I'd consider the possibility of using DIDs still within scope of Solid Protocol, but I wouldn't list it as a deliverable. When the WG happens that discussion can continue and if it seems feasible to use DIDs as interchangeable identifiers, that might be what's in the scope, but right now it is not a deliverable. Similarly, out of scope is: we will not be coming up with a new identity mechanism.
```
TallTed, line 146:
```suggestion
* SC: I think I'd like to ask PAC to assist with scope and deliverables. The deliverables are concrete things that we will give. Scope will be things that are still within consideration of the work. For example, we use WebID. I'd consider the possibility of using DIDs still within scope of Solid Protocol, but I wouldn't list it as a deliverable. When the WG happens, that discussion can continue, and if it seems feasible to use DIDs as interchangeable identifiers, that might be in scope, but right now it is not a deliverable. Similarly, "coming up with a new identity mechanism" is out of scope.
```

TallTed (changes requested):
The usual... 

And recognition that JZ seems to have had (and may still have) the impression that there will only ever be *this* WG focused on Solid-relevant topics — which is why he's concerned that topics that "are out of scope they will never graduate to WG". That misunderstanding appears to still need to be addressed.

csarven:
>That misunderstanding appears to still need to be addressed.

In the meeting we arrived at (which may not capture all of the said details) clarifying the focus of this particular charter:

>* SC: In this particular charter. That's one criterion for this charter (server-client) but there's other criteria for incubation.
>* JZ: That clarifies, thanks.

It goes without saying that it is best to reach consensus as open and transparent as possible in the Group (and on public record) to avoid potential misunderstandings. Random private conversations that have no bearing on Group's position tend to cause confusion, potential harm, or waste time. And the text we write must reflect Group's position.

PR #535, opened 2023-06-21
Add 2023-06-21 agenda and minutes

csarven:

### DIFF
--- a/meetings/2023-06-21.md
+++ b/meetings/2023-06-21.md
@@ -1,0 +1,63 @@
+# W3C Solid Community Group: Weekly
+
+* Date: 2023-06-21T14:00:00Z
+* Call: https://meet.jit.si/solid-cg
+* Chat: https://gitter.im/solid/specification
+* Repository: https://github.com/solid/specification
+* Status: Draft
+
+## Present
+* [Sarven Capadisli](https://csarven.ca/#i)
+
+---
+
+## Announcements
+
+### Meeting Guidelines
+* [W3C Solid Community Group Calendar](https://www.w3.org/groups/cg/solid/calendar).
+* [W3C Solid Community Group Meeting Guidelines](https://github.com/solid/specification/blob/main/meetings/README.md).
+* No audio or video recording, or automated transcripts without consent. Meetings are transcribed and made public. If consent is withheld by anyone, recording/retention must not occur.
+* Join queue to talk.
+* Topics can be proposed at the bottom of the agenda to be discussed as time allows. Make it known if a topic is urgent or cannot be postponed.
+
+### Participation and Code of Conduct
+* [Join the W3C Solid Community Group](https://www.w3.org/community/solid/join), [W3C Account Request](http://www.w3.org/accounts/request), [W3C Community Contributor License Agreement](https://www.w3.org/community/about/agreements/cla/).
+* [Solid Code of Conduct](https://github.com/solid/process/blob/main/code-of-conduct.md), [Positive Work Environment at W3C: Code of Ethics and Professional Conduct](https://www.w3.org/Consortium/cepc/)
+* Operating principle for effective participation is to allow access across disabilities, across country borders, and across time. Feedback on tooling and meeting timing is welcome.
+* If this is your first time, welcome! please introduce yourself.
+
+
+### Scribes
+* name
+
+### Introductions
+* name: text
+
+---
+
+
+## Topics
+
+### WIP Implementation Feedback
+* SC: We'll allocate some time for implementation feedback or interest to implement. Links to products/projects and demos welcome.
+
+
+### Add WebID to Dependencies
+URL: https://github.com/solid/solid-wg-charter/pull/40
+
+* SC: Resolves https://github.com/solid/solid-wg-charter/issues/39 .
+>* SC: I suggest if no objections by 2023-06-21, we can merge with WebID CG Chair's go ahead on behalf of the group.
+>* HS: Ok, I sent [an email message out to WebID CG group](https://lists.w3.org/Archives/Public/public-webid/2023Jun/0001.html). Seems people were ok. But we can wait a week.
+* SC: How about now?
+
+
+### Update mission
+URL: https://github.com/solid/solid-wg-charter/issues/7
+
+* SC: This is editorial. PRs welcome.
+
+
+### Alternative solutions to container HATEOAS
+URL: https://github.com/solid/specification/issues/525
+
+* SC: Proposed by RG and eP.
--- a/meetings/2023-06-21.md
+++ b/meetings/2023-06-21.md
@@ -8,7 +8,15 @@
 
 ## Present
 * [Sarven Capadisli](https://csarven.ca/#i)
-
+* [elf Pavlik](https://elf-pavlik.hackers4peace.net)
+* [Pierre-Antoine Champin](https://solid.champin.net/pa/profile/card#me)
+* Laurens Debackere
+* Hadrian Zbarcea
+* [Jesse Wright (jeswr)](https://www.jeswr.org/#me)
+* Maxime Lecoq
+* April Daly
+* Jeff Zucker
+* [Rahul Gupta](https://cxres.pages.dev/profile#i)
 ---
 
 ## Announcements
@@ -28,7 +36,7 @@
 
 
 ### Scribes
-* name
+* Pierre-Antoine
 
 ### Introductions
 * name: text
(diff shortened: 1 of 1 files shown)


bblfish (approved):
I checked again on the WebID mailing list and did not see any objections to this happening. 
People are interested in JSON-LD serialisations of WebIDs, which seems like something that would best be determined by a WG with a view on the use cases.

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+* [elf Pavlik](https://elf-pavlik.hackers4peace.net)
+* [Pierre-Antoine Champin](https://solid.champin.net/pa/profile/card#me)
+* Laurens Debackere
+* Hadrian Zbarcea
+* [Jesse Wright (jeswr)](https://www.jeswr.org/#me)
+* Maxime Lecoq
+* April Daly
+* Jeff Zucker
+* [Rahul Gupta](https://cxres.pages.dev/profile#i)
+---
```
bblfish, line 20:
```suggestion
* [Henry Story](https://bblfish.net/people/henry/card#me)
---
```

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+### Introductions
+* name: text
+
+---
+
+
+## Topics
+
+### Recording demos
+* SC: I don't mind recording it but this raises the question of where to host the video afterward. I could put it on my website but the CG probably does not want to rely on my website being up -- and I don't want this responsibility.
```
TallTed, line 50: I missed the meeting, so technically this comment doesn't exist, but...

The obvious thought for video uploads is YouTube. I don't know current bandwidth or volume limitations on github.io, drive.google.com, dropbox.com, box.com, etc., but those seems like other possible places. For the latter bunch, I would think it OK to force download for local playback, as opposed to streaming video, which might lower the pipe if not disk consumption.

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+* SC: would you consider the Turtle / JSON-LD and the HTML representation to be equivalent?
+* eP: not really. They both describe me, but intended to different audiences.
+* PA: Have you tried using this? I have experimented recently to have a pod on my CSS that is not using /profile/card#me but rather that I've been using some URL for some URL. I've tried it something on that page to be Solid compliant. WebID which is not hosted on a pod even though it contains the required properties, to find my pod and access. Looking at your, it redirects?
+* eP: It is just a static file currently. I hope to have an answer to your question next week. I can file bug reports about where the WebID is hosted.
+* JZ: about profiles not stored in a Pod: that's what ESS does currently does. And this cause problems in the WebID profile group. We are interested in your feedback.
+* PA: happy to join. My reading of the WebID profile spec was that it was ok, but let's discuss this.
```
TallTed, line 60: From perspective of expired WebID WG, it shouldn't matter where a WebID Profile is hosted — in a pod, in a filesystem, on a non-Solid (HTTPS, SFTP, etc.) server, etc. (Originally, HTTP, FTP, and the like were OK, but since security has been more of a concern, HTTPS or other secured (typically by TLS) protocols are generally a SHOULD if not a MUST.) It's just necessary that the WebID dereference to the Profile document, which contains the WebID again and one or more associated public keys.

In other words — if any Solid server chokes on a WebID Profile because it's not hosted on a Solid server, that's a bug in the Solid server (or perhaps in the non-Solid server hosting the Profile document).

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Flemish Government Solid MVP on University Degrees
+
+* LD: this MVP was released in May.
+  (possible to share the slides ?)
+  S3: diploma data already available from a central database, but not easily shareable by users with future employers.
+  S4: still discussing the terminology most appropriate for users ("personal safe"/"personal data vault", also thinking about "WebID")
+  40 Job applications happened through this MVP in the last month(?), no major issue encountered
```
TallTed, line 68:
```suggestion
    (possible to share the slides ?)
    S3: diploma data already available from a central database, but not easily shareable by users with future employers.
    S4: still discussing the terminology most appropriate for users ("personal safe"/"personal data vault", also thinking about "WebID")
    40 Job applications happened through this MVP in the last month(?), no major issue encountered
```

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+* JZ: fantastic stuff. Could you explain again what the entities are and their responsibilities?
+* LD: (showing Slide 6, which was skipped during the demo). Data Utility company provides GDPR-compliant pods on behalf of the government -- in the future, multiple providers. Ranstat acts as the verifier, on behalf of the potential employers.
+* eP: congratulations. Is the server using the latest version of Solid-OIDC or the old one?
+* LD: the servers implement the latest Solid-OIDC, but our ecosystem does not use DPoP yet. Support for this will follow after the summer.
+* PA: re JSON-LD and VCs, VC is a RDF dataset with multiple named graphs, which is not supported by LDP.
+* LD: yes, the ESS and the CSS behave differently, this is undefined behaviour currently in LDP and the Solid Protocol. We should address that.
+* SC: We currently refer to JSON-LD 1.1, which is not equivalent to turtle.
+* PA: JSON-LD 1.0 already supported named graphs. This was an oversight of LDP -- I plead guilty, I was part of that WG.
+* HS: you can not do credentials without named graphs. This is an interesting issue.
```
TallTed, line 77:
```suggestion
* PA: re JSON-LD and VCs, VC is an RDF dataset with multiple named graphs, which is not supported by LDP.
* LD: yes, ESS and CSS behave differently. This is currently undefined behaviour in LDP and the Solid Protocol. We should address that.
* SC: We currently refer to JSON-LD 1.1, which is not equivalent to Turtle.
* PA: JSON-LD 1.0 already supported named graphs. This was an oversight of LDP -- I plead guilty, I was part of that WG.
* HS: you cannot do credentials without named graphs. This is an interesting issue.
```

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+
+* SC: I'll see what I can do with the screencast. Useful even out of the context of this call.
+* HS: I think we should adapt the rules. The privacy concerns have evolved since COVID.
+* SC: the consent of the group is still necessary before any recording. 
+
+
+### Add WebID to Dependencies
+URL: https://github.com/solid/solid-wg-charter/pull/40
+
+* SC: Resolves https://github.com/solid/solid-wg-charter/issues/39 .
```
TallTed, line 90:
```suggestion
* SC: Resolves [`solid/solid-wg-charter/issues/39`](https://github.com/solid/solid-wg-charter/issues/39).
```

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+>* SC: I suggest if no objections by 2023-06-21, we can merge with WebID CG Chair's go ahead on behalf of the group.
+>* HS: Ok, I sent [an email message out to WebID CG group](https://lists.w3.org/Archives/Public/public-webid/2023Jun/0001.html). Seems people were ok. But we can wait a week.
+* SC: How about now?
+  Note that the goal is not to turn WebID into a purely Solid thing. The WebID deliverable from the Solid WG would be usable in other contexts.
```
TallTed, line 94:
```suggestion
  ... Note that the goal is not to turn WebID into a purely Solid thing. The WebID deliverable from the Solid WG would be usable in other contexts.
```

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+
+* SC: This is editorial. PRs welcome.
+
+
+### Alternative solutions to container HATEOAS
+URL: https://github.com/solid/specification/issues/525
+
+* SC: Proposed by RG and eP.
+* RG: if you want to upload something in a container, then GET the container, it is not clear what you should retrieve: only a link to the uploaded content, or also a description of it?
+  Maybe a `Prefer` header could be used? eP has made a proposal in that sense.
```
TallTed, line 112:
```suggestion
* RG: if you want to upload something in a container, then GET the container, it is not clear what you should retrieve: only a link to the uploaded content, or also a description of it?
  ... Maybe a `Prefer` header could be used? eP has made a proposal in that sense.
```

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+* SC: can you explain what you mean by "upload something"?
+* RG: imagine a container `/foo/`. Either you POST `bar` of put to `foo/bar`. What if you want to put in the container that is elsewhere?
+* SC: Can't link to a resource outside of the container path since client can't update the containment statements.
+* PA: Ran into this.. Re-read the LDP spec for that matter, it seems to me that LDP contains some guidance on Prefer header. I tested on CSS and NSS. It appears not implemented. LDP says MAY (or SHOULD?). Just implementing LDP may solve that problem. https://www.w3.org/TR/ldp/#prefer-parameters
+* eP: for me it is a broader issue. It could be a custom relation, not only `ldp:contains`. Related to server-managed statements, which the client can not fully control. Also occurs in Notifications where server should manage Subscription Service discovery. We need a longer meeting to dive into this problem. (discussion about the managed statement occurring in an auxiliary resource). Look at the issue that RG created, it also links to other long-discussed issues.
```
TallTed, line 117:
```suggestion
* RG: Imagine a container `/foo/`. Either you `POST` to `bar` or `PUT` to `foo/bar`. What if you want to put something in the container that is elsewhere?
* SC: Can't link to a resource outside of the container path, since client can't update the containment statements.
* PA: Ran into this. Re-read the LDP spec for that matter. It seems to me that LDP contains some guidance on `Prefer` header. I tested on CSS and NSS. It appears not implemented. LDP says `MAY` (or `SHOULD`?). Just implementing LDP may solve that problem. https://www.w3.org/TR/ldp/#prefer-parameters
* eP: For me, it is a broader issue. It could be a custom relation, not only `ldp:contains`. Related to server-managed statements, which the client cannot fully control. Also occurs in Notifications where server should manage Subscription Service discovery. We need a longer meeting to dive into this problem. (discussion about the managed statement occurring in an auxiliary resource). Look at the issue that RG created; it also links to other long-discussed issues.
```

meetings/2023-06-21.md
```diff
@@ -0,0 +1,130 @@
...
+* RG: not sure that breaking apart is a good solution. The risk is to solve one part and break another one.
+* SC: agreed, but we are looking at different specs. Need to identify if that's fixing something in a spec, adding something in it, or adding a new spec.
+* RG: we are mostly stating the problem, not coming with a solution. That's why we need the longer discussion.
+
+### WG Charter
+* PA: I wanted to ask something about permissions for merging editorial changes for the mission. Is that something that will impact horizontal review?
+* SC: No. I would just still like that update on the mission to go through even while in horizontal review to have an accurate representation of the work.
+* PA: I just wanted to keep the ball rolling.
+* SC: Sure. We can capture this conversation here in the minutes.
+* PA: We need that for the horizontal review, once WebID is merged I'll ping the different parts of W3C to get the review started.
```
TallTed, line 129:
```suggestion
* PA: We need that for the horizontal review. Once WebID is merged, I'll ping the different parts of W3C to get the review started.
```

TallTed (changes requested):

meetings/2023-06-21.md
csarven: Not sure I understand your reference to "technically this comment doesn't exist".. but I did say what's in the minutes in the meeting. If you meant your comment above, we briefly / informally? touched on that.

What's not captured in detail / the point was that it is CG's asset and that it should be controlled by the CG preferably under solidproject.org or solidcommunity.net. While some of those third-party services (github included) may be convenient to some - IMO - we/CG shouldn't rely on them (if anything but the negative impression it gives).

https://github.com/solid/solidproject.org/issues/248

meetings/2023-06-21.md
csarven: There is more to what's minuted here and perhaps the context wasn't entirely clear to everyone either. There is indeed the general read aspect of WebID Profiles (as per WebID 1.0) - which is what you're referring to - and then there is the more specific read/write aspect of Solid WebID Profile. Obviously we can't claim interoperability between Solid applications and arbitrary HTTP servers with respect to modifying the (Solid) WebID Profile Document. The Solid WebID Profile panel/team/force is trying to clarify this stuff.

meetings/2023-06-21.md
TallTed: I meant, technically [_my_ comment](https://github.com/solid/specification/pull/535#discussion_r1237827765) doesn't exist.

In reply to [your comment](https://github.com/solid/specification/pull/535#discussion_r1241298254) here, "under solidproject.org or solidcommunity.net" only covers a couple of URL authority segments, which could lead *anywhere* for the final resource. So far as I'm aware, the CG doesn't have any bank account, server hardware, or other material assets with which to "control" (or host) assets such as demo videos; thus, third-party hosting services are worth consideration (at least).

PR #538, opened 2023-06-29
Clarify that slash doesn't entail containement

### LINKED ISSUE #505: Are uri semantics just neccessary or sufficient to determine containment relation?
Spec says that:

> The slash (/) character in the URI path indicates hierarchical relationship segments, and enables relative referencing [[RFC3986](https://solidproject.org/ED/protocol#bib-rfc3986)]. The semantics of the slash character is shared by servers and clients. Paths ending with a slash denote a container resource.

### If it is `necessary`:

If above statement is expressing a `necessary` condition, then if two resources are `known` to be in container-child relation, their uri `MUST follow` (neccessary) follow above condition.

But vice-versa may not be true. I.e, if uris of two resources share hierarchical relation ship, then they `need not` be in containment relation, as the condition is `not sufficient`.

This indeed seems the case, that resource `with uri a/abc.acl`, need not be child of resource with uri of  `a/`, even though their uris are in hierarchical relation. But it instead may be aux of `a/abc`.

## If it is also `sufficient`:
Let's assume, there are three resorces:
1. `R1` with uri `/path/to/r1/`,
2. `R2` with uri `/path/to/r1/abc`,
3. `Rx` with uri `/path/to/r1/<slug>`.

Then if condition is also `sufficient`, one can conclude that `Rx` `MUST be child of `R1`, for they satisfy suffient condition for containment.

But we see this is not the case. We can see that if `<slug>` in `Rx` is `abc.acl`, then it may be instead acl of `R2` rather than child of `R1`.

## Problem

We above saw that, uri-slash-semantics are actually necessary for containment to hold, but not sufficient to unambiguously infer containment between to resources.

But we have been seeing in many places, the slash semantics beeing inconsistently treated as sufficient, and being used to infer containment when that may not be guaranteed.

For example:
If a container resource ``R1` with uri `/path/to/r1/` is known to exist, and resource Rx` with uri `/path/to/r1/<slug>` is is known to `not yet exist`, and a `GET` is sent for `Rx`, then should we consult `acl:default` of `R1`'s acl for determining status code to 403/404?. My sense says that, as resource `Rx` known to not exist yet, it is not child of any resource, as semantics are necessary but not sufficient. This is indeed the case if `<slug> may be `abc.acl`, it can be slot for aux of other instead. But many places in the issues, it is taken for granted that `R1`'s acl's `acl:default` will be applied to determine method access over `Rx`.

elf-pavlik:
The wording most likely can be improved. I aim to clarify that while URI Semantics require contained resources IRI must have a hierarchical path component, having a hierarchical component doesn't mean that the resource is contained. It can only be determined based on `ldp: contains` statements in the container description.

closes #505


damooo:
Thanks for that.

Here is the issue #505 .

It also raises few other dependent issues.

ED/protocol.html
```diff
@@ -686,6 +686,12 @@ <h3 property="schema:name">URI Slash Semantics</h3>
...
+                  <div class="note" id="slash-and-containment" inlist="" rel="schema:hasPart" resource="slash-and-containment">
+                    <h4 property="schema:name"><span>Note</span>: Slash and containement</h4>
+                    <div datatype="rdf:HTML" property="schema:description">
+                      <p>URI Slash Semantic on its own is not sufficient to determine containment. For example <code>/albums/garfield</code> does not entail <code>&lt;/albums/&gt; ldp:contains &lt;/albums/garfield&gt;</code></p>
```
woutermont, line 692: The current wording (_"URI Slash Semantic on its own is not sufficient to determine containment."_) is possibly confusing. After all, <ins>within the context of Solid</ins>, slash semantics ARE sufficient to determine containment (as of the current spec, which reads in the section on [Resource Containment](https://solidproject.org/TR/protocol#resource-containment): _"There is a 1-1 correspondence between containment triples and relative reference within the path name hierarchy."_). 

I would therefore insert the following, and possibly also move/copy the note to the Resource Containment section (in fact, it would probably be clearer if the Slash Semantics section was part of Resource Containment altogether, as that is where its impact lies.).

```suggestion
                      <p>Outside of the context of Solid, URI Slash Semantic on its own is not sufficient to determine containment. For example, without knowing whether the resource server adheres to Solid, <code>/albums/garfield</code> does not entail <code>&lt;/albums/&gt; ldp:contains &lt;/albums/garfield&gt;</code></p>
```

elf-pavlik:
>Here is the issue https://github.com/solid/specification/issues/505 .

Thank you @damooo !

I'm going to comment directly on that issue to answer the question first there and later just fine-tune the wording here.

ED/protocol.html
kjetilk: Yes, I'm :-1: on the original wording in this PR, but I think with the suggestion, I'm :+1: , as it is indeed a clarification that could be useful to some.

csarven: I find Wouter's suggestion to be generally a bit more useful here as a note. (I'm pressed on time to look at the exact location where it'd be best to put this in... but can review again after the change goes through, if any).

I'd suggest paraphrasing "outside of the context of Solid" into something more concrete, e.g., whether that's explicitly about the Solid Protocol (including its dependencies) or in particular the Storage space, or something else.

I'd also putting a bit more emphasis / clarification / distinction on the expectations/interpretations (if any) from the identifier vs. what might a representation of it describe it to be.

Pavlik, since you originally proposed, if you're on board with these, and other suggestions, could you like to or have Wouter give it another pass at the note?

elf-pavlik:
During the meeting we agreed to continue in #505 and based on resolution create new PR

PR #558, opened 2023-08-20
Fix: update SAI primer link

woutermont:
Going through the TRs I noticed the link to the SAI primer resulted in a 404. I believe this has not been updated since SAI split the primer into separate parts aimed at the different conformance classes. Can you confirm this, @elf-pavlik?

I also corrected the status of the primer(s), which said ED but is in fact UD according to the documents themselves.

Target | Correction class | Note
-- | -- | --
#work-item-technical-reports | [2](https://www.w3.org/2021/Process-20211102/#class-2) | update broken link
#work-item-technical-reports | [2](https://www.w3.org/2021/Process-20211102/#class-2) | correct displayed draft status

_(The TR table does not containe more finegrained IDs, so could not put more detail in correction table above.)_

[HTML Preview](https://htmlpreview.github.io/?https://github.com/solid/specification/blob/95d3904794ebef4c1f7e66d33299a899f889dd86/index.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -256,7 +256,7 @@ border-bottom: 2pt solid #000;
                     <td><abbr title="To Be Determined">TBD</abbr></td>
                   </tr>
                   <tr>
-                    <td><a href="https://solid.github.io/data-interoperability-panel/primer/" rel="cito:citesForInformation">Solid Application Interoperability Primer</a></td>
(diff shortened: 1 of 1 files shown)


elf-pavlik (approved):
Thank you @woutermont !

csarven (approved):

Issue #22, opened 2019-07-31
Add URI Normalization


kjetilk:
URI Normalization is needed for consistency when comparing URIs.

RFC3986 gives some guidelines, but it is hardly enough, as we already noticed with the http vs https debate that we had because of breakage with vocabs that were being loaded. It is also extremely important for components that do any kind of query that they agree on how to tell if to URIs refer to the same resource. Another field where it is important is with HTTP proxies, if one proxy has a different idea of a resource than the server behind it, it could lead to very strange bugs, even security failures if the proxy and the server doesn't agree on what ACL applies to which resource.

We cannot eliminate false negatives but we should minimize their impact by being strict when we can.

RubenVerborgh:
Agreed; we should be careful though given that RDF considers equivalent but syntactically different IRIs as entirely different IRIs.

csarven:
Case: https://github.com/solid/web-access-control-spec/issues/93

kjetilk:
> Agreed; we should be careful though given that RDF considers equivalent but syntactically different IRIs as entirely different IRIs.

Indeed, RDF is very strict on this point, perhaps overly so ([link for easy reference](https://www.w3.org/TR/rdf11-concepts/#section-IRIs)). 

This makes it a somewhat difficult problem to address on the spec side, as we need to ensure that implementations do normalization before it enters an RDF representation.

csarven:
https://github.com/solid/vocab/issues/20#issuecomment-1111993208

woutermont:
@csarven, I propose to take this issue (and the related https://github.com/solid/specification/issues/347) up in the milestone for v0.11. As long as we specify Solid's handling of identifiers, there should be no problem, so it is important to do so.

I will do a proposal based on the relevant specs and the conversation table sketch of #347, to get the conversation going again.

PR #575, opened 2023-09-29
i18n and n11n of resource identifiers

woutermont:
This PR updates the Solid protocol with a normative section on the use of Internationalized Resource Identifiers (IRIs).

It attempts to address the following issues (and possibly other related ones):

- https://github.com/solid/specification/issues/347
- https://github.com/solid/specification/issues/22
- https://github.com/solid/web-access-control-spec/issues/93

The gist of it is as follows:

- Require Solid servers to only mint IRIs for their resources according to a normalized form.
- When writing HTTP messages metadata, transform IRI to URI.
- When reading HTTP messages metadata, transform URI to IRI and normalize.
- Always leave the data itself (content) alone.

What I did not (yet) do: 

- Change mentions of URI to IRI throughout existing text.
- Adhere to [W3C i18n best practices](https://www.w3.org/TR/international-specs/#resid_misc): WHATWG URL and UTF-8 as additional restrictions.

### Changes
Commit | Target | Correction class | Note
-- | -- | -- | --
65df99b | #uri | 2 | rename to #identifiers ("Resource Identifiers")
65df99b | #identifiers | 2 | rename from #uri ("Uniform Resource Identifier")
65df99b | #toc | 1 | toc update
98cad8b | #bib-rfc3987 | 1 | add rfc3987 to normative references
5974359 | #bib-rfc3987 | 1 | add rfc3987 to normative references
e679fe0 | #iri | 2 | add iri definition to terminology
1e5b0bd | #bib-uax15 | 1 | add uax15 to normative references
dd2c455 | #bib-ascii | 1 | add ISO/IEC 646:1991 to informative references
b12fa2e | #uniform-resource-identifier | 2 | update uri definition in terminology
6a225a0 | #iris | 4 (new requirement) | add section requiring IRIs
6a225a0 | #toc | 1 | toc update

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -1526,6 +1526,8 @@ margin-top:1em;
                     <dd><a href="https://www.w3.org/TR/sparql11-query/" rel="cito:citesAsAuthority"><cite>SPARQL 1.1 Query</cite></a>. Steve Harris; Andy Seaborne; Eric Prud'hommeaux.  W3C. 21 March 2013. W3C Recommendation. URL: <a href="https://www.w3.org/TR/sparql11-query/">https://www.w3.org/TR/sparql11-query/</a></dd>
                     <dt id="bib-turtle">[Turtle]</dt>
                     <dd><a href="https://www.w3.org/TR/turtle/" rel="cito:citesAsAuthority"><cite>RDF 1.1 Turtle</cite></a>. Eric Prud'hommeaux; Gavin Carothers.  W3C. 25 February 2014. W3C Recommendation. URL: <a href="https://www.w3.org/TR/turtle/">https://www.w3.org/TR/turtle/</a></dd>
+                    <dt id="bib-uax15">[UAX15]</dt>
+                    <dd><a href="https://www.unicode.org/reports/tr15/" rel="cito:citesAsAuthority"><cite>Unicode normalization Forms</cite></a>. K. Whistler, Ed..  The Unicode Consortium. 12 August 2023. Unicode Standard Annex #15. URL: <a href="https://www.unicode.org/reports/tr15/">https://www.unicode.org/reports/tr15/</a></dd>
                     <dt id="bib-w3c-html">[W3C-HTML]</dt>
                     <dd><a href="https://www.w3.org/TR/html/" rel="cito:citesAsAuthority"><cite>HTML</cite></a>.  W3C. 28 January 2021. W3C Recommendation. URL: <a href="https://www.w3.org/TR/html/">https://www.w3.org/TR/html/</a></dd>
                     <dt id="bib-wac">[WAC]</dt>
--- a/protocol.html
+++ b/protocol.html
@@ -502,7 +502,7 @@ margin-top:1em;
                     <dd about="#iri" property="skos:definition">An <dfn>Internationalized Resource Identifier</dfn> (<abbr title="Internationalized Resource Identifier">IRI</abbr>) is an identifier as defined in [<cite><a class="bibref" href="#bib-rfc3987">RFC3987</a></cite>], which identifies an entity in a global context, using a Unicode character sequence [<cite><a class="bibref" href="#bib-unicode">UNICODE</a></cite>].</dd>
 
                     <dt about="#uniform-resource-identifier" property="skos:prefLabel" typeof="skos:Concept"><dfn id="uniform-resource-identifier">URI</dfn></dt>
-                    <dd about="#uniform-resource-identifier" property="skos:definition">A <dfn>Uniform Resource Identifier</dfn> (<abbr title="Uniform Resource Identifier">URI</abbr>) provides the means for identifying resources [<cite><a class="bibref" href="#bib-rfc3986">RFC3986</a></cite>].</dd>
+                    <dd about="#uniform-resource-identifier" property="skos:definition">A <dfn>Uniform Resource Identifier</dfn> (<abbr title="Uniform Resource Identifier">URI</abbr>) is an identifier as defined in [<cite><a class="bibref" href="#bib-rfc3987">RFC3986</a></cite>], which identifies an entity in a global context, using an ASCII character sequence [<cite><a class="bibref" href="#bib-ascii">ASCII</a></cite>]. Every URI is also an IRI.</dd>
 
                     <dt about="#resource" property="skos:prefLabel" typeof="skos:Concept"><dfn id="resource">resource</dfn></dt>
                     <dd about="#resource" property="skos:definition">A resource is the target of an HTTP request identified by a URI [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>].</dd>
--- a/protocol.html
+++ b/protocol.html
@@ -498,6 +498,9 @@ margin-top:1em;
                     <dt about="#solid-app" property="skos:prefLabel" typeof="skos:Concept"><dfn id="solid-app">Solid app</dfn></dt>
                     <dd about="#solid-app" property="skos:definition">A Solid app is an application that reads or writes data from one or more <a href="#storage">storages</a>.</dd>
 
+                    <dt about="#iri" property="skos:prefLabel" typeof="skos:Concept"><dfn id="iri">IRI</dfn></dt>
+                    <dd about="#iri" property="skos:definition">An <dfn>Internationalized Resource Identifier</dfn> (<abbr title="Internationalized Resource Identifier">IRI</abbr>) is an identifier as defined in [<cite><a class="bibref" href="#bib-rfc3987">RFC3987</a></cite>], which identifies an entity in a global context, using a Unicode character sequence [<cite><a class="bibref" href="#bib-unicode">UNICODE</a></cite>].</dd>
+
                     <dt about="#uniform-resource-identifier" property="skos:prefLabel" typeof="skos:Concept"><dfn id="uniform-resource-identifier">URI</dfn></dt>
                     <dd about="#uniform-resource-identifier" property="skos:definition">A <dfn>Uniform Resource Identifier</dfn> (<abbr title="Uniform Resource Identifier">URI</abbr>) provides the means for identifying resources [<cite><a class="bibref" href="#bib-rfc3986">RFC3986</a></cite>].</dd>
 
--- a/protocol.html
+++ b/protocol.html
@@ -1556,6 +1556,8 @@ margin-top:1em;
                     <dt id="bib-solid-websockets-api">[SOLID-WEBSOCKETS-API]</dt>
                     <dd><a href="https://github.com/solid/solid-spec/blob/master/api-websockets.md" rel="cito:citesAsPotentialSolution"><cite>Solid WebSockets API</cite></a>. Nicola Greco; Dmitri Zagidulin; Ruben Verborgh.  W3C Solid Community Group. 17 June 2020. Unofficial Draft. URL: <a href="https://github.com/solid/solid-spec/blob/master/api-websockets.md">https://github.com/solid/solid-spec/blob/master/api-websockets.md</a></dd>
                     <dt id="bib-uaag20">[UAAG20]</dt>
+                    <dd><a href="https://www.unicode.org/versions/latest/" rel="cito:citesAsPotentialSolution"><cite>The Unicode Standard</cite></a>. The Unicode Consortium.  URL: <a href="https://www.unicode.org/versions/latest/">https://www.unicode.org/versions/latest/</a></dd>
+                    <dt id="bib-unicode">[UNICODE]</dt>
                     <dd><a href="https://www.w3.org/TR/UAAG20/" rel="cito:citesAsPotentialSolution"><cite>User Agent Accessibility Guidelines (UAAG) 2.0</cite></a>. James Allan; Greg Lowney; Kimberly Patch; Jeanne F Spellman.  W3C. 15 December 2015. W3C Note. URL: <a href="https://www.w3.org/TR/UAAG20/">https://www.w3.org/TR/UAAG20/</a></dd>
                     <dt id="bib-wai-aria-1.2">[WAI-ARIA-1.2]</dt>
                     <dd><a href="https://www.w3.org/TR/wai-aria-1.2/" rel="cito:citesAsPotentialSolution"><cite>Accessible Rich Internet Applications (WAI-ARIA) 1.2</cite></a>. Joanmarie Diggs; James Nurthen; Michael Cooper.  W3C. 2 March 2021. W3C Candidate Recommendation. URL: <a href="https://www.w3.org/TR/wai-aria-1.2/">https://www.w3.org/TR/wai-aria-1.2/</a></dd>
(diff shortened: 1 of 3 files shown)


protocol.html
```diff
@@ -1554,6 +1593,8 @@ <h3 property="schema:name">Informative References</h3>
...
                     <dd><a href="https://github.com/solid/solid-spec/blob/master/api-websockets.md" rel="cito:citesAsPotentialSolution"><cite>Solid WebSockets API</cite></a>. Nicola Greco; Dmitri Zagidulin; Ruben Verborgh.  W3C Solid Community Group. 17 June 2020. Unofficial Draft. URL: <a href="https://github.com/solid/solid-spec/blob/master/api-websockets.md">https://github.com/solid/solid-spec/blob/master/api-websockets.md</a></dd>
                     <dt id="bib-uaag20">[UAAG20]</dt>
+                    <dd><a href="https://www.unicode.org/versions/latest/" rel="cito:citesAsPotentialSolution"><cite>The Unicode Standard</cite></a>. The Unicode Consortium.  URL: <a href="https://www.unicode.org/versions/latest/">https://www.unicode.org/versions/latest/</a></dd>
+                    <dt id="bib-unicode">[UNICODE]</dt>
                     <dd><a href="https://www.w3.org/TR/UAAG20/" rel="cito:citesAsPotentialSolution"><cite>User Agent Accessibility Guidelines (UAAG) 2.0</cite></a>. James Allan; Greg Lowney; Kimberly Patch; Jeanne F Spellman.  W3C. 15 December 2015. W3C Note. URL: <a href="https://www.w3.org/TR/UAAG20/">https://www.w3.org/TR/UAAG20/</a></dd>
```
csarven, line 1598:
```suggestion
                    <dd><a href="https://www.w3.org/TR/UAAG20/" rel="cito:citesAsPotentialSolution"><cite>User Agent Accessibility Guidelines (UAAG) 2.0</cite></a>. James Allan; Greg Lowney; Kimberly Patch; Jeanne F Spellman.  W3C. 15 December 2015. W3C Note. URL: <a href="https://www.w3.org/TR/UAAG20/">https://www.w3.org/TR/UAAG20/</a></dd>
                    <dt id="bib-unicode">[UNICODE]</dt>
                    <dd><a href="https://www.unicode.org/versions/latest/" rel="cito:citesAsPotentialSolution"><cite>The Unicode Standard</cite></a>. The Unicode Consortium.  URL: <a href="https://www.unicode.org/versions/latest/">https://www.unicode.org/versions/latest/</a></dd>
```

protocol.html
```diff
@@ -682,6 +686,35 @@ <h3 property="schema:name"><span>Note</span>: Storage Owner and URI Ownership</h
...
+                    <li id="iris-hex">Hexadecimal digits within percent-encoding triplets corresponding to <em>reserved</em> characters are represented using <em>uppercase</em> letters.</li> 
+
+                    <li id="iris-scheme">The <code>http</code> or <code>https</code> scheme of the IRI is represented using <em>lowercase</em> characters.</li>
+
+                    <li id="#iris-host">The host of the IRI is represented using <em>lowercase</em> characters.</li>
+
+                    <li id="iris-port">If the port of the IRI is the default port for its scheme, the port subcomponent is left out.</li>
+
+                    <li id="iris-path">The path of the IRI does not contain dot-segments.</li>
+                  </ul>
```
csarven, line 708: What do you think about leaving this out? Would conforming to 3987's sections 3.1-3.2 suffice?

```suggestion
```

protocol.html
```diff
@@ -682,6 +686,35 @@ <h3 property="schema:name"><span>Note</span>: Storage Owner and URI Ownership</h
...
+
+                  <p><span about="" id="server-iris-norm" rel="spec:requirement" resource="#server-iris-norm"><span property="spec:statement">A Solid <span rel="spec:requirementSubject" resource="#Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> create IRIs that do not conform to this normal form.</span></span></p>
```
csarven, line 710: When a server conforms to 3987's sections 3.1-3.2 (`#server-iris-to-http` , `#server-iris-from-http`), does it satisfy the `#server-iris-norm` requirement? If so, can `#server-iris-norm` be removed?

```suggestion
```

csarven (commented):
Thank you, good stuff! I think this PR captures the essence and needs of the referenced issues. As I understand it, this PR (or in this direction) would satisfy https://www.w3.org/TR/international-specs/#resid_what_to_spec .

It would be good to document some implementation experience. What currently conforms to this? How complex is it implement?

Is there a reason to introduce additional restrictions to WHATWG URL and UTF-8 ( https://www.w3.org/TR/international-specs/#resid_what_to_spec_protocol mentions )?

protocol.html
woutermont: As clarified in reply to your other comment, it does not in all cases.

woutermont:
> As I understand it, this PR (or in this direction) would satisfy https://www.w3.org/TR/international-specs/#resid_what_to_spec.

Yes and no. It is compliant with the therein prefered specs, but our text does not mention WHATWG URL, which the iSpecs document demands. I have not done so because it really does not help clarity at all.

damooo:
> It attempts to address the following issues (and possibly other related ones):
>
> - https://github.com/solid/specification/issues/347
> - https://github.com/solid/specification/issues/22
> - https://github.com/solid/web-access-control-spec/issues/93

There are two orthogonal issues here. One is about normalization of identifiers, and other is about allowing iris as identifiers for solid resources.

Though it is important to address identifier-normalization, it is not going to be correct/unambiguous/practical to allow iris as identifiers for solid resources. There are many issues. I will specify a few.

# 1. Incompatible with http & web tools

Solid protocol is **a profile** of http. Http allows only uris for resources. Thus, as all Solid resources are http resources, their identifiers must also be uris. Otherwise Solid will not be compatible with much of the web and most of the tools.

For example any generic link traverser (like rdflib/tpf's/communica/...) don't kmow these random mappings, and directly use uri as the identifier for resource they fetched.

Many tools like rdf parsers, jsonld processors use http resource uri as the base uri for processing representations. Thus it is going to be nightmare to deal with all existing generic tools.

Basically, if we cross the semantics of http, solid is no longer going to be a profile of it, and none of the http tools work seamlessly.

# 2. Impossible to specify

Essentially, it is not possible to specify about these mappings with sufficient rigour. I will specify few fundamental issues in proposed spec for example.

Proposed: 
> When interpreting a URI in an [HTTP](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#http) message, except in the content, as a resource identifier, a Solid server MUST map the URI to an IRI according to the algorithm provided by [[RFC3987](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#bib-rfc3987)] ([section 3.2](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#section-3.2)), and normalize the resulting IRI to the [normal form](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#iris-norm) provided in this section.

First of all, resource servers **doesn't even recieve a URI** for target resource in a http request message. And what server receives as part of message is also varies between http versions. A request essentially contains only [request-target](https://www.rfc-editor.org/rfc/rfc9112#section-3.2) in a message in one of allowed forms (asterik-/origin-/absolute-/authority- forms). It is server's work to reconstruct final resource uri using request-target hint and it's own ** custom routing config**, as per its discretion. As rfc 9110 says:

> Upon receipt of a client's request, a server reconstructs the target URI from the received components in accordance with their local configuration and incoming connection context. This reconstruction is specific to each major protocol version. For example, [Section 3.3](https://www.rfc-editor.org/rfc/rfc9112#section-3.3) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110.html#HTTP11)] defines how a server determines the target URI of an HTTP/1.1 request.

Thus resolved resource uri will never be part of the message, and thus proposed statement has no effect.

Along with that, there can be relative uri-references, which are not uris.

And the most serious issue is, given a uri in a message (e.g in `Link` headers), we **don't know if it refers to a solid resource or not**. Thus we can never be sure whether to apply such a transform.

These issues are just on the surface, and already breaks a lot. Thus it may be good, if we address only uri-normalization **restriction** part only.

@woutermont, @csarven

jeff-zucker:
A somewhat related issue.  The [WebID 1.0 spec](https://www.w3.org/2005/Incubator/webid/spec/identity/) says this : 

> A WebID is a URI with an HTTP or HTTPS scheme  ...". 

 Would that need to be modified to accomodate IRIs?

woutermont:
Thanks @damooo, for the elaborate comment, which raises some good questions. I will try to address them all below.

> [@damooo:] There are two orthogonal issues here. One is about normalization of identifiers, and other is about allowing iris as identifiers for solid resources.

True, there are two issues here; but they are not orthogonal. The choice for URI, IRI and/or WHATWG URL as identifiers is tightly bound with the normalization procedures, since it is the specifications of these identifiers that specify the possible procedures.

---

### Compatibility with HTTP

> [@damooo:] Solid protocol is **a profile** of http. Http allows only uris for resources. Thus, as all Solid resources are http resources, their identifiers must also be uris. Otherwise Solid will not be compatible with much of the web and most of the tools.

To be entirely correct, there is no such thing as an "HTTP resource" (apart from the trivial fact that HTTP URIs are URIs, which refer to resources). 

But to address your concerns about compatibility: any resource can be identified by more than one identifier. This is how the IRI specification maintains compatibility with URI: it provides a unique mapping from IRIs to URIs. Every HTTP IRI therefore has an HTTP URI that refers to the same resource.
 
> [@damooo:] For example any generic link traverser (like rdflib/tpf's/communica/...) don't kmow these random mappings, and directly use uri as the identifier for resource they fetched. [...] Many tools like rdf parsers, jsonld processors use http resource uri as the base uri for processing representations. Thus it is going to be nightmare to deal with all existing generic tools.

Given the existance of the mapping, as explained above, this danger is avoided: HTTP programs can simply continue to use the corresponding (mapped) URI. They do not need to know anything about IRI or the IRI<->URI mapping. It is the IRI-capable server that does the mapping upon entry and exit of messages.

> [@damooo:] Basically, if we cross the semantics of http, solid is no longer going to be a profile of it, and none of the http tools work seamlessly.

As concluded above, this is not true. Moreover, I would contend the idea that "Solid is a profile of HTTP." A specification profile is an extension point defined by the specification itself (see the [W3C Note on Variability in Specifications](https://www.w3.org/TR/spec-variability/#subdivision-profile)). HTTP, however, does not define such an extensoin point; "HTTP profiles" do not exist. Solid is a separate, standalone specification that _uses_ HTTP as a communication protocol. Given the compatibility of HTTP and IRI, as sketched above, it is thus perfectly possible to combine that with using IRI as an identification scheme.

---

### URI<->IRI mapping

> > [@woutermont (PR proposal):] When interpreting a URI in an [HTTP](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#http) message, except in the content, as a resource identifier, a Solid server MUST map the URI to an IRI according to the algorithm provided by [[RFC3987](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#bib-rfc3987)] ([section 3.2](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#section-3.2)), and normalize the resulting IRI to the [normal form](https://htmlpreview.github.io/?https%3A%2F%2Fraw.githubusercontent.com%2Fwoutermont%2Fsolid-specification%2Fresource-identifiers%2Fprotocol.html#iris-norm) provided in this section.
> 
> [@damooo:] First of all, resource servers **doesn't even recieve a URI** for target resource in a http request message. [...] A request essentially contains only [request-target](https://www.rfc-editor.org/rfc/rfc9112#section-3.2) in a message in one of allowed forms (asterik-/origin-/absolute-/authority- forms). It is server's work to reconstruct final resource uri using request-target hint and it's own **custom routing config**, as per its discretion. [...] Thus resolved resource uri will never be part of the message, and thus proposed statement has no effect.

What you describe is correct: an HTTP server reconstructs the URI of the target resource based on some algorithm (that we can treat as a blackbox here). However, this does not contradict my proposal (although it points out that it should probably be rephrased): when I say "a URI in an HTTP message", I do mean "the URI of the target resource, as reconstructed by the server." According to my proposal, given that URI, a Solid server should then immediately map it to its corresponding IRI, thereby forming the boundary between the URI-based communication channel (i.e. HTTP) and the IRI-based resource storage (i.e. Solid).

> [@damooo:] Along with that, there can be relative uri-references, which are not uris.

The 'blackbox' algorithm the server applies to identify the target resource will always result in an URI; i.e. the algorithm _includes_ the logic for going from a URI reference to a URI (typically by prepending a base URI). This is described in [RFC 9110 § 7.1 Determining the target resource](https://www.rfc-editor.org/rfc/rfc9110.html#name-determining-the-target-reso) (¶ 2).
 
> [@damooo:] [G]iven a uri in a message (e.g in `Link` headers), we **don't know if it refers to a solid resource or not**. Thus we can never be sure whether to apply such a transform.

This is true, but the proposal only applies to the target URI. Any other URI in the request (Link headers, redirects etc.) is left alone, since they will typically be used in subsequent HTTP requests. It is not the job of Solid to interfere in that.

--- 

### Compatibility with WebID

I will also shortly address @jeff-zucker's question:

> [@jeff-zucker:] The [WebID 1.0 spec](https://www.w3.org/2005/Incubator/webid/spec/identity/) says this :
> 
> > A WebID is a URI with an HTTP or HTTPS scheme ...".
> 
> Would that need to be modified to accomodate IRIs?

No, that is not necessary. Since any URI is also an IRI, WebIDs can still be Solid resources. For some time, WebIDs will then not benefit of the extension of Solid to IRI, but I suspect that eventually WebID will also make the move, so that WebIDs can also include international characters. In any case, using IRIs in Solid will not break existing usage of WebIDs.

--- 

Hope to have addressed your issues/questions adequately. They shed some missing clarity that I will try to add to the phrasing of the proposal. Thanks for that.

jeff-zucker:
>> Would that [WebID 1.0] need to be modified to accomodate IRIs?

> No, that is not necessary. Since any URI is also an IRI, WebIDs can still be Solid resources. For some time, WebIDs will then not benefit of the extension of Solid to IRI, but I suspect that eventually WebID will also make the move, so that WebIDs can also include international characters. In any case, using IRIs in Solid will not break existing usage of WebIDs.

But since  the Solid WebID Profile spec does not define a WebID,  rather counts on WebID 1.0 to define it, we would not be able to remain spec compliant if we use IRIs in WebIDs.

woutermont:
> But since the Solid WebID Profile spec does not define a WebID, rather counts on WebID 1.0 to define it, we would not be able to remain spec compliant if we use IRIs in WebIDs.

No. We we would only break compliance if we use non-URI IRIs as a WebID. So we simply continue to mint only URIs for WebIDs, until WebID also supports IRIs. This is not an issue.

damooo:
> Given the existance of the mapping, as explained above, this danger is avoided: HTTP programs can simply continue to use the corresponding (mapped) URI. They do not need to know anything about IRI or the IRI<->URI mapping. It is the IRI-capable server that does the mapping upon entry and exit of messages.

That is an issue. Thus, all the http tools determine the uri as the identifier.

And they will use that as the base uri, to resolve the resource representations. Thus any relative identifiers in a rdf resource will be resolved against the uri by these clients. While server uses the iri for the resolution. This will soon lead to chaos, as clients also start writing. For a single resource, some places an iri will be used, some places a uri. Some places absolute, some places relative. We cannot deterministically do anything with that data. This is what i meant to be incompatible.

> This is true, but the proposal only applies to the target URI. Any other URI in the request (Link headers, redirects etc.) is left alone, since they will typically be used in subsequent HTTP requests. It is not the job of Solid to interfere in that.

That is an issue. Solid uses link headers to link to other resources (acls, descriptions, storage root, ....). If their identifiers are iris, there is no way to send them in the links, as link headers accept only uris. If we encode as uris and send, then client will deduce the uri as identifier for acls, etc. and uses them when writing statements. While server using iris. Repeat of the above problem. With severe security consequences.

elf-pavlik (approved):
The proposed text looks equivalent to https://www.w3.org/TR/rdf12-concepts/#section-IRIs

Since Solid heavily builds on RDF, we should closely follow the latest specifications.

michielbdejong (approved):

PR #598, opened 2023-11-22
Add security consideration for serving user-created files

### LINKED ISSUE #514: Security issues because of serving html files
Currently Solid pods serve html files which get rendered in the browser. This breaks one base assumption of web security, namely that html files from the same origin are equally trustworthy, and leads to various potential security issues. From my point of view this should be addressed in the specification, to ensure the pods have the same behaviour in this regard.

## Security issue

In web security it is assumed that apps on the same origin are equally trustworthy. Many security mechansisms block cross-origin communication and access, but within the same origin apps are less limited. Citing [MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy): "The same-origin policy is a critical security mechanism that restricts how a document or script loaded by one origin can interact with a resource from another origin.".

Solid pods break the assumption that apps on the same origin are equally trustworthy. I could give someone access to `/music/index.html` and someone else to `/banking/index.html`. I do not want someone with access to `/music` to be able to access my banking details, hence I trust them in different ways. However, they are on the same origin and thus the browser does not prevent access between these apps. Frankly, we throw away the "critical security mechanism".

EDIT: this does not only apply to html files. For instance opening malicious svg files in the browser can execute javascript on the same origin.

## Example issues

The following issues arise from the fact that html files on the same pod share the same origin. The first ones require that the user already uses one app (e.g. `/banking/index.html`) and someone with access to `/drawing/index.html` creates a malicious app:

1. LocalStorage: this storage is shared within the same origin. The drawing app would be able to access stored data from the banking app.
2. Cookies: By default cookies use SameSite: Lax which means they are not send cross-origin, but are sent on the same origin. Requests from the drawing app would include cookies from the banking app. A real case is [this issue](https://github.com/nodeSolidServer/node-solid-server/issues/1719) where the Drawing would be able to use the NSS session cookies.
3. Iframes: The drawing app can include the banking app as an iframe. As it is same origin, it can access its contents. A real case is [this issue](https://github.com/SolidOS/solid-panes/issues/372) where Mashlib accidentically gives html files it displays access to its window.
4. Url history: the url is modifiable as long as the origin stays the same. Thus the drawing app could use `window.history.pushState('page2', 'Title', '/banking/index.html');` to trick the user into believing they are looking at the banking app.

The following example does not require any previously used app on the pod, ie it is a security issue even when there are currently no html files on the pod:

1. Service Workers: Service Workers are installed client side and then have control over all the network traffic within their scope. The scope is everything within the same folder or subfolders (eg `/foo/serviceWorker.js` can intercept requests to `/foo/**/*`). This can circumvent access control and other pod restrictions.

I’m pretty sure there are other cases, where it could be troublesome that apps of different trust-level are hosted on the same origin.

## Example usages

So far I have heard of @josephguillaume that he uses custom html files to add custom extensions to SolidOS (see [this issue](https://github.com/SolidOS/pane-registry/issues/28)). He also mentioned other possible usages in [this thread](https://forum.solidproject.org/t/is-it-secure-for-pods-to-serve-html-files/6379/4?u=a_a): direct use of web components, local add-ons as per the other thread, hosting apps yourself, not having to run a separate server. 

There are probably more unknown usages that rely on pods serving html files.

## Solution ideas

Following potential solutions were mentioned so far. Note that they are mostly intended as a starting idea, not fully thought through solution proposals.

1. Do not serve html files at all (ie nothing with `content-type: text/html`)
2. Sandbox html files, for instance with the [CSP sandbox](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox) header
3. Sanitize html, eg remove javascript (on save or when serving)
4. "placing additional restrictions on where public resources can be stored and how clients are allowed to interact with pods, etc." (by @joachimvh, not sure how this could look like)
5. (EDITED) Add the `Content-Disposition: attachment` header for all files, to ensure they are downloaded and not rendered in the browser

One thing to note is, that it is not only about html files, but anything that is served as text/html. For instance CSS can serve markdown files as html files with the same security implications.

## Related links

- Original discussion: https://forum.solidproject.org/t/is-it-secure-for-pods-to-serve-html-files/6379
- small discussion on CSS: https://github.com/CommunitySolidServer/CommunitySolidServer/issues/1596
- iframe XSS and some html-as-a-feature discussion: https://github.com/SolidOS/solid-panes/issues/372

Otto-AA:
Fixes #514 

This is a draft on how the security consideration could look like.

Note that there was not yet an agreement on a solution in the issue. However, as everyone agreed that it is an issue, I thought I'd propose this. As the implementers decide on the actual security measure, I also don't think we need to discuss this at length, and hopefully can add it soon.


michielbdejong:
This PR is on the [agenda for the 17 January CG Call](https://hackmd.io/1YFPpDmnR-GwpjTGugVXcw)

Otto-AA:
I've had a short talk with @csarven last year discussing this. I am still in favor of this PR (and would probably leave it as it is) and for sure this security issue needs to be addressed in some way or another. However, there are some legitimate concerns, as it would (or at least could) affect webpages hosted on Solid pods.

Here are some notes I made back then, maybe they are helpful. I don't think I will attend the meeting this week, so if you have questions, you can ask me earlier.

```
In general, what this PR implies:
Html files stored on pods may or may not get rendered like normal applications by the browser. This is, because pods may (and should) restrict those for security reasons. Details will likely depend on the pod provider, but users/developers cannot rely on html being served as normal applications on every pod. They would have to choose a specific pod that allows it, or move the webpage hosting to a different service.

If this is in contrast with an existing specification, then this would also need to be updated/clarified.

If the pod uses the CSP sandbox header with allow-scripts (like ESS does since some months):
Applications hosted on pods could still work. However, some stuff may break without additional "allow-*" fields (forms, popups, etc). So it would be a 2nd-level experience, but still enough for simple projects (rendering blog html, charts, etc).

Security-wise, they would be treated as if they are on a different origin, thus the problems mentioned in the github issue are resolved.

If the pod uses the CSP sandbox header without allow-scripts:
Applications hosted on pods could only render html with no javascript, so very little interaction. I guess most of them would need to move to different

In both cases: phishing would still be possible (eg making a "[myaccount.solidcommunity.net](http://myaccount.solidcommunity.net/)" pod which could show a malicious webpage asking them download evil.exe, etc). This would allow phishing websites having a better domain. The sandbox header is no solution to phishing, this would need a different solution (eg the Content-Disposition: attachment header).

Regarding Whitelisting applications:

The pod provider can whitelist any html that they control (eg SolidOS as default application), but should not whitelist applications created by users.

The problem with whitelisting user applications is, that they are on the same "site" as other attackable content (other user pods, or account pages). While this gives less capabilities to an attacker than same-origin, same-site still has advantages. For instance, saved passwords can also filled on different subdomains from the same site (depending on the browser or browser extensions and their settings). As a user I could ask the pod owner to whitelist "myblog.html" and then send it to other users and phish for saved credentials.

This could be prevented if the pod registers for the https://publicsuffix.org/ (like [github.io](http://github.io/)), however I don't think we should consider this as a normal case.
```

TallTed:
A more immediately readable rendition (wrapped in `<blockquote>` instead of ` ``` `) of @Otto-AA's [notes](https://github.com/solid/specification/pull/598#issuecomment-1892656519) --

<blockquote>In general, what this PR implies:
Html files stored on pods may or may not get rendered like normal applications by the browser. This is, because pods may (and should) restrict those for security reasons. Details will likely depend on the pod provider, but users/developers cannot rely on html being served as normal applications on every pod. They would have to choose a specific pod that allows it, or move the webpage hosting to a different service.

If this is in contrast with an existing specification, then this would also need to be updated/clarified.

If the pod uses the CSP sandbox header with allow-scripts (like ESS does since some months):
Applications hosted on pods could still work. However, some stuff may break without additional "allow-*" fields (forms, popups, etc). So it would be a 2nd-level experience, but still enough for simple projects (rendering blog html, charts, etc).

Security-wise, they would be treated as if they are on a different origin, thus the problems mentioned in the github issue are resolved.

If the pod uses the CSP sandbox header without allow-scripts:
Applications hosted on pods could only render html with no javascript, so very little interaction. I guess most of them would need to move to different

In both cases: phishing would still be possible (eg making a "[myaccount.solidcommunity.net](http://myaccount.solidcommunity.net/)" pod which could show a malicious webpage asking them download evil.exe, etc). This would allow phishing websites having a better domain. The sandbox header is no solution to phishing, this would need a different solution (eg the Content-Disposition: attachment header).

Regarding Whitelisting applications:

The pod provider can whitelist any html that they control (eg SolidOS as default application), but should not whitelist applications created by users.

The problem with whitelisting user applications is, that they are on the same "site" as other attackable content (other user pods, or account pages). While this gives less capabilities to an attacker than same-origin, same-site still has advantages. For instance, saved passwords can also filled on different subdomains from the same site (depending on the browser or browser extensions and their settings). As a user I could ask the pod owner to whitelist "myblog.html" and then send it to other users and phish for saved credentials.

This could be prevented if the pod registers for the https://publicsuffix.org/ (like [github.io](http://github.io/)), however I don't think we should consider this as a normal case.</blockquote>

csarven (commented):
Let me first say that I appreciate the careful consideration here a lot. I also think we should discuss and work through it further before introducing something like this into the specification, even as a consideration. It seems to introduce other complications and makes some assumptions about the overall design that's not necessarily true or desired.

(Let's use terminology like allow-listing instead of white-listing. Otto, I'd appreciate it you can update your comment along those lines.)

First, the term "pod provider" doesn't align with existing notions. Relevant notions in use are: identity provider, server, storage owner, and URI owner.

According to the Web architecture, the URI owner allocates and delegates that intended behaviour to the server. That's orthogonal to reserved-allocations by the specification (like for instance the container, root container) which is assumed to be agreed by the URI owner the moment they use a product (server) that conforms to the specification. That said, the suggestion in this PR seems to introduce some discrepancy because it gives more authority to the server (or perhaps the storage owner?) than the URI owner with such constraints.

The allow/blocklisting suggestion introduces a gap in the flow and specification as it doesn't specify how a URI owner can instruct the server to allow specific resources to include or exclude the CSP header (or other, which is also part of a broader topic discussed elsewhere.) And so achieving that functionality would be implementation specific / unique to each vendor until something specifies how. This could be addressed by introducing information in the original HTTP request - think along the lines of an "interaction model". If that's not already specified somewhere (in the context of CSP headers) that we can refer to, we may need to specify it. That aside, I would suggest to reach a better understanding in this space before suggesting servers to block off certain things. That said, any server already has the option to include certain header irrespective to the proposed consideration here.

There is a wide range of content and user interactions that use markup languages as the host language. Those kinds of expressions are integral to the resource, and blocking them off at the gate dismisses a major category of use cases that's already out there on the Web. I don't think this should be simply hand-waved or expect what people currently do to make drastic changes when they try to onboard to Solid. If anything, Solid should be acknowledging their existence and allow a smooth transition from their existing servers and publishing practises to Solid. A similar argument holds for Solid applications being able to read-write a plain ol' "homepage".

One other assumption here is how applications could be introduced to the Solid storage in the first place. Being able to "spawn" instances of an application using the Solid Protocol is an amazing feature. Telling the world that they should use non-Solid servers to host their applications also seems awkward and counter, and that's irrespective to the applications that may currently be using non-Solid servers to host their applications.

michielbdejong:
Good point @csarven, maybe worth rewording it a bit to make that clear.
Let's merge this and iterate on it if we have improvements?

VirginiaBalseiro:
> Good point @csarven, maybe worth rewording it a bit to make that clear. Let's merge this and iterate on it if we have improvements?

This was not consensus from the meeting. We agreed to have more discussion/review on  the PR, not to merge and reiterate.

elf-pavlik:
If we can't agree on the current text, we should still add an inline issue linking the original issue from the specification. 
I'm thinking equivalent of [Remote Issues in Bikeshed](https://speced.github.io/bikeshed/#remote-issues)

elf-pavlik:
> For smaller servers (eg the personal own server), maybe it would be secure to allow serving user-created applications. However, even there, if you eg use the current CSS recipe with mashlib, an evil agent/application with access to your personal server may be able to make requests in the name of you to other servers. So the smaller server could impact other servers through your WebID (given some circumstances I can try to explain in case someone is interested).

I would appreciate it if you could further elaborate on this point.

Otto-AA:
> > For smaller servers (eg the personal own server), maybe it would be secure to allow serving user-created applications. However, even there, if you eg use the current CSS recipe with mashlib, an evil agent/application with access to your personal server may be able to make requests in the name of you to other servers. So the smaller server could impact other servers through your WebID (given some circumstances I can try to explain in case someone is interested).

> I would appreciate it if you could further elaborate on this point.

I'll explain two attack scenarios, there could be more as it's a complex topic. The attacker can be an agent with a WebID or an application. They need append/write access to your server to be able to store the malicious html file, eg because you allowed them to write a blog post, or they even have their own storage in a different path on the same domain.

1. Use SolidOS's DPoP tokens to make authenticated requests

Prerequesites:
- The attacker has append/write access to a publicly readable folder/file on the server
- The server serves SolidOS under the same domain (in CSS everything is served under the same domain per default)
- The victim is logged in via SolidOS, so opening it automatically logins again

Attack:
The attacker writes a malicious html file to the server. When this file is opened by the user:
- It opens SolidOS in a new tab and save the window reference (iirc, opening any non-existing resource from the server returns SolidOS)
- SolidOS automatically logins, as the user made a login there previously
- The html file can access SolidOS via the window reference, including the authenticated fetch
- The html file can make any requests through this authenticated fetch in the name of the currently logged-in user

2. Steal login credentials

Prerequesites:
- The attacker has append/write access to a publicly readable folder/file on the server
- The victim uses the IDP of the (small) CSS server
- The storages are on the same domain as the IDP (default for CSS)
- The user saved their login credentials (`/idp/login/` for CSS) via the browser or a password manager extension

Attack:
The attacker writes a malicious html file to the server. Depending on the application you used to store the login credentials, the concrete autofill/suggestion behaviour is different. But for instance, if the user saved it with Chrome and opens the malicious file:
- The victim need to make any interaction with the site (eg clicking on a cookie banner, whatever)
- Chrome automatically fills in `<input>` fields for the user name and password with the saved credentials
- The malicious html file can read and send the credentials to the attacker
- The attacker can use it to login with the IDP of the victim

Impact of both scenarios: The attacker can access anything that the victim has access to with this account.

elf-pavlik:
Thank you @Otto-AA 

> The attacker writes a malicious html file to the server. Depending on the application you used to store the login credentials, the concrete autofill/suggestion behaviour is different. But for instance, if the user saved it with Chrome and opens the malicious file

Does it only apply to the default CSS deployment where the OIDC Provider and the Resource Server are on the same origin? For example, I run two separate CSS instances on different subdomains, one for my OP and the second for my RS.  

> SolidOS automatically logins, as the user made a login there previously

I want to check my understanding with you.
I think that this relies on SolidOS having a cookie session with the OP and using the `prompt=none` to obtain the ID token via a silent redirect. How does the `redirect_uri` in SolidOS' ClientID document take the user back to the malicious HTML document?

Otto-AA:
> Does it only apply to the default CSS deployment where the OIDC Provider and the Resource Server are on the same origin? For example, I run two separate CSS instances on different subdomains, one for my OP and the second for my RS.

Yes, the 2nd attack only works if the OIDC Provider (and its login page where you save the credentials) and the Resource Server are on the same origin (That's what I meant with the "The victim uses the IDP of the (small) CSS server" precondition).

> > SolidOS automatically logins, as the user made a login there previously

> I want to check my understanding with you.
I think that this relies on SolidOS having a cookie session with the OP and using the prompt=none to obtain the ID token via a silent redirect. How does the redirect_uri in SolidOS' ClientID document take the user back to the malicious HTML document?

Yes, it uses the silent login (though I don't think this is based on cookies; it doesn't matter here, though). However, it never redirects to the malicious html document, this is not necessary for this attack.

There are two tabs involved:
1. The malicious html. This is opened by the victim (via phishing, etc) and performs the attack. It opens SolidOS in a new tab and stores the reference to it.
2. The SolidOS page where the user previously logged in.

The SolidOS page automatically does a silent login, involving some redirects and finally getting back to the original SolidOS url. When the login is finished, the malicious html file can use the reference to the SolidOS tab to access the SolidOS `window` (including eg `newTabReference.window.UI.authn.session.fetch`, which is the authenticated fetch from SolidOS). As the malicious html file is on the same origin as the SolidOS page, it is allowed to access its [window](https://developer.mozilla.org/en-US/docs/Web/API/Window) and use the authenticated fetch.

elf-pavlik:
@Otto-AA we are setting up a dedicated document for security practices, I have extracted scenarios you described here into an initial PR
* https://github.com/solid/security-bp/pull/2

Would you be interested in collaborating on that draft?

Issue #119, opened 2019-11-17
Handling of resources which did not go through the HTTP interface


csarven:
Should the Solid spec say anything about handling external resources mapped to pod's storage and URI space? Similarly, resources may come or cease to exist in a storage without going through the HTTP interface. Examples: `index.html` is created manually, `bar/` is mapped (or symlinked) under `/foo/`.

To what extent is this in scope and should have defined behaviours? If at all:

* Should containers list resources sharing the same path segment?
* Do these resources adhere to the same lifecycle as resources which went through the HTTP interface?
* What are the interaction models on these resources? Is there any? A default? Determine/Derived.. how?
* ..

kjetilk:
I don't see how we can possibly enforce anything that doesn't go through the interface we are specifying.

However, we also note that container membership is just a temporarily varying membership function, based on the resources that exist at any point in time. So, if some resource emerges through other means than the interface we specify, that "other means" must update the membership function, or the server has to protect itself from emerging resources. 

Since the membership function is determined at request time, it shouldn't be a big problem for e.g. file systems, since the membership function is determined by looking at the file system right then. So, if something writes to the file system by other means, it should Just Work[tm].

kjetilk:
Thinking about this, I don't think this is something we can tackle in near future. @timbl has mentioned that there should be a spec for a file-based backend of Solid. That would end up specifying many of the things we consider implementation details in the current work. 

Backend access has the potential of messing up things where we rely on side-effects, like that an ACL is deleted as a side-effect of deleting its resource. It is hard to spec this in the general case, as it has to concern itself with a lot of the things that we now consider implementation details. 

That said, it becomes important that we spec emphasizing what exists at some point, and not make assumptions about materializing certain artifacts, that will make us more resilient against such problems.

csarven:
Agreed.

A server's observable universe of resources is limited by the interactions as prescribed by the spec.

TallTed:
This whole "rely on side-effects" thing is troublesome, especially if there are ways to interact with server-managed content without going through the server (e.g., filesystem actions on sever-managed directories/resources).

There seems to be a need for a way (a script of some sort? either scheduled a la `cron`, or somehow triggered by filesystem changes) to cause the side-effects to happen when (or at least, after) actions are taken outside the view of the Solid server (e.g., filesystem additions and/or deletions)...

[...]

PR #608, opened 2023-12-21
Clarify requests with N3 document in server-representation-turtle-jsonld

csarven:
This PR is a https://www.w3.org/2023/Process-20230612/#class-3 change in the specification updating the requirement #server-representation-turtle-jsonld .

The current requirement ( #server-representation-turtle-jsonld ) only mentions the case where an RDF document is used to create a resource and then having representations of the resource available in at least JSON-LD and Turtle.

The Solid Protocol currently only specifies updating resources using HTTP PATCH requests with N3 document using the N3 Patch format ( #n3-patch ). The specification does not require the use of N3 document and with #n3-patch to create resources.

There are some implementations that already create resources with PATCH using N3 document and #n3-patch. However, there are no known (Solid) implementations creating resources with PATCH in either RDF document or N3 document with a particular format (besides #n3-patch). This is based on anecdotal evidence, some citations / test results supporting, conflicting, or clarifying this further would be great.

What's currently implied in #server-representation-turtle-jsonld in the context of PATCH and creating resources is using an RDF document-based patch format, e.g., LDP Patch. In other words, PATCH requests with N3 or SPARQL Update are not even covered. Currently implementations based on N3 or SPARQL Update to create resources with PATCH are not conflicting with the current specification, but there is no interoperability based off such conformance on paper.

The proposed change in this PR clarifies ambiguity in that when N3 documents are used to create resources, the server should respond in the same way as it does for RDF documents.

Existing implementations that accept PATCH with N3 document using N3 Patch are still conforming. This PR acknowledges their behaviour and specifies it. Implementations that do not accept N3 documents to create resources are not required to do anything and they do not become non-conforming. The behaviour is quite literally *when* a server accepts N3 document to create a resource, they must do as follows.

This change intentionally leaves the specific format of the N3 document unspecified at the level of creating a resource (whether with POST, PUT or PATCH):
* So that "N3 document" is specified on the same level as "RDF document";
* It allows servers and the "market" to experiment, adopt alternative approaches, and have reasonable variations if they choose to use different N3 formats.
* If N3 Patch changes or gets removed from the Solid Protocol in favour of another N3-based format, the general conformance to creating resources with PATCH need not be updated. Aside: if SPARQL Update makes a come back, it would need to be added to this requirement ( #server-representation-turtle-jsonld ).

---

[Preview](https://htmlpreview.github.io/?https://github.com/solid/specification/blob/5baf19af307c5f5c0f6d991484fc9a6eb192cd3b/ED/protocol.html) | [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2Fa5cdbabfab6df5b93ea8631c9bc2a3a751388750%2FED%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F5baf19af307c5f5c0f6d991484fc9a6eb192cd3b%2FED%2Fprotocol.html)


gibsonf1:
Right now TrinPod supports SPARQL Update, but we have not looked at the newer N3 spec yet.

jeff-zucker (approved):

damooo:
> [🔗](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2Fa5cdbabfab6df5b93ea8631c9bc2a3a751388750%2FED%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F5baf19af307c5f5c0f6d991484fc9a6eb192cd3b%2FED%2Fprotocol.html#server-representation-turtle-jsonld)
When a server creates a resource on HTTP PUT , POST or PATCH requests such that the request’s representation data encodes an RDF document [[ RDF11-CONCEPTS ](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2Fa5cdbabfab6df5b93ea8631c9bc2a3a751388750%2FED%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F5baf19af307c5f5c0f6d991484fc9a6eb192cd3b%2FED%2Fprotocol.html#bib-rdf11-concepts)] or an[ N3 ](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2Fa5cdbabfab6df5b93ea8631c9bc2a3a751388750%2FED%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F5baf19af307c5f5c0f6d991484fc9a6eb192cd3b%2FED%2Fprotocol.html#bib-notation3)document [ N3 ] (as determined by the Content-Type header), the server MUST accept GET requests on this resource when the value of the Accept header requests a representation in text/turtle or application/ld+json [[ Turtle ](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2Fa5cdbabfab6df5b93ea8631c9bc2a3a751388750%2FED%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F5baf19af307c5f5c0f6d991484fc9a6eb192cd3b%2FED%2Fprotocol.html#bib-turtle)] [[ JSON-LD11 ](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2Fa5cdbabfab6df5b93ea8631c9bc2a3a751388750%2FED%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F5baf19af307c5f5c0f6d991484fc9a6eb192cd3b%2FED%2Fprotocol.html#bib-json-ld11)]. [[ Source ](https://github.com/solid/specification/issues/45)][ Source ](https://github.com/solid/specification/issues/69)] [[ Source ](https://github.com/solid/specification/issues/109)] [[ Source ](https://github.com/solid/specification/issues/195)]

This may be is problematic. The clause "request's representation data", is not correct. http requests contain some "[content](https://www.rfc-editor.org/rfc/rfc9110.html#name-content)", that is not required to be that of target resource's representation. PATCH is an example case. Here the request's content is not *representation data*, but of a document encoding "patch directives".

It was so happened that protocol defined a specific patch named "solid:InsertDeletePatch", in n3's logic language to encode patch instructions over rdf resources.

But that doesn't prevent us from defining another patch type that operates over **non-rdf resources (LDP-NR)**, also encoded as n3 document. Say a "solid:JsonPatch", which contain instructions on how to patch/create a json document. Or some patch to do image crop/rotate/resize/effect operations expressed in n3 directives.

The patch document only carries some instructions to process target resource. The instruction-carrier content type has no correspondence with target resource's content-type. A json can be used to describe patch on LDP-RS, and an n3-doc can be used to describe patch on image/* resources, etc.

jeff-zucker:
> > [🔗](https://services.w3.org/htmldiff?
 
> The patch document only carries some instructions to process target resource. The instruction-carrier content type has no correspondence with target resource's content-type.

I agree in the case of existing resources - an N3-Patch on a Turtle document should result in a Turtle document.  But what about the case, which I think this PR is meant to address  where we create a new document with n3-patch.  Is that resulting document text/n3 ? I'm not sure what else it could be.  So we need to ensure that even though it has a content-type of text/n3 it should still be available as text/turtle and json-ld.

damooo:
> I agree in the case of existing resources - an N3-Patch on a Turtle document should result in a Turtle document. But what about the case, which I think this PR is meant to address where we create a new document with n3-patch.

Here, there is a conflation with terms `n3-doc`, `n3-patch-doc`. The term n3-patch is collaquial, and we are using it to refer a special patch of class [`solid:InsetDeletePatch`](https://solidproject.org/ED/protocol#server-patch-n3-simple-type)` which happened to be encoded in n3, and operates over rdf resources, or create an empty rdf resource. But that doesn't prevent us from defining new classes of patches (like ex:JsonPatch for eg.), also encoded in n3,  but operates on non-rdf resources. Thus the the patch-carrier document's content-type has no essential meaning vis-a-vis the target resource's representation content type. Only patch directives can conclusively defines.

elf-pavlik:
The spec should define it in the way that already covers the SPARQL update which uses `application/sparql-update`

michielbdejong:
I see what you're trying to do here but I think the mention of N3 documents in general can add confusion.o

michielbdejong:
This PR is on the [agenda for the 17 January CG Call](https://hackmd.io/1YFPpDmnR-GwpjTGugVXcw)

michielbdejong:
Or instead of
>  the request’s representation data encodes an RDF document

say:
> the request’s representation data encodes or updates an RDF document

ED/protocol.html
```diff
@@ -1047,7 +1047,7 @@ <h3 property="schema:name">Deleting Resources</h3>
...
...UT</code>, <code>POST</code> or <code>PATCH</code> requests such that the request’s representation data encodes an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>] or an <em>N3 document</em> [<cite><a class="bibref" href="#bib-notation3">N3</a></cite>] (as determined by the <code>Content-Type</code> header), the server <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> accept <code>GET</code> requests on this resource when the value of the <code>Accept</code> header requests a representation in <code>text/turtle</code> or <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-turtle">Turtle</a></cite>] [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].</span></span> [<a href="https://github.com/solid/specification/issues/45" rel="cito:citesAsSourceDocument">Source</a>] <a href="https://github.com/solid/specification/issues/69" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/109" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/195" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
elf-pavlik, line 1050:
```suggestion
                  <p><span about="" id="server-representation-turtle-jsonld" rel="spec:requirement" resource="#server-representation-turtle-jsonld"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="#Server">server</span> creates an RDF resource on HTTP <code>PUT</code>, <code>POST</code> or <code>PATCH</code> requests, the server <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> accept <code>GET</code> requests on this resource when the value of the <code>Accept</code> header requests a representation in <code>text/turtle</code> or <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-turtle">Turtle</a></cite>] [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].</span></span> [<a href="https://github.com/solid/specification/issues/45" rel="cito:citesAsSourceDocument">Source</a>] <a href="https://github.com/solid/specification/issues/69" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/109" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/195" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

elf-pavlik:
Wouldn't it be better to have separate requirements for

1. Creating RDF resources
2. Serving RDF resources

?

I'm thinking of a case where data gets imported to storage. In that case, the RDF resource wasn't created by HTTP
request but still needs to be served correctly.

ED/protocol.html
```diff
@@ -1465,6 +1465,11 @@ <h2 property="schema:name">Changelog</h2>
                     <td><a href="#relationship-to-ldp" rel="spec:changeSubject">#relationship-to-ldp</a></td>
                     <td property="spec:statement">Add section describing the <a href="#relationship-to-ldp">the relationship between Solid Protocol and LDP</a>.</td>
                   </tr>
+                  <tr about="#b0af4fa6-5d35-47b5-8fe3-ad597dc8bdb5" typeof="spec:Change">
+                    <td><a href="https://www.w3.org/Consortium/Process/#class-3" rel="spec:changeClass" resource="spec:ChangeClass3">3</a></td>
+                    <td><a href="#server-representation-turtle-jsonld" rel="spec:changeSubject">#server-representation-turtle-jsonld</a></td>
+                    <td property="spec:statement">Amend <a href="#server-representation-turtle-jsonld">server-representation-turtle-jsonld</a> requirement to take an HTTP request including an N3 document into account.</td>
```
csarven, line 1471:
```suggestion
                    <td property="spec:statement">Amend <a href="#server-representation-turtle-jsonld">server-representation-turtle-jsonld</a> requirement with emphasis on the created RDF source.</td>
```

csarven:
@elf-pavlik , Ack'ing your suggestions in https://github.com/solid/specification/pull/608#discussion_r1455842337 and https://github.com/solid/specification/pull/608#issuecomment-1896056023 . I updated the suggestion to use and refer to [RDF source](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-source).

I'm support merging this PR at this point. I am also okay with the consideration split the requirement but would like to doublecheck priors discussion on this before doing so, e.g., https://github.com/solid/specification/issues/119 and there are a number of other older PRs touching on this. There were considerations around `Accept` / content-negotiation that we touched on in the past. Just let me verify...

ED/protocol.html
```diff
@@ -1047,7 +1047,7 @@ <h3 property="schema:name">Deleting Resources</h3>
...
...="spec:requirement" resource="#server-representation-turtle-jsonld"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="#Server">server</span> creates an <em>RDF source</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>] on HTTP <code>PUT</code>, <code>POST</code> or <code>PATCH</code> requests, the server <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> accept <code>GET</code> requests on this resource when the value of the <code>Accept</code> header requests a representation in <code>text/turtle</code> or <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-turtle">Turtle</a></cite>] [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].</span></span> [<a href="https://github.com/solid/specification/issues/45" rel="cito:citesAsSourceDocument">Source</a>] <a href="https://github.com/solid/specification/issues/69" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/109" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/195" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
TallTed, line 1050: I *think* this means to say that GET requests for Turtle or JSON-LD must be _satisfied_ (i.e., the preferred media type must be delivered), not just that the GET requests must be "_accepted_" (whatever that would mean in HTTP). (Also, I added a missing comma.) 

(GitHub is having all sorts of issues with highlighting requested changes; I hope you have a handy DIFF tool.)
```suggestion
                  <p><span about="" id="server-representation-turtle-jsonld" rel="spec:requirement" resource="#server-representation-turtle-jsonld"><span property="spec:statement">When a <span rel="spec:requirementSubject" resource="#Server">server</span> creates an <em>RDF source</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>] on HTTP <code>PUT</code>, <code>POST</code>, or <code>PATCH</code> requests, the server <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> satisfy <code>GET</code> requests on this resource when the value of the <code>Accept</code> header requests a representation in <code>text/turtle</code> or <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-turtle">Turtle</a></cite>] [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].</span></span> [<a href="https://github.com/solid/specification/issues/45" rel="cito:citesAsSourceDocument">Source</a>] <a href="https://github.com/solid/specification/issues/69" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/109" rel="cito:citesAsSourceDocument">Source</a>] [<a href="https://github.com/solid/specification/issues/195" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

csarven: "Accept" as in 2xx-4xx with the exception of 406 "Not Acceptable" ( https://www.rfc-editor.org/rfc/rfc9110#status.406 ).

"Satisfy" is not jumping out to me as much as "accept" (which I find to be concrete). I'd like to look into splitting this requirement up any way. Can we live with "accept" for now or want to press on "satisfy"? :)

TallTed: "Satisfy" as in, the server must provide a representation of the RDF source serialized as the media type in the `Accept:` header, i.e., `text/turtle` or `application/ld+json` — regardless of whether that RDF source is stored by the Solid server as RDF in a triple/quad store, or as *any* RDF serialization.

Your 2xx-4xx "accept" is redundant to the HTTP spec, which mandates such codes be returned for GET on *any* requested resource, regardless of `Accept:` header, *including* non-existent resources (404 is neither your excluded 405 nor 406).

It seems like something's missing from this picture.

csarven: I mentioned 405 and 406 in my comment because they seemed most relevant. We know the resource is created, and so unless there are arbitrary exceptions or changes to the resource state, it is not going to be a 404. What's perhaps most significant is 406 because the requirement aimed to make sure that when a server creates an RDF source, it should make it available in Turtle and JSON-LD. It doesn't take much to come up with other arguments - not particularly worth the time to discuss (IMO) - such as "but Sarvie, if I only have append access on the container and I POST to the container, then a GET on the resource would be 403..." As true as that may be, it is irrelevant to the point in the requirement.

I'm okay to adopt your suggestion with "satisfy" but not because I find it particularly better than "accept". It is stepping into language-bikeshedding. "fulfill", "accommodate", "handle", "process", and other terms... could work too (possibly with some adjustments).

ED/protocol.html
```diff
@@ -1465,6 +1465,11 @@ <h2 property="schema:name">Changelog</h2>
                     <td><a href="#relationship-to-ldp" rel="spec:changeSubject">#relationship-to-ldp</a></td>
                     <td property="spec:statement">Add section describing the <a href="#relationship-to-ldp">the relationship between Solid Protocol and LDP</a>.</td>
                   </tr>
+                  <tr about="#b0af4fa6-5d35-47b5-8fe3-ad597dc8bdb5" typeof="spec:Change">
+                    <td><a href="https://www.w3.org/Consortium/Process/#class-3" rel="spec:changeClass" resource="spec:ChangeClass3">3</a></td>
```
csarven, line 1469: I believe the current state of the PR falls under correction class 2 in that the change does not functionally affect interpretation.

```suggestion
                    <td><a href="https://www.w3.org/Consortium/Process/#class-2" rel="spec:changeClass" resource="spec:ChangeClass2">2</a></td>
```

But if there is doubt, we can change it back to class 3.

ED/protocol.html
TallTed: Some language tweaks are definitely bike shedding. I don't think this one is.

To hopefully be clear(er), I would also be OK with "fulfill", "accommodate", "handle", "process", and (probably) other terms, instead of `satisfy`. "Accept" just doesn't convey the same intent as the rest of these, imho.

csarven: I simply borrowed from 406 Not **Accept**able because the point of the requirement was to have the server "accept" GET w/ particular media types (in Accept header) for what's originally acknowledged as an RDF source. If that doesn't convey what's intended to you (or others), that's fine, I will accept the suggested alternative.

elf-pavlik (approved):
LGTM

michielbdejong (approved):

PR #616, opened 2024-01-18
Add 2024-01-19 agenda and minutes

michielbdejong:


elf-pavlik (approved):

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Open spec PRs
+
+#### Add security consideration for serving user-created files
+URL: https://github.com/solid/specification/pull/598
+
+* VB: Please review.
+* SC: Tied to an issue, some discussion in the PR.Wrote a comment earlier. Thought this through for a while. I think the consideration is good and we should have more stuff along those lines. Depending on whether the protocol expresses a need like that says something else about the architecture. There's some assumptions baked into this. It has some tradeoffs/conflicts with other things. We should weigh a bit more and discuss further before assuming it is as intended. Does this belong in that general pattern? Is it consistent with the rest of the design? Just my considerations.
+* MdJ: Makes sense. 
+* eP: We should add it as soon as possible in whatever form we reach rough consensus. Still at least list the possible issue. We can't expect everyone to be aware of the issues. It could be an inline issue to make people aware possible exists. And perhaps people give valuable feedback.
```
TallTed, line 57:
```suggestion
* eP: We should add it as soon as possible in whatever form we reach rough consensus. Still at least list the possible issues. We can't expect everyone to be aware of all the issues. It could be an inline issue to help make people aware it exists. And perhaps people give valuable feedback.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Access to Chairs on the Repo
+* eP: Do all chairs have access to solid/specification?
+* VB: I can take care of it.
+
+ACTION: VB to give access to co-chairs to solid/specification
+
+### Project board
+* RG: Should the project board be linked from ???
+* SC: Stuff that has been mostly active until last release been updating in some ways more towards milestoning for next release. Doesn't have categorization of more recent issues. That can be improved. Process those last issues into this board. They're not there.
```
TallTed, line 72:
```suggestion
* SC: Stuff that has been mostly active until last release; been updating in some ways more towards milestoning for next release. Doesn't have categorization of more recent issues. That can be improved. Process those last issues into this board. They're not there.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* VB: I can take care of it.
+
+ACTION: VB to give access to co-chairs to solid/specification
+
+### Project board
+* RG: Should the project board be linked from ???
+* SC: Stuff that has been mostly active until last release been updating in some ways more towards milestoning for next release. Doesn't have categorization of more recent issues. That can be improved. Process those last issues into this board. They're not there. 
+* MdJ: If Sarven is not regularly grooming, then it's not good to link. Unless RG you want to pick it up yourself.
+* SC: Virginia will take care access to chairs and can say more on project board e.g. STM. 
+* VB: How are they related? Right for STM, not for issues. If that can be useful we can add more recent issues or talk about in these meetings.
```
TallTed, line 75:
```suggestion
* MdJ: If Sarven is not regularly grooming, then it's not good to link. Unless, RG, you want to pick it up yourself?
* SC: Virginia will take care of access to chairs and can say more on project board, e.g., STM. 
* VB: How are they related? Right for STM, not for issues. If that can be useful, we can add more recent issues or talk about them in these meetings.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* MdJ: We shouldn't link to it until someone works on it.
+* HZ: I can help if we agree there is value.
+* VB: I can look into using the board as mentioned.
+* HZ: If we agree it has value I can groom it.
```
TallTed, line 79:
```suggestion
* HZ: If we agree it has value, I can groom it.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* HZ: Sounds good.
+* RG: This is so that anybody who comes to this can have a centralised view which might make it easier.
+* MdJ: I'd be careful to not pick up work that we might not need. I'd say we try to work on the PRs and feel that not enough clarity and other issues to work, we can dig them up.
+
+
+#### Clarify requests with N3 document in server-representation-turtle-jsonld
+URL: https://github.com/solid/specification/pull/608
+
+* MdJ: This clarifies that when there is an N3 document, it should the same as RDF document. I suggested for the purposes of N3 patch
+* eP: I want to clarify the intent of this. In LDP we talk about (Non)RDF-Sources. Here we try to avoid that distinction but here we do. With N3 it happens to use an N3 document. The way to formulate may be to consider whether SPARQL Update may be used in the future. Take some time and see if we can formulate it in a way that's special for N3 but can also work with SPARQL. Also consideration on what's mature that can go to spec.
```
TallTed, line 94:
```suggestion
* MdJ: This clarifies that when there is an N3 document, it should be the same as RDF document. I suggested for the purposes of N3 patch
* eP: I want to clarify the intent of this. In LDP, we talk about (Non)RDF-Sources. Here we try to avoid that distinction, but here we do ??? . With N3, it happens to use an N3 document. The way to formulate may be to consider whether SPARQL Update may be used in the future. Take some time and see if we can formulate it in a way that's special for N3 but can also work with SPARQL. Also consideration on what's mature that can go to spec.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* MdJ: Want to make a suggestion into the spec of that wording?
+* eP: What do we think about instead of going through content types, like RDF and non-RDF sources. Then we don't have to list all patches.
+* SC: I asked for clarification in the chat. Some of that still applies. The requirement is more of a category of formats. The way it's written is not tied to a particular patch format. The only thing the spec says is patch is done this way, completely orthogonal to the formats that a server may be accepting, and if it does it needs to make available in ??? Right now the spec implies a server should be able to take an N3 document when you make a patch. Right now we're only saying RDF document but that is not entirely true. IF we want to say SPARQL is needed, that can be captured along those lines. There is nothing more generic to capture... right now we capture those two.
```
TallTed, line 97:
```suggestion
* eP: What do we think about ??? instead of going through content types, like RDF and non-RDF sources. Then we don't have to list all patches.
* SC: I asked for clarification in the chat. Some of that still applies. The requirement is more of a category of formats. The way it's written is not tied to a particular patch format. The only thing the spec says is _patch is done this way_, completely orthogonal to the formats that a server may be accepting, and if it does it needs to make available in ??? Right now the spec implies a server should be able to take an N3 document when you make a patch. Right now we're only saying RDF document, but that is not entirely true. IF we want to say SPARQL is needed, that can be captured along those lines. There is nothing more generic to capture... right now we capture those two. 
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* MdJ: Want to make a suggestion into the spec of that wording?
+* eP: What do we think about instead of going through content types, like RDF and non-RDF sources. Then we don't have to list all patches.
+* SC: I asked for clarification in the chat. Some of that still applies. The requirement is more of a category of formats. The way it's written is not tied to a particular patch format. The only thing the spec says is patch is done this way, completely orthogonal to the formats that a server may be accepting, and if it does it needs to make available in ??? Right now the spec implies a server should be able to take an N3 document when you make a patch. Right now we're only saying RDF document but that is not entirely true. IF we want to say SPARQL is needed, that can be captured along those lines. There is nothing more generic to capture... right now we capture those two. 
+* MdJ: What about ??? not say anything about the body but the target. 
+* SC: Creating.
+* eP: in the context of creating. If you use SPARLQ Update patch to create an RDF resource, ???
```
TallTed, line 100:
```suggestion
* eP: in the context of creating. If you use SPARQL Update patch to create an RDF resource, ???
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* SC: You're saying if I give you an image you can give that back in turtle. 
+* MdJ: Let's continue on PR. 
+
+#### Add document status to guides, templates
+URL: https://github.com/solid/specification/pull/613
+
+Proposed by: SC
+
+* MdJ: You're marking some documents as no longer in effect..
+* SC: We're using w3c solid cg repo and that's the source of truth. Minor.
```
TallTed, line 114:
```suggestion
* MdJ: You're marking some documents as no longer in effect?
* SC: We're using W3C Solid CG repo, and that's the source of truth. Minor. 
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* SC: We're using w3c solid cg repo and that's the source of truth. Minor. 
+* VB: Approved. 
+
+#### Should Solid (storage) servers support "RDF documents" containing multiple subjects (or quads)?
+URL: https://github.com/solid/specification/issues/610
+
+Proposed by: Maxime
+
+* MdJ: This is more a question than a PR maybe?
+* ML: Some implementation servers are not allowing multiple subjects in the payload when creating or updating resources. So, can't in some apps to apply the TypeIndex or any self-describing document. Was wondering it should not be part of the spec if / when we are accepting RDF document / graph we must accept the graph contains multiple subjects.
```
TallTed, line 123:
```suggestion
* ML: Some server implementations are not allowing multiple subjects in the payload when creating or updating resources. So, in some apps, can't apply the TypeIndex or any self-describing document. Was wondering if it should not be part of the spec, if/when we are accepting RDF document/graph, we must accept a graph containing multiple subjects.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* MdJ: The spec allows the server. So can never be sure. If a server only accepts triples about the document, I'd say that server is broken.
+* eP: Besides self-describing docs, certain predicates are used in a certain direction, like JSON-LD reverse. This is common in RDF. Since it is a graph, often things can be subject or object. I think it'd be interetsing to add a test case. What level would this requirement would be.
+* MdJ: Even if you do WAC for instance, you'd have different Authorization rules. A test would fail in that case. That server wouldn't be compliant.
```
TallTed, line 126:
```suggestion
* eP: Besides self-describing docs, certain predicates are used in a certain direction, like JSON-LD reverse. This is common in RDF. Since it is a graph, often things can be subject or object. I think it'd be interesting to add a test case. What level would this requirement be?
* MdJ: Even if you do WAC for instance, you'd have different Authorization rules. A test would fail in that case. That server wouldn't be compliant.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* PAC: Not sure I agree w/ you re any server should accept any type of content. I don't hav ea strong opinion. But can understand some people might want to develop servers maybe not generalist servers but still using some of the Solid server specification. Maybe it is not a good idea to rule out those people. I can share some things from the LDP work. I realised that some terms used in the group didn't make it to the spec. Used to talk about vanilla and chocolate servers - generalist servers and passive servers vs. the ones to put arbitrary resources.
+* eP: Should get back to this topic when server validation. If server advertises a shape, and not just arbitrary. I agree not every server can but can advertise the added payload.
```
TallTed, line 128:
```suggestion
* PAC: Not sure I agree with you, re: any server should accept any type of content. I don't have a strong opinion. But I can understand that some people might want to develop servers — maybe not generalist servers — but still using some of the Solid server specification. Maybe it is not a good idea to rule out those people. I can share some things from the LDP work. I realised that some terms used in the group didn't make it to the spec. Used to talk about vanilla and chocolate servers — generalist servers and passive servers vs. the ones to put arbitrary resources.
* eP: Should get back to this topic when doing server validation. If a server advertises a shape, and not just arbitrary. I agree not every server can, but can advertise the added payload.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* MdJ: That'd be perhaps in future spec?
+* SC: Current spec says.. category of contraints, one imposed by the server, expectation that container should behave.. Anything beyond is open, spec allows the case PAC was mentioning. There's also the possibility to advertise those contraints. Existing servers, if they want to continue doing what they're doing might want to get into that habit. 
+* SC: There's a similar consideration raised in Solid Protocol, whether a read-only Solid server should be conforming. Normally we think read-write. Wehther a conforming server not implementing those methods.. Some servers might want to do the mminimal thing. That's fine.
```
TallTed, line 131:
```suggestion
* SC: Current spec says... category of constraints, one imposed by the server, expectation that container should behave... Anything beyond is open; spec allows the case PAC was mentioning. There's also the possibility to advertise those contraints. Existing servers, if they want to continue doing what they're doing, might want to get into that habit.
* SC: There's a similar consideration raised in Solid Protocol, whether a read-only Solid server should be conforming. Normally we think read-write. Whether a conforming server not implementing those methods... Some servers might want to do the minimal thing. That's fine.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* SC: Current spec says.. category of contraints, one imposed by the server, expectation that container should behave.. Anything beyond is open, spec allows the case PAC was mentioning. There's also the possibility to advertise those contraints. Existing servers, if they want to continue doing what they're doing might want to get into that habit. 
+* SC: There's a similar consideration raised in Solid Protocol, whether a read-only Solid server should be conforming. Normally we think read-write. Wehther a conforming server not implementing those methods.. Some servers might want to do the mminimal thing. That's fine. 
+* MdJ: "Advertising its chocolateness" (TM).
+* ML: So this constraint would be know at run-time and not when they're choosing their pod. So then I get a pod but not might as expected.
+* eP: Separate topic perhaps how people can choose. Who is setting the constraints. It wouldn't be arbitrary constraints. How we can educate users to picking a storage.
+* SC: Server could pass the test suite. When it's deployed it could completely reject requests including multiple subject. I don't think test suite will catch that. Comes down to whether URI owner allows that.
```
TallTed, line 135:
```suggestion
* ML: So this constraint would be known at run-time, and not when they're choosing their pod. So then I get a pod but might not be as expected.
* eP: Separate topic perhaps; how people can choose. Who is setting the constraints. It shouldn't be arbitrary constraints. How we can educate users picking a storage?
* SC: Server could pass the test suite. When it's deployed, it could completely reject requests including multiple subjects. I don't think the test suite will catch that. Comes down to whether URI owner allows that.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
... datastore. The GUI running on top of it. We need a much stricter spec for pod as a data pirvot. App A writes to Solid Pod and read by App B. For one power user, usefullness comes if it can be reused by App B. For that to open apps need to know what to expect from the pod, when it will fail, or discoverably capabilities for minimal reliable working thing. Data application rights need to be available in an understandable way. but then the choice of ontology and data shape matters. The business logic around that need sto be understood. need server behaviour to be strict - can't have chocolate servers - have data conventions, which ontologies to use - even if we care a WebID Profile it owuld have x:name but the app is using y:name. No good data conventions and also access control and they go hand in hand. The activities in Data Modules, we document and write translations between formats, reusable code in JavaScript, which handles the translation for you. This is one of the sites where we document data conventions. For example, bookmarking is in different vocabs. It'd be nice to make that compatible. Some apps use x and other y without knowing the other exists. SDM also abstracts them.
```
TallTed, line 145:
```suggestion
* MdJ: Solid Data Modules uses the platform for data interoperability. I'd call "'small' Solid = RWW". As a power user, there can be data browsers. `mashlib` is one. Also from modula, `ontolo`. Kind of like SolidOS, where any data has a particular shape, can have a view. I also see a bigger vision "'big' Solid = data pivot". It is a datastore. The GUI running on top of it. We need a much stricter spec for pod as a data pivot. App A writes to Solid Pod, read by App B. For one power user, usefulness comes if App A's data can be reused by App B. For that to be open, apps need to know what to expect from the pod, when it will fail, or have discoverable capabilities for minimal reliable working thing. Data application rights need to be available in an understandable way, but then the choice of ontology and data shape matters. The business logic around that needs to be understood. need server behaviour to be strict — can't have chocolate servers — have data conventions, which ontologies to use — even if we share a WebID Profile, it might have `x:name`, but the app is using `y:name`. No good data conventions and also access control and they go hand in hand. The activities in Data Modules, we document and write translations between formats, reusable code in JavaScript, which handles the translation for you. This is one of the sites where we document data conventions. For example, bookmarking is in different vocabs. It'd be nice to make that compatible. Some apps use `x` and other `y` without knowing the other exists. SDM also abstracts them.
```

meetings/2024-01-17.md
```diff
@@ -0,0 +1,161 @@
...
+* MdJ: There is a chat room, NLnet funding for this project.
+* eP: Looks interestsing and with demos would be good in action. For different apps or templates from servers, they don't have a strong reason to use this or some kind of a best practice. Devs go to NPM and tend to pick things wiht most downloads. Most of us take our extensions to use what they'll use. Most devs would be happy to choose the simpler.
+* eP: As soon as we have something like recommended shape. Similar like schemaorg - tightly curated schema - and there is LOV to pick from.
+* MdJ: We don't have the issue tracker to ???
+* eP: You mentioned data types for access control. For SAI we have access needs to reference a project as an IRI. 
+* MdJ: I'd want to import difference ones from my pod and use code that can read both formats.
+
```
TallTed, line 152:
```suggestion
* eP: Looks interesting, and with demos would be good in action. For different apps or templates from servers, they don't have a strong reason to use this or some kind of a best practice. Devs go to NPM and tend to pick things wiht most downloads. Most of us take our extensions to use what they'll use. Most devs would be happy to choose the simpler.
* eP: As soon as we have something like recommended shape. Similar like schema.org — tightly curated schema — and there is [LOV](https://lov.linkeddata.es/dataset/lov/) to pick from.
* MdJ: We don't have the issue tracker to ???
* eP: You mentioned data types for access control. For SAI, we have access needs to reference a project as an IRI. 
* MdJ: I'd want to import difference ones from my pod and use code that can read both formats.

```

TallTed (changes requested):

michielbdejong:
@VirginiaBalseiro can you merge this please?

PR #617, opened 2024-01-24
Add 2024-01-24 agenda

VirginiaBalseiro:


meetings/2024-01-24.md
```diff
@@ -41,16 +50,96 @@
...
+* ...: Some of that information could bubble up into the Solid protocol. Will need to consider how those tables match the existing Solid API
+* ...: The protocol uses "read/write/append" terminology. Perhaps that terminology can be improved. Can others (esp. implementers) arrive at the same conclusion
+* TbL: if this is a big job to review the spec like this, perhaps the spec should just contain a table. This is important because this affects interoperability across servers (e.g. chat with append). Append functionality is important; consistency is important. In short, the table or similar should be in the spec document
+* SC: We did consider adding the tables directly into the spec, but the spec should also be able to tell you independently
+* ...: re: Append, if there are strange or unintuitive behaviors, it could be resolved in the authorization mechanism or in the Solid protocol (e.g. if an agent has append-only, changing the description of a resource; e.g. PATCH cannot allow creating a resource; PUT requires Write, but PATCH-to-create does work for append-only operations)
+* ...: PATCH RFC says PATCH-to-create is a MAY, but the Solid protocol says PATCH-to-create is a MUST
```
TallTed, line 101:
```suggestion
* ...: re: Append, if there are strange or unintuitive behaviors, they could be resolved in the authorization mechanism or in the Solid protocol (e.g., if an agent has `append-only`, changing the description of a resource; e.g., `PATCH` cannot allow creating a resource; `PUT` requires `Write`, but `PATCH-to-create` does work for `append-only` operations)
* ...: `PATCH` RFC says `PATCH-to-create` is a `MAY`, but the Solid protocol says `PATCH-to-create` is a `MUST`
```

meetings/2024-01-24.md
```diff
@@ -41,16 +50,96 @@
...
+* RG: suggestion: it might be useful to start a primer with this table in a separate document. This references several different specifications: access control, protocol, etc. This would allow us to bring together this information in a way that is hard with several different protocols
+* eP: The question is also where do access mode definitions belong? They are currently in WAC and ACP. This is also used in SAI and Inrupt's Access Grants. Not all of those want normative dependencies on WAC. If these are more universal access modes, they could be defined independently of WAC (e.g. in the Protocol spec)
+* ...: For PATCH-append, one can add alternative statements. For example, add another triple with the same subject and predicate (and different object value). Perhaps there could be cardinality constraints on the resource
+* ...: Or each statment is its own resource
+* ...: If Append is used for creating a new resource in a container, perhaps "Create" is a better terminology. Since append currently is used for both create and update. A different term would allow us to not overload these terms.
```
TallTed, line 106:
```suggestion
* eP: Another question is, where do access mode definitions belong? They are currently in WAC and ACP. This is also used in SAI and Inrupt's Access Grants. Not all of those want normative dependencies on WAC. If these are more universal access modes, they could be defined independently of WAC (e.g., in the Protocol spec)
* ...: For `PATCH-append`, one can add alternative statements. For example, add another triple with the same subject and predicate (and different object value). Perhaps there could be cardinality constraints on the resource
* ...: Or each statment is its own resource
* ...: If `Append` is used for creating a new resource in a container, perhaps "`Create`" is a better terminology. Since `append` currently is used for both `create` and `update`. A different term would allow us to not overload these terms.
```

meetings/2024-01-24.md
```diff
@@ -41,16 +50,96 @@
...
+* TbL: I do not want to change the terms defined in WAC. They have been around for a long time. Append on container means create, append on resource means update, which is logical, especially in the context of the larger graph
+* ...: These are concepts and URIs that users and developers see and it is important for them to be consistent. Concept of append is used not just by WAC
```
TallTed, line 108:
```suggestion
* TbL: I do not want to change the terms defined in WAC. They have been around for a long time. `Append` on container means `create`, `append` on resource means `update`, which is logical, especially in the context of the larger graph
* ...: These are concepts and URIs that users and developers see and it is important for them to be consistent. Concept of `append` is used not just by WAC
```

meetings/2024-01-24.md
```diff
@@ -41,16 +50,96 @@
...
+* VB: We will likely need more time to discuss this
+* SC: if the use cases allow someone to allow smth to a container, Append is perfect for that (e.g. POST). If someone has append on a container, using PATCH will allow someone to change the description. If we were to eliminate that, we would need to say something about that in the Protocol or else in the access control system. This in itself doesn't require defining new access modes.
```
TallTed, line 110:
```suggestion
* SC: if the use cases allow someone to allow something to a container, `Append` is perfect for that (e.g., `POST`). If someone has `append` on a container, using `PATCH` will allow someone to change the description. If we were to eliminate that, we would need to say something about that in the Protocol or in the access control system. This in itself doesn't require defining new access modes.
```

meetings/2024-01-24.md
```diff
@@ -41,16 +50,96 @@
...
+* VB: can we agree to revisit next week
+* TbL: the fact that we allow people to store random triples on a container is something I've never used. Agree that there are inconsistencies there. Perhaps we should consider removing that facility. Let's continue this discussion next week
+
+ACTION: VB will add this topic for next week
 
 ### CG Call Frequency
 * MdJ: We could consider having fewer and lighter plenary CG calls, in favour of more in-depth work meetings on specific work items, and moving spec PR reviews into either async for small ones or STMs for bigger ones (then also requiring the STM to write its conclusions into the issue it was about).
+* ...: CGs are often where different implementers come together to build a technology
+* ...: the specification work is split between the CG and the (yet to be formed) WG
+* ...: it is important to make our apps interoperable, with the weekly CG meeting, there is a big focus on specs, and I would like to propose that we use this weekly call as more of a weekly plenary
```
TallTed, line 120:
```suggestion
* ...: it is important to make our apps interoperable. In the weekly CG meeting, there is a big focus on specs, and I would like to propose that we use this weekly call as more of a weekly plenary
```

meetings/2024-01-24.md
```diff
@@ -41,16 +50,96 @@
...
+* ...: more demos, more discussion, less spec PR discussion
+* ...: perhaps a rotation: one week is demos, one week is spec PRs, one week for roadmap (e.g. topic proposed by eP)
+* TbL: this sounds good, can we talk about relationships between Solid and nearby projects? (e.g. DWEB, exports: sync with external services iCloud, Google)
```
TallTed, line 123:
```suggestion
* ...: perhaps a rotation: one week is demos, one week is spec PRs, one week for roadmap (e.g., topic proposed by eP)
* TbL: this sounds good. can we talk about relationships between Solid and nearby projects? (e.g., DWEB, exports: sync with external services iCloud, Google)
```

meetings/2024-01-24.md
```diff
@@ -41,16 +50,96 @@
...
+* SC: There is a PR about workspace selection. For CGs, there is a rough understanding that they take on issues in the incubation space
+* ...: historically, we have always had a space for implementation and demos
+* ...: we should be careful not to overstep what other groups take on. If/when WG is formed, the CG shouldn't reproduce what the WG is doing. Similar for developer experience
+* ...: No strong point on meeting structure. We used to have panels, and over time we didn't have the numbers to keep those topics alive
+* ...: Would like the weekly mtg should stay, allows us to welcome people to the group and help them navigate their way into Solid
+* ...: If we can grow the other meetings (e.g. special topic meeting), that would be great, but even there the numbers are not there
```
TallTed, line 129:
```suggestion
* ...: Would like the weekly mtg to stay, allows us to welcome people to the group and help them navigate their way into Solid
* ...: If we can grow the other meetings (e.g., special topic meeting), that would be great, but even there the numbers are not there
```

meetings/2024-01-24.md
```diff
@@ -74,3 +163,12 @@ URL: https://github.com/solid/specification/issues/227
 
 #### Server Description
 URL: https://github.com/solid/specification/issues/355
+
+
+### Should Solid (storage) servers support "RDF documents" containing multiple subjects (or quads)?
+URL: https://github.com/solid/specification/issues/610
+
+* SC: Chairs, can you use "status" labels, e.g., "waiting for commenter"
+* SC: I tihnk this issue can be closed but if there is anyhting, it should be broken up into specific questions/issues to follow up. I'd like to hear from ML on what needs further clarification or request...
```
TallTed, line 172:
```suggestion
* SC: I think this issue can be closed but if there is anything, it should be broken up into specific questions/issues to follow up. I'd like to hear from ML on what needs further clarification or request...
```

TallTed (changes requested):

meetings/2024-01-24.md
```diff
@@ -0,0 +1,174 @@
...
+* ...: also the test suite and QA, for interop, test suite, it is important to leverage that in CG meetings
+* VB: +1 on QA.
+* TbL: when proposing the agenda as a PR, I find that hard to process. It would be more reasonable to do something similar with (e.g.) Google docs. Especially for reading the agenda on a mobile device
+
+### Show some use cases on the "About" page of the solidproject.org website
+URL: https://github.com/solid/solidproject.org/issues/813
+
+* Proposed by Maxime Lecoq
+* ML: We already indicated ([in this PR](https://github.com/solid/solidproject.org/pull/814)) that Solid can be used for groups (the home page was only referring to individuals before). The idea is to show that Solid can be used for organisations, governments, and for the data space ecosystem.
+* ...: For instance, at Startin'Blox we received feedback from people of the data space community (like GaiX, IDSA) who were wondering how Solid work for their use cases. I'm not an expert on Data Space but it looks to me that some of their uses cases (like contracts) can be "implemented" in Solid (using client-to-client standard).
```
TallTed, line 141: I'm not certain about the Gaia-X or IDSA changes nor websites, but they seem the most likely. It would be good if Maxime (@lecoqlibre) can confirm/correct!
```suggestion
* ...: For instance, at [Startin'Blox](https://startinblox.com/) we received feedback from people of the data space community (like [Gaia-X](https://gaia-x.eu/), [IDSA](https://www.idsalliance.org/)) who were wondering how Solid work for their use cases. I'm not an expert on Data Space but it looks to me that some of their uses cases (like contracts) can be "implemented" in Solid (using client-to-client standard).
```

TallTed (approved):
Looks good, modulo response from @lecoqlibre

meetings/2024-01-24.md
lecoqlibre: For IDSA I was referring to this one: https://internationaldataspaces.org/ @TallTed 

Startin'Blox and Gaia-X links are good.

TallTed: Updated. Thanks, @lecoqlibre!

PR #633, opened 2024-03-04
CG Meeting agenda 2024-03-06

hzbarcea:
Draft agenda for the 03/06 CG meeting


TallTed:
(Change PR title from `CG Meeging` to `CG Meeting`)

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
 
 ## Topics
 
+### Status of the WG Charter
+* HZ: PAC, any updates?
+* PAC: The W3C Council that decides on FO's has started. Not sure how long it will take. The Council re recent Process is relatively new. The Team's recommendation was to submit a new charter. Moving with the existing charer is inappapropiate and that TAG also raised some concerns. When we submit a new charter that tells us objections that can be addressed in the new charter. Some of them may be raised again. Not all objections are 'justified'. We can argue against those. If they come up again, we can go to the Council again. We know these objections would come up. Then the home that the next Council can overrule them again if they come up.
+* TBL: With one of the objectors I had back and forth. Their issues could be resolvable. The name of the group changing being larger.
```
TallTed, line 56:
```suggestion
* PAC: The W3C Council that decides on FO's has started. Not sure how long it will take. The Council Process is relatively new. The Team's recommendation was to submit a new charter. Moving with the existing charter is inappropriate, and the TAG also raised some concerns. That tells us objections that can be addressed in the new charter. Some of them may be raised again. Not all objections are 'justified'. We can argue against those. If they come up again, we can go to the Council again. We knew some objections would come up. The hope is that the next Council will overrule them again if they come up again.
* TBL: I had back and forth with one of the objectors. Their issues could be resolvable. Changing the name of the group is larger.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* PAC: I wanted to bring this forward. I had this discussion with some people. You're going for a specific solution. THe Solid specificaiotn should be one input but not the predefined goal. As TBL mentioned, there were back and forth on addressing the problem space. My personal feeling is that, a sign of good to go is to keep the Solid name out of the Group name. We're not assuming that the recommendation will look like exactly the CG specification. I guess I would like to get a strawpoll on whether this is a good or bad idea. Should we stick to a Solid WG or some other name - something more abstract or general.
+* HZ: I'm understanding different themes. From TBL, There is no need for re-write but just minor changes. TBL?
```
TallTed, line 58:
```suggestion
* PAC: I wanted to bring this forward. I had this discussion with some people. You're going for a specific solution. THe Solid specification should be one input but not the predefined goal. As TBL mentioned, there was back and forth on addressing the problem space. My personal feeling is that, a sign of good to go is to keep the Solid name out of the Group name. We're not assuming that the recommendation will look exactly like the CG specification. I guess I would like to get a straw poll on whether this is a good or bad idea. Should we stick to a Solid WG or some other name — something more abstract or general?
* HZ: I'm understanding different themes. From TBL, there is no need for re-write but just minor changes. TBL?
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* TBL: Changes are addressing some of the questions that they (AC?) had. One of them was about dependencies. Some are for the REC track and so we need to prepare to move those things. Recognise them. The charter didnt' call that out. I pushed back on one point, the CSS WG for example, where it was raised to charter a general problem. But, it is not called the "Style System Where the User Can Something Output" WG. They didn't want other systems to be in there. So, like with VC. It is a spec. You could say that ensuring the certain things by people and cryptographically, ??? to do that. One of the htings is about the Mission and I'm happy to change that. Maybe not easy to do. From "Fix the Solid Spec" to "World of Read-Write Authenticated Data" would be fine.
+* PAC: TBL and I are saying the same thing I think we change the charter and for the Council to decide any way. Please adjust and then you can go. Truly this is not TBL's recommendation but ???. I still have some changes to push to GitHub and submit to the CG. I'm currently re-writing the Mission. So, TBL, this WIP.
```
TallTed, line 60:
```suggestion
* TBL: Changes are addressing some of the questions that they (AC?) had. One of them was about dependencies. Some are for the REC track and we need to prepare to move those things. Recognise them. The charter didn't call that out. I pushed back on one point, the CSS WG for example, where it was raised to charter a general problem. But, it is not called the "Style System Where the User Can Something Output" WG. They didn't want other systems to be in there. So, like with VC. It is a spec. You could say that ensuring the certain things by people and cryptographically, ??? to do that. One of the things is about the Mission, and I'm happy to change that. Maybe not easy to do. From "Fix the Solid Spec" to "World of Read-Write Authenticated Data" would be fine.
* PAC: TBL and I are saying the same thing. I think we change the charter and let the Council decide anyway. Please adjust and then you can go. Truly this is not TBL's recommendation but ???. I still have some changes to push to GitHub and submit to the CG. I'm currently re-writing the Mission. So, TBL, this WIP.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* TBL: Ok, thanks.
+* eP: I do support some natural language on read-write protected data. Nasty contrast with open data. So here is more protected resources. I participated in the FedCM WG proposal. There is possible overlap. There is no problem with it but Solid WG was pointe out. Especially with things like Solid-OIDC. If FedCM WG goes ahead then we might have alternative to Solid-OIDC. We should at least consider listing them as a possible Group to coordinate with.
```
TallTed, line 62:
```suggestion
* eP: I do support some natural language on read-write protected data. Nasty contrast with open data. So here are more protected resources. I participated in the FedCM WG proposal. There is possible overlap. There is no problem with it but Solid WG was pointed out. Especially with things like Solid-OIDC. If FedCM WG goes ahead, we might have an alternative to Solid-OIDC. We should at least consider listing them as a possible Group to coordinate with.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* HZ: Do you still want to do the poll for changing the name?
+* SC: Thinking about the name depends on the scope/mission. What do you want to get out of this poll 
+* PAC: Whether people think it is important to stick to "Solid" or not. I agree that the name depends on the mission. At this point do we feel we should stick with Solid or change.
```
TallTed, line 65:
```suggestion
* SC: Thinking about the name depends on the scope/mission. What do you want to get out of this poll?
* PAC: Whether people think it is important to stick to "Solid" or not. I agree that the name depends on the mission. At this point, do we feel we should stick with Solid or change?
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* HZ: I'm neutral but biased towards keeping the name because people are familiar with it. I don't know if you find a name that reflects.
+* SC: I don't have a strong opinion, I don't think it matters. In SocialWG, there was ActivityPub, AS2, LDN etc. Similarly this WG could publish a set of spec. In SocialWG some specs were overlaping. I don't think that the name has branding value to the working group. At some point it will be about sucess of the spec. I'm open to the idea to pick a different name. I think about Socially aware web, or talking about fixing the Web. I'm open to changing the name. Depends on the scope/mission/deliverables.
```
TallTed, line 67:
```suggestion
* HZ: I'm neutral but biased towards keeping the name because people are familiar with it. I don't know if you can find another name that reflects the same.
* SC: I don't have a strong opinion, I don't think it matters. In SocialWG, there were ActivityPub, AS2, LDN, etc. Similarly this WG could publish a set of specs. In SocialWG, some specs were overlapping. I don't think that the name has branding value to the working group. At some point it will be about sucess of the spec. I'm open to the idea of picking a different name. I think about Socially aware web, or talking about fixing the Web. I'm open to changing the name. Depends on the scope/mission/deliverables.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* AC: I'm interested in the scope/mission. Content coming out of the WG. If it helps us move the WG along, then happy to move. Basically neutral. Leaning slightly on changing the name.
+* TBL: In principle on changing the name, as mentioned before with read-write protected/authenticated data. Even discussing in this group we have different opinions.
+* eP: Sometimes branding can create tension especially in standards. To accomplish the mission of Solid, we need something more than what this Working Group can accomplish. Solid could push a storage spec, access control spec, etc. that other things approaches can also work with.. Thre are other projects in the same perimiter and don't worry about naming them because the Solid will build on them. There is more to Solid.
```
TallTed, line 70:
```suggestion
* TBL: In principle on changing the name, as mentioned before with read-write protected/authenticated data. Even discussing in this group, we have different opinions.
* eP: Sometimes branding can create tension, especially in standards. To accomplish the mission of Solid, we need something more than what this Working Group can accomplish. Solid could push a storage spec, access control spec, etc., that other things can also work with. There are other projects in the same perimeter; don't worry about naming them because Solid will build on them. There is more to Solid.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* AB: Neutral.
+* RG: The branding can be an asset or liability. Everyone knows there is something going on with Solid. Losing that could be a problem but everyone has a different idea on Solid. Compromise idea for the name would be using Solid as a different expansion. Something like "SOcial, LInked, Decentralization" or another play on Solid so keeping the branding but meaning is something more.
+* TBL: Maybe this Group can standardise different htings. Solid is an internet protocol. What bits go across the internet. Like CalDAV, IMAP - if I write a combination of different semantics. Philosophically very important to the WG because they'l be conscience of that. SC  understood ??? what ???. The Solid Protocol deliverable has a lot of value and use by different apps. The Mission defines the internet.
```
TallTed, line 73:
```suggestion
* RG: Branding can be an asset or liability. Everyone knows there is something going on with Solid. Losing that could be a problem, but everyone has a different idea on Solid. Compromise idea for the name would be using Solid as a different expansion. Something like "SOcial, LInked, Decentralization" or another play on Solid so keeping the branding, but meaning something more.
* TBL: Maybe this Group can standardise different things. Solid is an internet protocol. What bits go across the internet. Like CalDAV, IMAP — if I write a combination of different semantics. Philosophically very important to the WG because they'll be conscious of that. SC understood ??? what ???. The Solid Protocol deliverable has a lot of value and use by different apps. The Mission defines the internet.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* HZ: Was at Data Transfer Initiative. Talking about portability. Aware of Solid but wondering why work out there at W3C. IMO
+* PAC: Thanks all for sharing thoughts. Others have said before re branding, and liability. I like what eP said what the Group will do. Perhaps that's also how we present. To say that initiatives such as Solid. Which is differen tthan we are goin gto make Solid a rec. I expect than for Solid to be a thing. Then telling W3C Members we are not 
+* SC: Social Web had understanding about what Social meant. Controlling own data / Personal / Protected also brings some meaning, several initiatives touch area of people controling their own data. This is not a recent topic. If the group is focusing on personal compared to social, ..., the specs we come up with here are just scratching the surface when it comes to fixing the web. We need other gruops to also bring their expetise.
```
TallTed, line 76:
```suggestion
* PAC: Thanks all for sharing thoughts. Others have said before re branding, and liability. I like what eP said what the Group will do. Perhaps that's also how we present. To say that initiatives such as Solid, which is different than we are going to make as a Solid rec. I expect Solid to be a thing. Then telling W3C Members we are not ???
* SC: Social Web had understanding about what Social meant. Controlling own data / Personal / Protected also brings some meaning, several initiatives touch area of people controlling their own data. This is not a recent topic. If the group is focusing on personal compared to social, ..., the specs we come up with here are just scratching the surface when it comes to fixing the web. We need other groups to also bring their expertise.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* eP: Good to stay close to Linked Data Platform. Not sure if we should make it more prominent. At least point it to the paragraph in the Solid Protocol spec. What's missing or helpful towards LDP.
+* RG: Minor nit: Difference on protocols. Solid is working with any data not just personal. I think that needs to be emphasised as well. Not just contacts etc.
+* TBL: I think it is important that it is not just private data. A lot of it public, government data, maps. We need agreement in this group. These are not global. Some maybe restricted to the group. Most of the values on the Web is not just personal or completely private or public but in between. From the point of data spectrum, that's something Solid needs to address.
```
TallTed, line 79:
```suggestion
* RG: Minor nit: Difference on protocols. Solid is working with any data not just personal. I think that needs to be emphasised as well. Not just contacts, etc.
* TBL: I think it is important that it is not just private data. A lot of it is public, government data, maps. We need agreement in this group. These are not global. Some may be restricted to the group. Most of the data on the Web is not just personal or completely private or public but in between. From the point of data spectrum, that's something Solid needs to address.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+
 ### Demo
 
 ### Special Topic Meetings
 URL: https://github.com/orgs/solid/projects/16/views/1
-* Solid Notifications discussed on 03/05 
+* HZ: Solid Notifications discussed on 2024-03-05. If you have proposals for upcoming STMs we can add to schedule.
+* RG: Can we have a slot for Solid Notifications?
+* HZ: Yes.
+* TBL: It has been awhile since we had an STM.
```
TallTed, line 89:
```suggestion
* TBL: It has been a while since we had an STM.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* HZ: We had them in last weeks
+* SC: Was Solid Notifications discussed yesterday?
+* RG: Only audio.
+* eP: I guess we're not updating the CG calendar? Perhaps synchronise the board.
+* HZ: The meetings are in the Solid Calendar
+* TBL: There is a Discussion 555 which is where last discussions were mentioned. https://github.com/solid/specification/discussions/555
+* SC: We moved it to a project board, the discussion is now closed and chairs are maintaining the board. As far as the calendar, CG and STM have fixed slots. I used to update specific meetings to point to hackmd and the agenda. Right now the agenda is only mentioned in matrix chat, we could improve how we get the details out. https://github.com/orgs/solid/projects/16
```
TallTed, line 96: Note that #555 still *looks* alive. Someone coming back to it, to see the latest addition to the bottom of the table, won't necessarily see the top-posted notice that it's been closed with new info being posted at . I suggest adding `<strike>` or similar to make clear what is no longer valid. https://github.com/orgs/solid/projects/16 requires more thinking, parsing the date column which is no longer in ISO-numerics, figuring out that "status — done" means the meeting has been held... and now, it's even more challenging, because the last dated entry is March 5 (which is now in the past) and shows status "discussed" instead of "done" or at least "scheduled" which *seems* to make more sense for a date in the past....
```suggestion
* TBL: There is a [Discussion 555](https://github.com/solid/specification/discussions/555) which is where last discussions were mentioned. 
* SC: We moved it to [a project board](https://github.com/orgs/solid/projects/16). The discussion is now closed and chairs are maintaining the board. As far as the calendar, CG and STM have fixed slots. I used to update specific meetings to point to hackmd and the agenda. Right now the agenda is only mentioned in matrix chat; we could improve how we get the details out. 
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
 
 ### Release 0.11.0 Milestone issues
 URL: https://github.com/solid/specification/milestone/7
 
+* HZ: We have only 6 issues open. A lot are quite old. I'm curious if we can close or remove them from the release. IRI could be removed from scope. I'd like to close this.
+* TBL: Fake 404 instead of a 403. It is a bit philosophical. I prefer not to lie 404 but to give 403.
+* PAC: I want to make more general remark. The solid Spec seems to override HTTP but I don't htink that's the intend. Restating HTTP might be bad practice. There is some recommendation in the HTTP 7231.. do we have to respecify or override.
+* SC: There is a lot here. There are some requirements that could be inherited from HTTP, but nothing should spec thing as new requiremtn. We may only increase SHOULD to MUST. Ruben raised a question and I was trying to come up with argument where 404 is useful, in many places Solid uses 403. Currently we are consistent with 403 and sometimes 404 is allowed. There is a degree of freedom for the server to return apropriate status codes.
```
TallTed, line 104:
```suggestion
* PAC: I want to make a more general remark. The solid Spec seems to override HTTP but I don't think that's the intent. Restating HTTP might be bad practice. There is some recommendation in the HTTP 7231.. do we have to respecify or override?
* SC: There is a lot here. There are some requirements that could be inherited from HTTP, but nothing should spec thing as new requirement. We may only increase SHOULD to MUST. Ruben raised a question, and I was trying to come up with an argument where `404` is useful, in many places Solid uses `403`. Currently we are consistent with `403` and sometimes `404` is allowed. There is a degree of freedom for the server to return appropriate status codes.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* SC: I do not support idea to remove anything from the milestone. The milestone should include whatever is meaningful to release the new version. We could remove the milestone and just snapshot what we currently have. Implementation experience and other factors are taken into account for releasing new versions. This group should focus on the incubation aspect to give input / signal to the specification. It is a bit strange that the discusion should stop now.
```
TallTed, line 105:
```suggestion
* SC: I do not support the idea of removing anything from the milestone. The milestone should include whatever is meaningful to release the new version. We could remove the milestone and just snapshot what we currently have. Implementation experience and other factors are taken into account for releasing new versions. This group should focus on the incubation aspect to give input/signal to the specification. It is a bit strange that the discussion should stop now.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* HZ: If that's the case, that's fine, but then we should focus on making progress on those if they are important for the next release. Thanks!
+* eP: if we have an issue on milesone, someone has to sat aside to move forward. There are two issues assigned to SC, and one to JB. And we should make some criteria on moving them. Hard to handle without it.
+* SC: Sometimes arguments are balanced enough that the design could go in either direction. If people are implementing one approach it is a good signal for the spec. If there is not enough implementation experience, that's not adequate/sufficient input. Anyone can make a random call and the spec would be still consistent, we need to look at commitments and impelemtntations to be more grounded. Some of that is expressedin "Success/Exit"-like Criteria.
```
TallTed, line 108:
```suggestion
* eP: if we have an issue on a milesone, someone has to set it aside to move forward. There are two issues assigned to SC, and one to JB. We should make some criteria on moving them. Hard to handle without it.
* SC: Sometimes arguments are balanced enough that the design could go in either direction. If people are implementing one approach, it is a good signal for the spec. If there is not enough implementation experience, that's not adequate/sufficient input. Anyone can make a random call and the spec would be still consistent. We need to look at commitments and implementations to be more grounded. Some of that is expressed in "Success/Exit"-like Criteria.
```

meetings/2024-03-06.md
```diff
@@ -45,12 +50,93 @@ URL: https://github.com/solid/specification/discussions/618
...
+* RG: On the issue of differences in HTTP and Solid spec. That's what I gathered from PAC and people at IETF. Theren't many but still need to address. Have STM to get over the milestone hump.
+* TBL: We don't have a WG. We have to make sure to make progress on the spec in good order. As a general point when it comes to HTTP, it was initially a Read ??? protocol, and now we are doing write ???. Even if you get different formats there are consistency requirements. Basically moreso than HTTP. Where HTTP specs didn't.
```
TallTed, line 110:
```suggestion
* RG: On the issue of differences in HTTP and Solid spec — that's what I gathered from PAC and people at IETF. There aren't many but still need to address them. Have STM to get over the milestone hump.
* TBL: We don't have a WG. We have to make sure to make progress on the spec in good order. As a general point when it comes to HTTP, it was initially a Read-only protocol, and now we are doing read+write. Even if you get different formats, there are consistency requirements. Basically moreso than HTTP. Where HTTP specs didn't.
```

TallTed (changes requested):
Repeating inline comment, so it's more persistent/visible after change suggestions are applied.
Note that #555 still *looks* alive. Someone coming back to it, to see the latest addition to the bottom of the table, won't necessarily see the top-posted notice that it's been closed with new info being posted at . I suggest adding `<strike>` or similar to make clear what is no longer valid. https://github.com/orgs/solid/projects/16 requires more thinking, parsing the date column which is no longer in ISO-numerics, figuring out that "status — done" means the meeting has been held... and now, it's even more challenging, because the last dated entry is March 5 (which is now in the past) and shows status "discussed" instead of "done" or at least "scheduled" which *seems* to make more sense for a date in the past....

TallTed (changes requested):
_(Repeating inline comment here, where it will be more visible and persist after change suggestions are acted on...)_

Note that #555 still *looks* alive. Someone coming back to it, to see the latest addition to the bottom of the table, won't necessarily see the top-posted notice that it's been closed with new info being posted at https://github.com/orgs/solid/projects/16. I suggest adding `<strike>` or similar to make clear what is no longer valid. 

Also note that https://github.com/orgs/solid/projects/16 requires more thinking, parsing the date column which is no longer in ISO-numerics, figuring out that "status — done" means the meeting has been held... and now, it's even more challenging, because the last dated entry is March 5 (which is now in the past) and shows status "discussed" instead of "done" or at least "scheduled" which *seem* to make more sense for a date in the past....

meetings/2024-03-06.md
```diff
@@ -0,0 +1,142 @@
+# W3C Solid Community Group: Weekly
+
+* Date: 2024-03-06T14:00:00Z
+* Call: https://meet.jit.si/solid-cg
+* Chat: https://matrix.to/#/#solid_specification:gitter.im
+* Repository: https://github.com/solid/specification
+* Status: Draft
```
csarven, line 7:
```suggestion
* Status: Published
```

csarven (commented):
@hzbarcea , you can batch the suggestions from the "Files changed" tab ( https://github.com/solid/specification/pull/633/files ) and then make a single commit. That way, only a single (email) notification is sent out where individually accepting the suggestions sends out separate notifications. The former would be less annoying and would be sufficient for something like the minutes I think.

Also... when you finally merge the minutes, please Squash them.

VirginiaBalseiro (approved):

meetings/2024-03-06.md
hzbarcea: Thanks @csarven, I already made this change yesterday: https://github.com/solid/specification/commit/7dff115fd356f3c5a823363dd76865ffc2b985ba. Noted the squashing request.

TallTed:
Still titled "Meeging"...

Should be titled "Meeting"...

michielbdejong:
I'm seeing:
> Merging is blocked
> The base branch restricts merging to authorized users.

@VirginiaBalseiro can you see if you can fix that?

VirginiaBalseiro:
@michielbdejong can you check now? I updated one rule that might be related. If not, then I am not sure what is wrong.

michielbdejong:
That worked apparently, thanks @VirginiaBalseiro!

Issue #70, opened 2019-09-25
Content-Type MUST be used on all PUT/POST/PATCH requests?


kjetilk:
The [LDP spec](https://www.w3.org/TR/ldp/#ldpc-post-contenttype) says 
> 5.2.3.6 LDP servers SHOULD use the Content-Type request header to determine the request representation's format when the request has an entity body. 

I'm not sure how the SHOULD has been justified there, since I struggle to see any other valid way, so it seems we should have a MUST there.

pmcb55:
Couldn't the server use content sniffing to 'guess' the content type? I'm not sure if that counts as a 'valid way', but it seems to have been common enough in the past I believe...

RubenVerborgh:
> Couldn't the server use content sniffing to 'guess' the content type?

Security issue; let's not due that. Agree on MUST.

csarven:
[Saw this on my phone.. literally rushed to get my laptop to respond.. Ruben beat me to it by a second.]

https://tools.ietf.org/html/rfc7231#section-3.1.1.5 is fairly clear about potential issues around content sniffing.

csarven:
MUST raises the bar - so to speak from, SHOULD - but it will remove potential issues and any ambiguity. That's a quality on its own and what to expect from clients. So, +1 to MUST.

pmcb55:
Yeah, +1 to MUST from me too (I was only trying to point out that content sniffing was perhaps another 'valid way', I certainly wasn't trying to suggest that we ever rely on it, absolutely not :) !)

csarven:
For the sake of documentation and to offer an answer for why SHOULD and to eventually resolve this issue:

The server may not want to completely rely on what a client claims in the request. It gives a way out I suppose to do some verification. For authenticated agents, this would be less of a concern.

acoburn:
The `SHOULD`-level requirement in LDP is used in the context of clients sending `POST` requests -- requests with an entity body -- to a server. If this is increased to `MUST`, what are the implications if a client does not sent a `Content-Type` header?

For example, if a client attempts to create an LDP-NR and does not send a `Content-Type`, does that mean that the request must be rejected? If the request is accepted, does that mean that subsequent responses for that resource cannot use, for example, `Content-Type: application/octet-stream`?

kjetilk:
> For example, if a client attempts to create an LDP-NR and does not send a `Content-Type`, does that mean that the request must be rejected?

[...]

PR #660, opened 2024-05-27
Clarify requests including content; container creation; omit slug targetting auxiliary

csarven:
[Preview](http://htmlpreview.github.io/?https://github.com/solid/specification/blob/0e820c87437fa3f1509a169da8266733cd86edae/protocol.html) | [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F93901540e7efd22ce48692deb4298df30ea4c2a7%2FED%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F0e820c87437fa3f1509a169da8266733cd86edae%2FED%2Fprotocol.html)

---

Summary:

* This PR clarifies the Solid Protocol to enhance internal consistency and avoid conflicting implementations.
* Requests without a `Content-Type` header, particularly when the request message includes content, can potentially confuse servers and cause issues with existing or future implementations.
* `PUT` and `PATCH` requests can be used to create a container if the target URI ends with a slash (`/`) and the message content is empty, even without the `Content-Type` or `Link` headers indicating an LDP container.
* `POST` requests to create a container essentially depends on the `Link` header indicating an LDP container (or possible alternative considerations that were contemplated in the past).

This PR makes a [correction class 3 change](https://www.w3.org/2023/Process-20231103/#class-3):
 * Update requirement https://solidproject.org/ED/protocol#server-content-type-missing to clarify message content.
 * Update requirement https://solidproject.org/ED/protocol#client-content-type-includes to clarify message content.
 * Remove requirement https://solidproject.org/ED/protocol#server-post-container-create-container as creating a container with a server-assigned URI is not significant and relied on the LDP interaction model.
 * Remove requirement https://solidproject.org/ED/protocol#server-post-slug-auxiliary-resource as the `Slug` header is not widely implemented and should not be used to match existing resources ( https://github.com/solid/specification/pull/621 ).

---

A more high level explanation of the changes:

Servers conforming to the Solid Protocol can create a container using `PUT` and `PATCH` requests without the `Content-Type` header, provided that the message content is empty. Similarly, LDP server conformance generally requires the use of a `Content-Type` header when there is message content, and the behaviour without a `Content-Type` header is unspecified.

Where a server accepts a `POST` request with a `Link` header indicating the LDP container type but without a `Content-Type` header, the server could potentially conform to the Solid Protocol (given changes to the specification) but no conformance to LDP is unspecified.

The use case for servers assigning a URI for containers (e.g., `/{uuid}/`) with `POST` requests is generally uncommon.

To avoid potential conflicting requirements across specifications, it would be preferable not to specify that `POST` requests can create containers, regardless of the `Content-Type` header being omitted when message content is empty. LDP servers are not required to create containers that way either. This can make sure that Solid servers remain compatible with LDP servers without necessarily restricting container creation.

---

Background drawing from HTTP specs, LDP, and Solid Protocol specs:

In https://www.rfc-editor.org/rfc/rfc9110#section-9.3.3-1 :

>The POST method requests that the target resource process the representation enclosed in the request according to the resource's own specific semantics.

>The PUT method requests that the state of the target resource be created or replaced with the state defined by the representation enclosed in the request message content.

Server's expectation is that `Content-Type` is included in the request (irrespective to resource semantics).

In https://www.rfc-editor.org/rfc/rfc9110#section-8.3-5 :

>A sender that generates a message containing content SHOULD generate a Content-Type header field in that message unless the intended media type of the enclosed representation is unknown to the sender. If a Content-Type header field is not present, the recipient MAY either assume a media type of "application/octet-stream" ([RFC2046], Section 4.5.1) or examine the data to determine its type.

At the moment, the Solid Protocol tries to be strict on this point https://solidproject.org/ED/protocol#server-content-type-missing by requiring `Content-Type`, and generally on the basis of discouraging heuristics to figuring out representation's type ( https://www.rfc-editor.org/rfc/rfc9110#section-8.3-6 ). LDP is however more relaxed on this (see below)

While LDP does not generally describe or prohibit an interaction model to create a container ( https://www.w3.org/TR/ldp/#ldpc-post-createrdf ) with `POST` without `Content-Type`:

>5.2.3.4 LDP servers that successfully create a resource from a RDF representation in the request entity body MUST honor the client's requested interaction model(s). If any requested interaction model cannot be honored, the server MUST fail the request. 

>If the request header specifies a LDPC interaction model, then the server MUST handle subsequent requests to the newly created resource's URI as if it is a LDPC.
>This specification does not constrain the server's behavior in other cases.

it does explain the process to create LDP-RSs where it also includes the `Content-Type` header, emphasis on "creation of LDP-RS", which includes all LDP container types ( https://www.w3.org/TR/ldp/#ldpc-isldpr ):

https://www.w3.org/TR/ldp/#ldpc-post-turtle :

>5.2.3.5 LDP servers that allow creation of LDP-RSs via POST MUST allow clients to create new members by enclosing a request entity body with a Content-Type request header whose value is text/turtle [turtle].

https://www.w3.org/TR/ldp/#ldpc-post-jsonld :

>5.2.3.14 LDP servers that allow creation of LDP-RSs via POST MUST allow clients to create new members by enclosing a request entity body with a Content-Type request header whose value is application/ld+json [JSON-LD].

Essentially LDP does not specify conformance for `POST` requests without `Content-Type` and without message content. No interoperability between servers and clients specified. It doesn't even specify for `PUT`.

LDP also notes the possibility that servers might infer the content type (when the `Content-Type` header is missing) by sniffing ( https://www.w3.org/TR/ldp/#ldp-http-content-sniffing ), which is also summarised in https://github.com/solid/specification/issues/70#issuecomment-536684883 :

>6.2.6 When the Content-Type request header is absent from a request, LDP servers might infer the content type by inspecting the entity body contents ([RFC7231] section 3.1.1.5).

While content type sniffing is discouraged in Solid Protocol ( https://github.com/solid/specification/issues/70#issuecomment-547924171 ) the context of the discussion and the requirements that lead to it in the Solid Protocol was for requests including a message. It is important to maintain that case. However, when the request does not include content, and the request semantics does not require the `Content-Type` header, then it is possible to have a valid request without the `Content-Type` header.

---

Solid Protocol defines a container as a hierarchical collection of resources ( https://solidproject.org/ED/protocol#container-resource ) - not limited to LDP:

>A container resource is a hierarchical collection of resources that contains other resources, including containers.

but the described normative behaviours correspond to LDP Basic Container ( https://solidproject.org/ED/protocol#server-basic-container ):

>The representation and behaviour of containers in Solid corresponds to LDP Basic Container and MUST be supported by server.

With resource type heuristics ( https://solidproject.org/ED/protocol#resource-type-heuristics ) in Solid Protocol, it is possible to create a container without a `Content-Type` header if the request's target URI ends with a slash (`/`):

>When creating new resources, servers can determine a target URI’s type by examining the URI path ending (URI Slash Semantics).

https://solidproject.org/ED/protocol#uri-slashes-hierarchical-identifier :

>The slash (/) character in the URI path indicates hierarchical relationship segments, and enables relative referencing [RFC3986]. The semantics of the slash character is shared by servers and clients. Paths ending with a slash denote a container resource.

The ways in which PUT, PATCH, and POST can be used to create containers and URI assignment are based on determining the request semantics from HTTP request method ( https://www.rfc-editor.org/rfc/rfc9110#section-9 ). Summarised in:

https://solidproject.org/uri-allocation-put-patch :

>Clients can use PUT and PATCH requests to assign a URI to a resource.

https://solidproject.org/uri-allocation-put-patch :

>Clients can use POST requests to have the server assign a URI to a resource.

https://solidproject.org/ED/protocol#server-post-target-not-found

>When a POST method request targets a resource without an existing representation, the server MUST respond with the 404 status code.

Essentially only `PUT` and `PATCH` can be used to create a container in such requests.

---

Additional background on determining resource type was outlined here https://github.com/solid/specification/pull/160#issuecomment-635932724 , based on use cases borrowed from LDP where 1) the requirements building on from HTTP (as opposed to LDP), and 2) aiming to not conflict with LDP and be compatible as needed.

As it currently stands in the Solid Protocol, the only interoperable way for Solid Protocol to create a container with `POST` is by:

>C3.1: request including Link header with rel="type" targeting a valid LDP container type [F1.2].

where it only meets the functional requirement:

>F2.2: The system shall provide the ability to use server's preferred naming.

The `Slug` header is not required in Solid Protocol or LDP, and it is deemed to be unreliable or problematic ( https://github.com/solid/specification/pull/621 ).


ED/protocol.html
```diff
@@ -746,7 +746,7 @@ <h3 property="schema:name">HTTP Server</h3>
...
+                  <p about="" id="server-content-type-missing" rel="spec:requirement" resource="#server-content-type-missing"><span property="spec:statement"><span rel="spec:requirementSubject" resource="#Server">Server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> reject <code>PUT</code>, <code>POST</code> and <code>PATCH</code> requests including content but without the <code>Content-Type</code> header with a status code of <code>400</code>.</span> [<a href="https://github.com/solid/specification/issues/70#issuecomment-547924171" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
CxRes, line 749: Suggest making the language more RFC9110 like and also include validity as a condition:

"On receiving a message containing content but not including a content-type header field with a valid media type in a PUT, POST or PATCH request, the server MUST reject the request with a status code of 400."

ED/protocol.html
```diff
@@ -757,7 +757,7 @@ <h3 property="schema:name">HTTP Client</h3>
...
+                  <p about="" id="client-content-type-includes" rel="spec:requirement" resource="#client-content-type"><span property="spec:statement"><span rel="spec:requirementSubject" resource="#Client">Clients</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> use the <code>Content-Type</code> HTTP header in <code>PUT</code>, <code>POST</code>, and <code>PATCH</code> requests including content [<cite><a class="bibref" href="#bib-rfc9110">RFC9110</a></cite>].</span> [<a href="https://github.com/solid/specification/issues/70#issuecomment-547924171" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
CxRes, line 760: Suggest making the language more RFC9110 like and also include correctness of media-type as a condition:

"When sending a PUT, POST and PATCH requests with a message containing content, a client MUST generate a as part of the message the Content-Type header field with the media type of the content".

CxRes (changes requested):
I would have ideally preferred two commits for this, but it is not a big deal.

I have suggested changes where the requirement on the client is stricter than the server wrt content-type field (thereby server need not be obligated to match content-type with content), in keeping with Postel's law.

CxRes:
PS: the HTTP Police aka HTTPWG has ruled that it is not *header* but *field* or *header field*. I guess we'll need a separate PR for that!

ED/protocol.html
csarven: That seems like proposing a new requirement, "I guess we'll need a separate PR for that!"

But here is why I think it'd be preferable to not introduce that:

RFC doesn't explicitly require recipients to verify the content against the `Content-Type` header field value. Some servers may do that for practical reasons, if performance is not an issue, or if precision is expected by the server before storing the state, and so on. By definition, the `Content-Type` header field indicates the intended media type. It doesn't need to be further spelled out beyond what the RFC says.

We are highly unlikely to have a complete test coverage taking input payload and media type and checking whether they match, and this gets potentially more complex with polyglot formats.

Validity of the `Content-Type` header field value as a condition would in fact be a separate requirement than what's originally intended by the #server-content-type-missing requirement, and it would not entail `400` when the request is rejected either.

In cases that "the origin server is refusing to service the request because the content is in a format not supported by this method on the target resource", the 415 status code can be used.

So, I do not think that the requirement needs to validating the `Content-Type` header field-value against the content. It is already implicitly covered.

---

I don't think this (or any) requirement should be more RFC9110, not that there is anything wrong with that. The language in this document is aimed to be internally consistent first and foremost, and the conformance clause template that's in use is along the lines what we see right now. There are variations of course throughout the spec.

I have changed "including" to "containing" and added "field" to the text as per your suggestion. Thanks.

ED/protocol.html
csarven: I don't think this (or any) requirement should be more RFC9110, not that there is anything wrong with that. The language in this document is aimed to be internally consistent first and foremost, and the conformance clause template that's in use is along the lines what we see right now. There are variations of course throughout the spec.

I have changed "including" to "containing" and added "field" to the text as per your suggestion. Thanks.

csarven:
Thanks for the heads up. The specification already uses "header field" (as well as "field value") in some parts so I guess some instances were missed. I went ahead and made that change to clarify in https://github.com/solid/specification/commit/93901540e7efd22ce48692deb4298df30ea4c2a7 and it is integrated in this PR.

ED/protocol.html
CxRes: I think you misunderstood my worry here, I should have explained this better.

RFC9110, 8.3 in the paragraph quoted, is a requirement for the clients to provide a content-type field. Translating this for a server, not only should the content-type field be present, but it also must be in the correct syntax. 

It is NOT my intention to impose verification requirements on the server that the content-type matches the content. This is a purely header field syntax check!

ED/protocol.html
CxRes: This definitely is an extra requirement on the client. My reasoning being that if we get rid of sniffing completely, and the server does not (have to) have the possibility of checking the provided content type which it now MUST process (as opposed to SHOULD in RFC9110), the burden is entirely on the client to provide the right content-type and not expect the server to recover from the wrong one. This has to be made explicit.

CxRes: Having mention this being an extra requirement, I still think it belongs here because it is intimately tied to the change from SHOULD to MUST.

ED/protocol.html
csarven: Fair enough, but I think that level of correctness is already expected of Solid servers: https://solidproject.org/ED/protocol#http-server . And even more so that I've raised it as a significant security consideration with some advisement: https://solidproject.org/ED/protocol#http-request-headers-validity . 400 is indeed expected in these cases.

ED/protocol.html
csarven: The general point is that the sender of the message (i.e., both the server and the client) needs to know what they are sending. The context of the sniffing in RFC and in Solid Protocol is limiting or eliminating the MIME sniffing cases. When the client includes the Content-Type header field and a field value, the server is not forced into a situation (code path) to sniff or reject outright.

Surely the server already has, and needs to have, all the rights to inspect, process, and handle errors, as it deems necessary. That's not the same as "sniffing".

CxRes: Sure, server can reject any request it does not like. The point here is that client due to the change now has a mandatory obligation to ensure media-type correctness, which was not the case before 

Again lets look at the RFC9110 8.3-5,
> A sender that generates a message containing content SHOULD generate a Content-Type header field in that message unless the intended media type of the enclosed representation is unknown to the sender.

When we change the SHOULD in RFC9110 to a MUST in Solid specification, the exception provided to the client of not knowing the media type of the content no longer applies. The client hence has that additional burden over that of HTTP clients of providing the correct media-type which we should make explicit.

ED/protocol.html
CxRes: I think there is value to making this explicit to spec readers here. Rather than have them co-relate this in conjunction with security consideration (which is non-normative) and say, leading to a scenario where even one implementor amongst many will tolerate not well-formed content-type header, which is certainly not ideal. It also demands that the test suite include this case as an explicit test.

Thinking about it, I would ideally want servers to be strict not just on syntax but also on the iana registered types (but not subtypes) when processing the content type headers. Not sure how to work that in or even if that is advisable?

csarven: Correct use of HTTP is already required for interoperability:

* https://solidproject.org/ED/protocol#server-http
* https://solidproject.org/ED/protocol#client-http

Full stop.

*Advisements* under (security) considerations does not take away from *requirements*.

Whether an implementation of the Solid https://solidproject.org/ED/qa#TestSuite has its own https://solidproject.org/ED/qa#TestCase s checking HTTP conformance or imports an existing one is orthogonal to this matter. A conforming Solid Protocol product means that the outcome of applicable and testable tests is that the given product meets all the requirements.

ED/protocol.html
csarven: Again, correct use of HTTP is already required for interoperability:
* https://solidproject.org/ED/protocol#server-http
* https://solidproject.org/ED/protocol#client-http

And Conformance to HTTP is given: https://www.rfc-editor.org/rfc/rfc9110#name-conformance

Changing the requirement level from SHOULD to MUST does not constitute a harder, better, faster, stronger adherence to the definition of `Content-Type` header field and field value, or "media-type correctness" (or for any other pat of HTTP.) It is a given.

ED/protocol.html
CxRes: Sorry, but a RFC9110 compliant server need not send a 400 error code when sent a request with a bad/illegal `content-type` header, it can ignore it and send a 2xx response. And since Solid server extends an HTTP server, a Solid server can ipso-facto ignore it too (not withstanding the MUST for specifying a `content-type` header, which merely says that the header field has to be there not that field value has to be legal). Hence, this needs to be explicitly constrained.

This is not about just correct use of HTTP, but application of a rule that is more stringent in Solid than HTTP, which must be thoroughly specified.

csarven: I see no practical or theoretical benefit to further constraining the media type in the Solid Protocol. Just because a client includes a "legal" media type - whatever that is - it does not entail that a server accepts it. The server already has the liberty to accept or reject a request for various reasons and it is useful to maintain that right. Some servers may only acknowledge IANA registered media types, while others may allow other media types, yet some other servers may not want to handle multimedia formats, and some may be specialised to to super cool things with them, and so on. 415 exists. This is always a negotiation between the server-client pair, and it does not conflict with the kind of interoperability we can expect from the Solid Protocol.

michielbdejong (approved):

VirginiaBalseiro (approved):

ED/protocol.html
CxRes: > Just because a client includes a "legal" media type - whatever that is...

Do not strawman the argument. I am speaking of an illegal `content-type` header, not `illegal` media-type. As I have said, "_The suggested is a purely header field syntax check_". 

For IANA registered types (not even subtypes), I already indicated that it is a preference, which I am not sure how that would even work.  This was a separate musing (in a separate paragraph) from the request of making sure that the `content-type` header is syntactically correct.

> The server already has the liberty to accept or reject a request for various reasons and it is useful to maintain that right.

So you want to maintain the right of some random Solid server to accept a request with an illegal `content-type` header despite the change in Solid protocol. I cannot think of one situation where that is ever useful.

Yes, I want to constrain the server behaviour because it reflects the constraint added to client behaviour by Solid protocol, a constraint that HTTP itself does not have.

csarven: Excuse me, but I've been trying very hard to understand your concerns, and it still doesn't make sense to me.

You're using terminology like "valid media type," "correct syntax," "head field syntax check," "well-formed content-type header," "field value has to be legal", "syntactically correct." However, you also seem to find the HTTP definitions, such as the ABNF notations and conformance explanations, inadequate for defining what constitutes "valid," "correct," "legal," or similar terms for both the header field and its field value. You've been jumping back and forth, in addition to suggesting that the requirement level is relevant. When I tried to understand your statement "field value has to be legal" - again, whatever "legal" means (perhaps you should define it?) - with the possibility that there is a finite set of media types or categories of them, you accused me of straw-manning.

How about them straw-manning:

>So you want to maintain the right of some random Solid server to accept a request with an illegal content-type header despite the change in Solid protocol. I cannot think of one situation where that is ever useful.

How did you conclude that I want the server to accept an "illegal content-type header" when I literally argued earlier that there is nothing for the Solid Protocol to do here since it is building on HTTP by definition, and that products either conform to HTTP or not?

https://www.rfc-editor.org/rfc/rfc9110#section-5.1-3

https://www.rfc-editor.org/rfc/rfc9110#section-8.3-2

```
Content-Type = media-type
```

follow the white-rabbit for conformance.

So, at this juncture, I'm not sure if there is any new information I can add to this thread in this particular PR. Perhaps we need more bandwidth to discuss or draw things on a whiteboard. If you feel this is inadequate, please create a new issue and make your case there, kick off an STM, etc.

acoburn (approved):
Overall, this is a good clarification.

Also +1 on removing the link header requirement on POST-to-create-container operations.

ED/protocol.html
CxRes: Our disagreement here boils down to a very simple interpretation of how a change in client (sender for PUT, POST, PATCH) requirement translates into a change in server requirements. Where you reverse the Strawman accusation on me is exactly the thing I find to be the problem of interpretation. AFAICT, nothing in plain RFC9110 stops a server from processing a message with a `content-type` type header field that contains anything other than a media type and sending a 2xx response. Due to the product class definition, I would presume that would extend to a Solid server unless explicitly prohibited. You OTOH think that the CoP definition of server somehow also demands that the server sends a 400 in that case, which I do not follow! 

Maybe a higher bandwidth communication will help, I am open to that. But my worry is that, if this is an interpretation of formal language, that bandwidth will not do much to help. 

PS: You may feel that you are trying really hard to understand me and knowing you, I believe that you are, but the kind of "escalation dominance" that you engage in does not help!

ED/protocol.html
```diff
@@ -746,7 +746,7 @@ <h3 property="schema:name">HTTP Server</h3>
...
+                  <p about="" id="server-content-type-missing" rel="spec:requirement" resource="#server-content-type-missing"><span property="spec:statement"><span rel="spec:requirementSubject" resource="#Server">Server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> reject <code>PUT</code>, <code>POST</code> and <code>PATCH</code> requests containing content but without the <code>Content-Type</code> header field with a status code of <code>400</code>.</span> [<a href="https://github.com/solid/specification/issues/70#issuecomment-547924171" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
TallTed, line 749:
```suggestion
                  <p about="" id="server-content-type-missing" rel="spec:requirement" resource="#server-content-type-missing"><span property="spec:statement"><span rel="spec:requirementSubject" resource="#Server">Server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> reject <code>PUT</code>, <code>POST</code>, and <code>PATCH</code> requests that contain content but lack the <code>Content-Type</code> header field, with a status code of <code>400</code>.</span> [<a href="https://github.com/solid/specification/issues/70#issuecomment-547924171" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

ED/protocol.html
```diff
@@ -757,7 +757,7 @@ <h3 property="schema:name">HTTP Client</h3>
...
+                  <p about="" id="client-content-type-includes" rel="spec:requirement" resource="#client-content-type"><span property="spec:statement"><span rel="spec:requirementSubject" resource="#Client">Clients</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> use the <code>Content-Type</code> HTTP header field in <code>PUT</code>, <code>POST</code>, and <code>PATCH</code> requests containing content [<cite><a class="bibref" href="#bib-rfc9110">RFC9110</a></cite>].</span> [<a href="https://github.com/solid/specification/issues/70#issuecomment-547924171" rel="cito:citesAsSourceDocument">Source</a>]</p>
```
TallTed, line 760:
```suggestion
                  <p about="" id="client-content-type-includes" rel="spec:requirement" resource="#client-content-type"><span property="spec:statement"><span rel="spec:requirementSubject" resource="#Client">Clients</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> use the <code>Content-Type</code> HTTP header field in <code>PUT</code>, <code>POST</code>, and <code>PATCH</code> requests that contain content [<cite><a class="bibref" href="#bib-rfc9110">RFC9110</a></cite>].</span> [<a href="https://github.com/solid/specification/issues/70#issuecomment-547924171" rel="cito:citesAsSourceDocument">Source</a>]</p>
```

TallTed (changes requested):
Editorial tweaks for clarity and consistency

PR #661, opened 2024-05-29
Add meeting minutes 2024-05-28 (Special Topic Meeting)

VirginiaBalseiro:


meetings/2024-05-28.md
```diff
@@ -0,0 +1,165 @@
...
+* VB: +1
+* SC: Some specs say interop is defined by this or that product. Thinking along those lines help figure out what needs to communicate. Solid protocol is just server and client. Then Solid notification we have a nicer way of expressing.. w have 5 products, sender receiver subscriber.. There's an issue for solid spec where we need a bit more detail on these products: https://github.com/solid/specification/issues/480
+* SC: Right now we say server and client but maybe it needs more details. Consumer, storage.. thing doing the patching.. you can combine these in different ways. In other specs we have authz server, application, and so on.
+* VB: WebID Profile has Reader Application, Writer Application, and so on.
+* SC: Maybe we say app that lives in the browser, app that lives in the seerver.. Ways of being more specific about the types of products that could be built. If we get stuck on just server or authn server, we are going to assume what that product is capable of. Stuck on whether it should have issuer, should not...
+* VB: +1 
+* eP: Prefer we use established terms. oAuth have well established terms.
```
TallTed, line 161:
```suggestion
* SC: Some specs say interop is defined by this or that product. Thinking along those lines helps figure out what needs to be communicated. Solid protocol is just server and client. Then we have a nicer way of expressing Solid notification. We have 5 products, sender, receiver, subscriber... There's an [issue for solid spec](https://github.com/solid/specification/issues/480) where we need a bit more detail on these products.
* SC: Right now we say server and client, but maybe it needs more details. Consumer, storage, thing doing the patching... You can combine these in different ways. In other specs we have authz server, application, and so on.
* VB: WebID Profile has Reader Application, Writer Application, and so on.
* SC: Maybe we say app that lives in the browser, app that lives in the seerver... Ways of being more specific about the types of products that could be built. If we get stuck on just server or authn server, we are going to assume what that product is capable of. Stuck on whether it should (or should not) have issuer...
* VB: +1 
* eP: Prefer we use established terms. oAuth has well established terms. 
```

TallTed (changes requested):

PR #684, opened 2024-09-17
Add agenda 2024-09-18

VirginiaBalseiro:

### DIFF
--- a/meetings/2024-09-18.md
+++ b/meetings/2024-09-18.md
@@ -4,13 +4,23 @@
 * Call: https://meet.jit.si/solid-cg
 * Chat: https://matrix.to/#/#solid_specification:gitter.im
 * Repository: https://github.com/solid/specification
-* Status: Draft
+* Status: Published
 
 
 ## Present
 * [Virginia Balseiro](https://virginiabalseiro.com/#me)
-* 
-
+* [elf Pavlik](https://elf-pavlik.hackers4peace.net)
+* Hadrian Zbarcea
+* Willem Datema
+* Grace Elcock
+* [Rahul Gupta](https://cxres.pages.dev/profile#i)
+* Maxime Lecoq-Gaillard
+* Michiel de Jong
+* Jeff Zucker
+* Tim Berners-Lee
+* Laurens Debackere (from 14:30UTC)
+* [Pierre-Antoine Champin](https://champin.net/#pa) (from 14:45UTC) 
+* [michal/mrkvon](https://id.mrkvon.org)
 ---
 
 ## Announcements
@@ -30,12 +40,12 @@
 
 
 ### Scribes
+* Grace Elcock
 * 
 
 
 ### Introductions
 * [name]:
-
 
 ---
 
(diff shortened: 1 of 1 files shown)


meetings/2024-09-18.md
```diff
@@ -30,22 +40,95 @@
...
+* LD: Yes, difficult next week. Don't have to be present but would like to join
+* VB: Best to skip so have people's undivided attentions.
+* eP: Agree to have this discussion in 2 weeks. Could be opportunity next week to get input re use cases from people present during TPAC.
+* VB: Sounds great - plese add to the agenda
+* MdJ: Would be amazing to ask them to add but in genral have more use cases than we can serve so main effort is to document and label so when we have a discussion we can refer to them.
+* LD: Looking forward to working on this and starting from a good basis of use cases and requirements.
+* eP: Impression on pproaching use cases in the past - use cases should not suggest specific solution, they should just describe the problem. 
+* TBL: SolidOS accumulated over many years. Hpw should we connect together the scheduling and calendar...way things connect in Solid is really interesting to keep track of as that is where Solid delivers value.
+* eP: With Hadrian met w Data Food Consortium to say there are existing applications that would benefit from Solid so don't need new use cases i.e. extract from what exists than imagine new things
```
TallTed, line 111:
```suggestion
* MdJ: Would be amazing to ask them to add, but I think we have more use cases than we can serve, so main effort is to document and label, so when we have a discussion, we can refer to them.
* LD: Looking forward to working on this and starting from a good basis of UCR.
* eP: Impression on approaching use cases in the past — use cases should not suggest a specific solution, they should just describe the problem. 
* TBL: SolidOS accumulated over many years. How should we connect the scheduling and calendar together? The way things connect in Solid is really interesting to keep track of, as that is where Solid delivers value.
* eP: With Hadrian, met with Data Food Consortium, to say there are existing applications that would benefit from Solid, so don't need new use cases, i.e., extract from what exists, than imagine new things.
```

meetings/2024-09-18.md
```diff
@@ -30,22 +40,95 @@
...
+
+### Solid catalog
+
+URL: https://github.com/solid-contrib/catalog
+URL: https://github.com/elf-pavlik/solid-efforts
+
+eP: I plan to merge the dataset used by solid-effort into the catalog
+eP: We will have two different apps using it, Jeff's webcomponents based one and the Vue + LDO I started prototyping.
+eP: Let's meet next week for focused discussion. We can also invite people during the TPAC to explore and PR their efforts/inititives.
+JZ: I'd like to demo the catalog and say a bit about its intention
```
TallTed, line 122:
```suggestion
eP: I plan to merge the dataset used by solid-effort into the catalog.
eP: We will have two different apps using it — Jeff's webcomponents-based one and the Vue + LDO I started prototyping.
eP: Let's meet next week for focused discussion. We can also invite people to explore and PR their efforts/initiatives during TPAC.
JZ: I'd like to demo the catalog and say a bit about its intention.
```

meetings/2024-09-18.md
```diff
@@ -30,22 +40,95 @@
...
+* Have subcategories of things within communication channels. Have forum, chat and mailing list all representes. This call and others listed.
+* Category 'services for specific communities' - i.e. have dedicated pod server (there are many). each presents a pod for a particular set of users
+* Servers not done yet but will include keywords etc.
+* If wanted to look at only calendars then can show calendar apps. If wanted to look at 'bookmarks' then shows number of different bookmark apps
+* Can do a full text or fielded search
+* Developer tools - need to figure out how to sub categorise them 
+* Specifications - and vocabularies - each section will have its' own shape
+* Resources offered - was tracked at one point on solidproject.org
```
TallTed, line 132:
```suggestion
JZ: Some people have seen this, so apologies for that. Re intention, relation to what this is. Think about names for what we are doing — eP called his Solid Efforts (what people already involved in Solid have done). JZ is Solid resources (what anyone in Solid can use). Different approaches to the same thing — this one about supporting bigger community, etc., that need to focus on the practicality and what is occurring in the community vs. how things theoretically fit together. Think important to head in this direction as well as direction eP is going in.
  * Based on RDF data source, and uses web components to draw information out. Everything is based on an RDF SHACL file and a set of sub-classes. All of the major categories seen in 'table contents' and the tabs are derived from classes of the major types that are used. The idea is that for each kind of service, we have the name, type of thing, and links that go outside of the catalogue and highlighted like the service endpoint. This brings you to Inrupt pod sources. If looking at ESF, then can see it is used by Inrupt pod spaces, and then see it uses Solid server. For subcategories, we're going to need to figure out what the main things we use are, and talk about how people can find them.
  * Have subcategories of things within communication channels. Have forum, chat, and mailing list all represented. This call and others listed.
  * Category 'services for specific communities' — i.e., have dedicated pod server (there are many). Each presents a pod for a particular set of users.
  * Servers not done yet but will include keywords, etc.
  * If wanted to look at only calendars, then can show calendar apps. If wanted to look at 'bookmarks', then shows number of different bookmark apps.
  * Can do a full text or fielded search.
  * Developer tools — need to figure out how to sub categorise them.
  * Specifications — and vocabularies — each section will have its own shape.
  * Resources offered — was tracked at one point on solidproject.org.
```

TallTed (changes requested):

PR #685, opened 2024-09-23
Fix typo: Link to LDP, not LDN

Peeja:
Easy typo to make. 😛


csarven:
Good catch, thank you! I've update 2024/20240512-protocol and ED/protocol as well: https://github.com/solid/specification/commit/97abf1874af067a2c5e3a4b352176c6f8ff07b2f

Normally, changes first go into ED/protocol but for these kind of changes they can be corrected on the published versions - same as W3C publishing rules.

PR #690, opened 2024-10-23
2024-10-23 meeting notes

michielbdejong:


meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+* RG: You need to be able to communicate when there is an error between the resource server and the notficiation centre. You need to communicate that to the client, should be RDF, some discussion on GH, but wanted to bump here so that anyone with an opinion can chime in.
+* Elf: We need some more specific use cases and requirments from the perspective of the client. We need to arrive at the interface we want based on who consumes it. We have already arrived at a general interface based on the fact that error was unreliable, for instance may know that resource needs to be re-requested if the error was unreliable. This is because the notification system knows that it could have missed some updates. If you know that there are some scenarios where you would like an error to happen, then identify what you would have like to have happened.
+ * RG: I've begun that process by putting what those generic requirements might look like in terms of classes of products.
+ * Elf: I've been doing webhooks in the last few months, if we are doing errors then it impacts the notification recievers as they need to check if at as error message etc if errors are included with notifications.
```
TallTed, line 95:
```suggestion
* RG: We need a unified pattern for notification errors.
* RG: You need to be able to communicate when there is an error between the resource server and the notification centre. You need to communicate it to the client, should be RDF, some discussion on GH, but wanted to bump here so that anyone with an opinion can chime in.
* Elf: We need some more specific use cases and requirements from the perspective of the client. We need to arrive at the interface we want based on who consumes it. We have already arrived at a general interface based on the fact that error was unreliable; for instance, may know that a resource needs to be re-requested if the error was unreliable. This is because the notification system knows that it could have missed some updates. If you know that there are some scenarios where you would like an error to happen, then identify what you would have liked to happen.
 * RG: I've begun that process by putting what those generic requirements might look like in terms of classes of products.
 * Elf: I've been doing webhooks in the last few months. If we are doing errors, it impacts the notification receivers, as they need to check for error messages, etc., if errors are included with notifications.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Activity Mapping for Notifications
+
+https://github.com/solid/notifications/issues/183
+
+* RG: Added a new column; need to revisit this. Looking at what are the container mappings, because this is something we need to revisit soon, the purpose of this issue was to have an [?? office hours STM??]
+* eP: Should not map to HTTP Headers. Solid or LWS should define set of operations and notification system hooks into set of operations. Then there will be HTTP interface that is binding this to HTTP methods but notif. system should not bind directly to that. Relevant to discuss in relation to access control and notif system, both should have a common layer of internal operations supported by the storage. That is something LWS may need to take care of by making those operations defined
```
TallTed, line 102:
```suggestion
* RG: Added a new column; need to revisit this. Looking at the container mappings, because this is something we need to revisit soon. The purpose of this issue was to have an [?? office hours STM??]
* eP: Should not map to HTTP Headers. Solid or LWS should define a set of operations, and the notification system hooks into the set of operations. Then there will be an HTTP interface that binds this to HTTP methods by notifications. System should not bind directly to that. Relevant to discuss in relation to access control and notification system; both should have a common layer of internal operations supported by the storage. That is something LWS may need to take care of by defining those operations.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+* TBL: Tricky thing about notifications is that people generalise. People say, e.g. don't just standardise a version of websockets or "there is a world of notification out there and we should model them all in our notifications". Can be unhelpful by running into danger of assuming that there will be lots of different notif features that will have too much in common, e.g., if user moves mouse into space you'll get a notif - lots of cases where the timescale of things are very different. I hope that we have a very effective & test notif. systems for what I need in SolidOS where we can update the display when someone changes something. Interested in relation to local-first and CRDTs, if you're making a CRDT you cant be ??? about it. I hope if we are trying to make something that works at CRDT level + at higher levels we don't end up causing confusion. Lets make sure that notifs at one level work well, and then generalise.
```
TallTed, line 103:
```suggestion
* TBL: One tricky thing about notifications is that people generalise. People say, e.g., "don't just standardise a version of websockets", or "there is a world of notifications out there and we should model them all in our notifications". Can be unhelpful by running into the danger of assuming that there will be lots of different notification features that will have too much in common; e.g., if the user moves their mouse into a space, you'll get a notification; lots of cases where the timescales of things are very different. I hope that we have a very effective ??? & test notification systems for what I need in SolidOS, where we can update the display when someone changes something. Interested in relation to local-first and CRDTs, if you're making a CRDT you can't be ??? about it. I hope that if we are trying to make something that works both at CRDT level and at higher levels, we don't end up causing confusion. Let's make sure that notifications at one level work well, and then generalise.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+* MDJ: Question about linked issue - it sounds like a better idea to just use what is there rather than taking to office hours or take to working group
+* RG: A lot is already implemented e.g. in CSS, and there is already usage. But in implementation people found underspecified cases and need clearer requirements based on those things. We need to take another specification step to iron out those details.
```
TallTed, line 105:
```suggestion
* MDJ: Question about linked issue — it sounds like it's a better idea to just use what is there, rather than taking to office hours or to WG.
* RG: A lot is already implemented — e.g., in CSS — and there is already usage. But in implementation, people found underspecified cases and need clearer requirements based on those things. We need to take another specification step to iron out those details.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+* eP: We have created a crdt4rdf CG, no activity yet. Next week planning to send out emails and start organising. Please join if you're interested. https://www.w3.org/community/crdt4rdf/
+* RG: I agree mapping should be 1 step removed. In notifs we want to make sure that we have proper security, such that if access controls are changed to a parent resource, that is properly handled when we are dealing with the person recieving the notifications (you should be able to stop those notifs from coming in). For the table, we have been iterating it, and there are a bunch of question marks there. When we are implementing this, some case comes up that we are unsure how to deal with in that moment. More about exchanging notes and making sure we are using a common format for the messages. I also agree that we should be ???
```
TallTed, line 107:
```suggestion
* eP: We have created a crdt4rdf CG; no activity yet. Next week, we plan to send out emails and start organising. Please join if you're interested. https://www.w3.org/community/crdt4rdf/
* RG: I agree mapping should be 1 step removed. In notifications, we want to make sure that we have proper security, such that if access controls are changed to a parent resource, that is properly handled when we are dealing with the person receiving the notifications (e.g., you should be able to stop those notifications from coming in). For the table, we have been iterating it, and there are a bunch of question marks there. When we are implementing, some use case comes up that we are unsure how to deal with in that moment. More about exchanging notes and making sure we are using a common format for the messages. I also agree that we should be ???
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+* eP: Taking some actions - fits under LWS. Is there interest from the WG to at least see if it is fitting under the work, and making sure we are coordinating that in some way.
+* MDJ: Think we should go with the table that you created, if we need clarifications for some cases - then we should clarify. If there is a contradiction then we should discuss until we agree in the community group. Lets keep working with running code and when we run into a question like that we shouldn't wait on the WG to give us an answer as it will take 2 years. Should keep working on Solid as we have it, and make things interoperable on top of it. Will always be non-interopable things in an incubator and community, we should just document when that is the case and see if it converges. We cannot even really force anybody to converge - so should start with what people build, describe that, and then the WG can look, learn from that, and come up with a spec that hopefully aligns.
```
TallTed, line 109:
```suggestion
* eP: Taking some actions — fits under LWS. Is there interest from the WG to at least see if it is fitting under the work, and making sure we are coordinating that in some way?
* MDJ: I think we should go with the table that you created. If we need clarifications for some cases — then we should clarify. If there is a contradiction, then we should discuss until we agree in the CG. Let's keep working with running code, and when we run into a question like that, we shouldn't wait on the WG to give us an answer, as it will take 2 years. Should keep working on Solid as we have it, and make things interoperable on top of it. There will always be non-interoperable things in an incubator and community; we should just document when that is the case and see if it converges. We cannot even really force anybody to converge — so should start with what people build, and describe that; then the WG can look, learn from that, and come up with a spec that hopefully aligns.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+* eP: The table has a column operation that is not part of any existing specs - that is something where ?? and access are not defined. They need to be defined and added to spec
+* MDJ: Yes, but that doesn't contradict what I said. We just need to work with what is in this table and iterate on it.
```
TallTed, line 111:
```suggestion
* eP: The table has a column, "operation", that is not part of any existing specs — that is something where ?? and access are not defined. They need to be defined and added to spec.
* MDJ: Yes, but that doesn't contradict what I said. We just need to work with what is in this table and iterate on it.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+
+### DST limbo warning
+
+* eP: some timezones will shift at the end of the month, CG meeting stays pinned to UTC
```
TallTed, line 115: It would be good to link to a calendar... or a worldclock... or something.
```suggestion
* eP: Some timezones will shift at the end of the month. CG meeting stays pinned to UTC.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Special Topic Meetings
+
+
+#### Summary of CRDT meeting
+
+URL: https://wickie.invisible.college/solid-crdt
+
+* eP: I proposed creation of **CRDT for RDF CG** please support - https://www.w3.org/community/groups/proposed/#crdt4rdf
+* ep: Have discussed, Michiel, myself, Tim, Rahul, Niko from Nextgraph, Michael from Braid were present. Discussed CRDTS specifically w. RDF, and then other things (e.g. text, JSON, etc.). For this community have followed up with people already experience with RDF CRDTS, need to follow up with people with experience with other CRDTS. Also need to prioritise which formats we want to work with beyond RDF. There will be more expertise from other efforts as well, as lots of communities with interest.
```
TallTed, line 125:
```suggestion
* eP: I proposed creation of **CRDT for RDF CG** please support — https://www.w3.org/community/groups/proposed/#crdt4rdf
* ep: Have discussed. Michiel, myself, Tim, Rahul, Niko from Nextgraph, Michael from Braid were present. Discussed CRDTS specifically toward RDF, and then other things (e.g., text, JSON, etc.). For this community, have followed up with people already experience with RDF CRDTs; need to follow up with people with experience with other CRDTs. Also need to prioritise which formats we want to work with beyond RDF. There will be more expertise from other efforts as well, as lots of communities have interest.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+
+URL: https://www.w3.org/events/meetings/b277ff65-0aad-425e-bd1d-64758cd4547a/20241022T140000/
+* MDJ: Was supposed to be yesterday, moved to Nov. 5
+
+### Vocab PRs
+
+#### Add solid:proxy to Solid Terms
+
+URL: https://github.com/solid/vocab/pull/94
+* eP: My suggestion for this and the data wallet PR, we shouldn't add tests just because someone has idea for new terms. We should only add terms that are in drafts of the spec. We had some conversations with Sarven about where the requirements should go. He put them in a comment in vocab, I don't see this as appropriate place. We need to follow up over there to move specific requirements over to the approproate draft. If normative should go in a draft not vocab comment.
```
TallTed, line 138:
```suggestion
* eP: My suggestion for this and the data wallet PR — we shouldn't add tests just because someone has anidea for new terms. We should only add terms that are in drafts of the spec. We had some conversations with Sarven about where the requirements should go. He put them in a comment in vocab; I don't see this as the appropriate place. We need to follow up over there to move specific requirements over to the appropriate draft. If normative, should go in a draft, not a vocab comment.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Add solid:Wallet for a Data Wallet
+
+URL: https://github.com/solid/vocab/pull/93
+* eP: This should be used in one of the drafts, and then referenced. All reqiurements will be in relevant draft.
+* TBL: Principle of what you should put on the ontology itself. Push back on ontology should be in draft, when the ontology should be the raw definition of the term. If you use range, domain, foaf:account etc. then you know what those terms looks like. So where some terms of the wallet are well-defined and you can put in things. The idea of outting the normative information in the comment, if it helps explain that is fine - you can always kick through to the draft. If you find yourself writing comments in the turtle file, or the RDF, don't waste your expressive power - instead put them into the ontology e.g. rdfs:comment. Good ontologies have comments about themselves. Put stuff in the terms where it will be useful to someone who comes accross on a random piece of information on the net, which some system is looking at in the future, where they can click on the predicate and see what it means.
+* eP: We can follow up in PR.
```
TallTed, line 145:
```suggestion
* eP: This should be used in one of the drafts, and then referenced. All requirements will be in relevant draft.
* TBL: Principle of what you should put on the ontology itself. Push back on ontology should be in draft, when the ontology should be the raw definition of the term. If you use `range`, `domain`, `foaf:account`, etc., then you know what those terms look like. So where some terms of the wallet are well-defined, you can put in things. The idea of putting the normative information in the comment — if it helps explain, that is fine — you can always kick through to the draft. If you find yourself writing comments in the turtle file, or the RDF, don't waste your expressive power — instead put them into the ontology, e.g., `rdfs:comment`. Good ontologies have comments about themselves. Put stuff in the terms where it will be useful to someone who comes across a random piece of information on the net, which some system is looking at in the future, where they can click on the predicate and see what it means.
* eP: We can follow up in PR.
```

meetings/2024-10-23.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+#### http://www.w3.org/ns/solid/terms#updatesViaStreamingHttp2023
+
+URL: https://communitysolidserver.github.io/CommunitySolidServer/latest/usage/notifications/#streaming-http
+
+* eP: Possibly should have created it in `/ns/solid/notifications` instead of `/ns/solid/terms`. We can eaither leave it under `solid:` or I would PR it as a breaking change for CSS v8.0.0
+* eP: I'm going to make PRs to the StreamingHTTPChannel2023 and solid/vocab once we decide on the final namespace
+* eP: I don't have PR yet, but in my implementation for CSS I have a new IRI for ??? which links to ???. One doesn't have to link to notif. channel since someone can stream straight from HTTP. I used this relation but maybe should have been defined in solid notifs. (`/ns/solid/notifications`) rather than solid terms (`/ns/solid/terms`). There will be CSS v8 which could incorporate that breaking change.
```
TallTed, line 154:
```suggestion
* eP: Possibly should have created it in `/ns/solid/notifications` instead of `/ns/solid/terms`. We can leave it under `solid:`, or I could PR it as a breaking change for CSS v8.0.0.
* eP: I'm going to make PRs to the `StreamingHTTPChannel2023` and `solid/vocab` once we decide on the final namespace
* eP: I don't have a PR yet, but in my implementation for CSS, I have a new IRI for ??? which links to ???. One doesn't have to link to notification channel since one can stream straight from HTTP. I used this relation, but maybe should have been defined in Solid notifications (`/ns/solid/notifications`) rather than Solid terms.(`/ns/solid/terms`). There will be CSS v8 which could incorporate that breaking change. 
```

TallTed (changes requested):

PR #701, opened 2024-12-11
Create 2024-12-11.md

michielbdejong:


meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+URL: https://matrix.to/#/!QxZtVBYQfMeMTnespj:gitter.im/$GTTvQgauz8PY6ky2bRSMwyFVxPceA47C52asMPYtDlY?via=gitter.im&via=matrix.org&via=chat.semantic.works
+
+- MdJ: the discussion about C2C: is it Data Formats or something else?
+  - ... it is also how you link things together
+- eP: also discovery (what about path, or just "follow your nose")
+- eP: also authorization is involved. if app not authorized, should not see any data. Even if we try to decouple, there is still dependency to auth/permissions.
+  - ... the main idea is how apps can operate on common data.
+  - ... meaningful interactions, which goes beyond just data formats
+- MdJ: client client specs? I like the name for now.
+  - ... there is no direct communication between clients. there is no realtime communicatino. no P2P. Do we have a better name?
```
TallTed, line 85:
```suggestion
- eP: also authorization is involved. if app not authorized, should not see any data. Even if we try to decouple, there is still dependency on auth/permissions.
  - ... the main idea is how apps can operate on common data.
  - ... meaningful interactions, which goes beyond just data formats
- MdJ: client client specs? I like the name for now.
  - ... there is no direct communication between clients. there is no realtime communication. no P2P. Do we have a better name?
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+- eP: "app interp" suggested as name
+  - ... freedom of choice with diversity of apps, still we guarantee interop of data.
+- MdJ: we can also see the matter as : apps talking to other apps.
+- eP: hearing "client to client" could be confused with P2P. it also confusing as client is used in "client-server" but here we not necessarily talk about a client as opposed to server
+- MdJ: the apps are abiding by protocols or specs. users don't , they just "click away". so more in favor of using the "client" term. but client can be confusing, I agree. can sound to much like WebRTC. But client-server-client is also fine.
```
TallTed, line 90:
```suggestion
- MdJ: We can also see the matter as apps talking to other apps.
- eP: Hearing "client to client" could be confused with P2P. It is also confusing as client is used in "client-server", but here we are not necessarily talking about a client as opposed to server
- MdJ: The apps are abiding by protocols or specs. Users don't, they just "click away". So more in favor of using the "client" term. But client can be confusing, I agree. Can sound too much like WebRTC. But client-server-client is also fine.
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+- eP: is it a star topoly of many clients to one server, or many clients to many servers. There are more clients and more servers in the topology. It could be client2server2client2server2server2client.... etc
+- MdJ: the main question: how the data moves
+  - ... who is sending data, who is receiving data
+  - ... it doesn't matter the exact route the data takes
+  - ... let's decide on "ClientClient" spec?
+- eP: it can be confusing, very similar C2C or CC specs.
```
TallTed, line 96:
```suggestion
- eP: Is it a star topology of many clients to one server, or many clients to many servers? There are more clients and more servers in the topology. It could be client2server2client2server2server2client..., etc.
- MdJ: The main question: how the data moves.
  - ... Who is sending data, who is receiving data?
  - ... It doesn't matter the exact route the data takes.
  - ... Let's decide on "ClientClient" spec?
- eP: It can be confusing, very similar C2C or CC specs.
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+- eP: "apps and pods"?
+- MdJ: what about SAI ? (joke)
+    - ... by example the [chat spec](https://github.com/solid/chat) is a good example of "client client spec".
+- eP. and we havent said anything about "domain" yet
+  - ... knowledge domain? domain interop? 
+- H: "data domain specs"
+- eP: data domain specs. but no concsensus or contract as it blends with other concepts
+  - ... Thanks, Niko, i'll take it from here.
+- eP: Data Domain Specs seems neutral, probably the best choice
+- MdJ: Maybe we should talk to Tim
```
TallTed, line 107:
```suggestion
- MdJ: What about SAI? (joke)
    - ... By example, the [chat spec](https://github.com/solid/chat) is a good example of "client client spec".
- eP. And we havent said anything about "domain" yet.
  - ... Knowledge domain? Domain interop?
- H: "Data domain specs"?
- eP: Data domain specs. But no consensus or contract as it blends with other concepts.
  - ... Thanks, Niko, I'll take it from here.
- eP: Data Domain Specs seems neutral, probably the best choice.
- MdJ: Maybe we should talk to Tim.
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Trusted applications
+* MdJ: The 'trusted applications' mechanism we implemented in [NSS v5](https://github.com/nodeSolidServer/node-solid-server/blob/v5.0.0/CHANGELOG.md#500) was always meant to be a stop-gap. As solidcommunity.net is switching from NSS to Pivot, we need something to replace it. We are not yet at the point where working implementations of SAI exist. Therefore, I'm thinking of adding an app launcher along the lines of https://solid.pondersource.com/, that [edits the WAC ACLs and type index files](https://github.com/inrupt/launcher-exploration/blob/master/src/services/preparePod/preparePodForApp.ts) to allow apps as you launch them.
+  - ... we don't have SAI working in CSS.
```
TallTed, line 111:
```suggestion
* MdJ: The 'trusted applications' mechanism we implemented in [NSS v5](https://github.com/nodeSolidServer/node-solid-server/blob/v5.0.0/CHANGELOG.md#500) was always meant to be a stop-gap. As solidcommunity.net is switching from NSS to Pivot, we need something to replace it. We are not yet at the point where working implementations of SAI exist. Therefore, I'm thinking of adding an app launcher along the lines of <https://solid.pondersource.com/>, that [edits the WAC ACLs and type index files](https://github.com/inrupt/launcher-exploration/blob/master/src/services/preparePod/preparePodForApp.ts) to allow apps as you launch them.
  - ... We don't have SAI working in CSS.
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+* ep: How do you identify apps in WAC ACLs?
+* MdJ: Web apps use the WebID of the logged in user and the client id is the app URL.
+  - ... the WAC spec says that that should work
+  - ... I don't know how it's implemented in specific servers.
+  - ... The special thing use is that the client ID matches the domain of the redirect URL from OAuth.
```
TallTed, line 116:
```suggestion
* MdJ: Web apps use the WebID of the logged in user, and the client id is the app URL.
  - ... The WAC spec says that that should work.
  - ... I don't know how it's implemented in specific servers.
  - ... The special thing is that the client ID matches the domain of the redirect URL from OAuth.
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+* eP: If that that uses ACL orgin but not as the orgin header, I think it's not currently defined that way in WAC.
+  - ... In Solid OIDC we have to 2 options currently. [...]
+  - ... There is no global ID identifier such as WebID. 
+  - ... We can pick it up at one of the next meetings.
```
TallTed, line 120:
```suggestion
* eP: If that uses ACL origin but not as the origin header, I think it's not currently defined that way in WAC.
  - ... In Solid OIDC, we have 2 options currently. [...]
  - ... There is no global identifier such as WebID. 
  - ... We can pick it up at one of the next meetings.
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+* MdJ: we should check the WAC spec to make sure it's clear. I don't think we should move away from dynamic registrations. That's because you can trust a random app, even if the server never saw that app before.
+* eP: if you use URLs as client IDs, you don't need to do dynamic registrations. https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-metadata-document/
+* MdJ: so it's dynamic, although it doesn't use the OIDC dynamic registration.
+  - ... I opened a PR on pivot to change a dialog, related to this conversation here.
+  - ... [from chat]: https://github.com/solid-contrib/pivot/pull/38
+* eP: I don't think r/w that data, but all the data you have access to. If you're in a social graph, interacting with other users, that's a problem.
```
TallTed, line 126: What does `r/w` mean, here? (See the last line.) The usual `read/write` doesn't make sense in this context.
```suggestion
* MdJ: We should check the WAC spec to make sure it's clear. I don't think we should move away from dynamic registrations, because you might trust a random app, even if the server never saw that app before.
* eP: If you use URLs as client IDs, you don't need to do dynamic registrations. https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-metadata-document/
* MdJ: So it's dynamic, although it doesn't use the OIDC dynamic registration.
  - ... I opened a PR on Pivot to change a dialog, related to this conversation here.
  - ... [from chat]: https://github.com/solid-contrib/pivot/pull/38
* eP: I don't think r/w that data, but all the data you have access to. If you're in a social graph, interacting with other users, that's a problem.
```

meetings/2024-12-11.md
```diff
@@ -0,0 +1,134 @@
...
+* Niko: yes, that's the main reason I came today, this subject is very interesting to me. I am local first oriented. URLs are tied to servers, but local first does not involve a server [...]
+  - ... I am very interested in identifying applications. I am thinking of taking a hash on the binary. 
+* MdJ: There is also the concept of an instance of an app running on a specific device.
+  - ... Because there is no point of entry, as there is in a normal web server, probably the permissions should travel with the data. I am giving access to this data, and here's how to request it in the future.
+* eP: E2EE comes into play too. Should we put this on next week's agenda? How do we follow up?
+* MdJ: I'm gonna do some implementing and some documentation.
+* ...: There's an ongoing topic, how do we trust apps in Solid.
```
TallTed, line 133:
```suggestion
* Niko: Yes, that's the main reason I came today, this subject is very interesting to me. I am local-first oriented. URLs are tied to servers, but local-first does not involve a server [...]
  - ... I am very interested in identifying applications. I am thinking of taking a hash on the binary. 
* MdJ: There is also the concept of an instance of an app running on a specific device.
  - ... Because there is no point of entry, as there is in a normal web server, the permissions should probably travel with the data. I am giving access to this data, and here's how to request it in the future.
* eP: E2EE comes into play, too. Should we put this on next week's agenda? How do we follow up?
* MdJ: I'm gonna do some implementing and some documentation.
* ...: There's an ongoing topic, how do we trust apps in Solid.
```

TallTed (changes requested):

Issue #704, opened 2025-01-07
N3 Patch contradiction about blank nodes: can they be part of `?insertions`?


lecoqlibre:
In the current section [5.3.1 Modifying Resources Using N3 Patches](https://solidproject.org/TR/protocol#n3-patch) of the Solid protocol (v0.11.0 - 12 May 2024), we can both read that:

`The ?insertions and ?deletions formulae MUST NOT contain blank nodes.` ([source](https://solidproject.org/TR/protocol#server-patch-n3-blank-nodes))

AND:

`The triples resulting from ?insertions are to be added to the RDF dataset, with each blank node from ?insertions resulting in a newly created blank node.`

The latter suggests that `?insertions` can contain blank nodes while the former explicitly forbids.

bourgeoa:
@csarven @CxRes @timbl 

This RdfLib valid test should be allowed in spec and then in CSS

N3PATCH https://github.com/linkeddata/rdflib.js/blob/82f693dc10d5b915f97eedeab5351cf831c3611e/tests/unit/update-manager-n3patch-insert-test.ts#L84-L99

This is a valid  `application/sparql-update` situation including in CSS

csarven:
There is a proposal to resolve this issue in https://github.com/solid/specification/pull/711 .

michielbdejong:
Here is another proposal to resolve this, but in the opposite direction: #716

PR #710, opened 2025-02-05
Agenda for 2025-02-05 CG meeting

hzbarcea:

PR #652, opened 2024-05-15
Add TR/2024/wac-20240512

csarven:
Add Web Access Control, Version 1.0. Publishes:

* This version: https://solidproject.org/TR/2024/wac-20240512 (add)
* Latest published version: https://solidproject.org/TR/wac (update)
* TimeMap: https://solidproject.org/TR/wac.timemap (update)

"This version" is a request to be published as a CG-DRAFT report of the Solid CG.

---

The `#changelog` section indicates changes since [TR/2022/wac-20220705](https://solidproject.org/TR/2022/wac-20220705).

[Preview](http://htmlpreview.github.io/?https://github.com/solid/specification/blob/6dfc5b77c6d6f7b37d8bc98fcd7f49c917c658ef/wac.html) | [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2Fcc42a36928a7961ecc34923fe4b375f5a95f6389/wac.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fspecification%2F6dfc5b77c6d6f7b37d8bc98fcd7f49c917c658ef/wac.html)

Issue #480, opened 2022-11-07
Consider further detailing the classes of products for the conformance model


csarven:
Solid Protocol's #conformance model identifies servers and clients as the kinds of products that will implement the specification ( as clarified by https://github.com/solid/specification/pull/478 alongside https://github.com/solid/vocab/pull/84 ).

Servers and clients is just one generalisation. The Solid Protocol normatively requires other specifications with their own conformance classes of products.

Identifying other meaningful generalisations of the classes of products opens up a new space in that different conforming products can be implemented by different parties and codebases, modularised and used in the Solid ecosystem, exchanged in a marketplace, or searched in stores.

Another generalisation of the conformance classes in the Solid Protocol could also be the following for example:
* server (or responding agent)
* client (or producer or consumer)
* producer of content
* consumer
* producer
* storage
* processor

Besides the definitions and classifications, conformance clauses need to be updated to refer to specific concepts, e.g., is it the server or the storage product that is the subject of that requirement that needs to conform?

Meaningful conformance classes of products and cross linking is expected in all Solid specifications.

PR #711, opened 2025-02-05
change: allow blank nodes to be inserted in N3 Patches

jeswr:
In todays CG and SolidOS calls we discussed this change, to allow blank node insertions in N3 patches.

In both cases, all present members saw no reason why blank node insertions should be forbidden. @bourgeoa indicated that it was likely a mistake, rather than an intentional decision to forbid blank node insertions in the spec.

If you have any objections to this change, please raise them in this PR. If no objections are raised, this PR will be merged 10 days from now.

Context:
 - https://github.com/linkeddata/rdflib.js/issues/659#issuecomment-2624237341
 - https://github.com/solid/specification/issues/704

cc @csarven @bourgeoa @CxRes @timbl @lecoqlibre


csarven (commented):
Some comments and a question before I can proceed with a technical review:

In https://lists.w3.org/Archives/Public/public-solid/2025Jan/0028.html , I asked:

>Has the Solid CG recently decided to continue working on (some or all)  items that are also part of the LWS WG charter deliverables?

The position of the Solid CG was stated in https://lists.w3.org/Archives/Public/public-solid/2025Jan/0044.html :

>The general understanding is that only the core Solid Protocol report is currently an input of the WG.

My understanding of this is that the Solid Protocol is not something the Solid CG will continue to work on. However, this PR, along with the decision made during the Solid CG meeting on 2025-02-05 ([minutes under review](https://github.com/solid/specification/pull/710/files#diff-657fdb8484f4c1f5e66eac906ae818492a83e5695849c68f96eb5b9e4d08389cR115)), appears to conflict with that understanding as per the Solid CG charter's [Out of Scope](https://www.w3.org/community/solid/charter/#out-of-scope) section:

>Community Group’s Work Items that have transitioned to a deliverable of an active Working Group.

This is further complicated by an unnecessary/redundant "decision" - since the charter had already ruled it out of scope - made in the [2024-09-11 meeting](https://github.com/solid/specification/blob/main/meetings/2024-09-11.md#lws-wg), literally:

>PROPOSAL: CG stops working on the Solid Protocol draft
>* MdJ: +1
>* eP: +1
>* PAC: +1
>
>RESOLUTION: CG stops working on the Solid Protocol draft

This sends mixed signals. I ask once again: **Is the Solid CG working on reports like the Solid Protocol or not?**

* If yes, the CG's "final" decision should also be reflected in the CG charter and communicated to LWS WG. I'm happy to review if that understanding is made clear, with prior consent to review and edit.
* If not, then this PR can remain open but unmerged (while still collecting data, additional implementation reports, reviews etc.)

Please also note that as per [Solid CG Contribution Guidelines](https://github.com/w3c-cg/solid/blob/main/CONTRIBUTING.md#processing-pull-requests), I believe this PR qualifies as a [Class 3](https://www.w3.org/Consortium/Process/#class-3) change. If you believe it falls under a different category, please indicate that in the PR comments. Assuming it is a Class 3 change, it would be best to follow the guidelines to processing PRs - in this case, allowing "within 10 days or 2 meetings" to provide sufficient time for review.

jeswr:
> "within 10 days or 2 meetings" to provide sufficient time for review.

Good shout - I've updated the timeline in the original message.

> Is the Solid CG working on reports like the Solid Protocol or not?

My interpretation (including the 2024-09-11 resolution) is that the CG should focus energy on deliverables outside of LWS scope - however, this does not mean that the CG spec is locked.

This change would fix an active blocker in the solidcommunity.net migration of NSS -> CSS, and land before the LWS starts to actively use the input documents such as the CG spec; so pragmatically I think it is a good idea for the change to be made now if there are no objections on technical grounds.

elf-pavlik:
I would propose to keep the issue focused on technical conversation and move all the process-related chat to mailing list thread https://lists.w3.org/Archives/Public/public-solid/2025Feb/0019.html

csarven:
Jesse, I find your interpretation of the charter's out of scope section and the resolution "CG stops working on the Solid Protocol draft" as meaning the spec isn't locked rather unusual. If the intent was simply to indicate that the spec isn't locked so to speak, there could've been many other, clearer ways to express that.

Who exactly shouldn't be actively working on it? https://github.com/solid/specification/commits/main/ED/protocol.html

And if that's the case, isn't it strange that only three people in the entire CG voted "CG stops working on the Solid Protocol draft" on the spot, without consulting the actual primary contributors to the specification? I'm not asking you to justify it. If the intent wasn't to lock, what did the resolution even achieve given CG charter's out of scope?

This seems part of a broader pattern, particularly with the reluctance to move things forward in the Solid CG before the LWS WG charter was approved. For instance, see the concerns as mentioned here:
https://github.com/solid/specification/issues/227#issuecomment-1993840255

---

Pavlik, there's no reason why process-related comments can't be made here especially in this case when I'm raising concerns about mixed signals regarding the process for handling updates to certain specs. And when merge dates are being mentioned, it's better to err on the side of caution.

In any case, I've proposed a simple approach for mutual understanding between the LWS WG and Solid CG here: https://lists.w3.org/Archives/Public/public-lws-wg/2025Feb/0000.html

That'd be all I have to say on process here and happy to follow up once the mutual understanding is accepted/rejected.

csarven (approved):
Leaning on SPARQL Update and SPARQL Query for some guidance pertaining to INSERT and blank nodes, SPARQL Update doesn't prohibit INSERT's with blank nodes, but there are some expectations on how they are applied, e.g.:

https://www.w3.org/TR/2013/REC-sparql11-update-20130321/#insertData states:

>Blank nodes in QuadDatas are assumed to be disjoint from the blank nodes in the Graph Store, i.e., will be inserted with "fresh" blank nodes.

Only blank nodes seem to be disallowed in DELETE, https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#sparqlGrammar :

>8. Blank node syntax is not allowed in DELETE WHERE, the DeleteClause for DELETE, nor in DELETE DATA.
>9. Rules for limiting the use of blank node labels are given in section 19.6.

https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#grammarBNodes :

>The same blank node label can not be used in:
>two INSERT DATA operations within a single SPARQL Update request

https://www.w3.org/TR/2013/REC-sparql11-update-20130321/#deleteInsert :

>Blank nodes that appear in an INSERT clause operate similarly to blank nodes in the template of a CONSTRUCT query, i.e., they are re-instantiated for any solution of the WHERE clause;

So, as far as I can tell (and [my hunch](https://matrix.to/#/#solid_specification:gitter.im/$CF-iWpmkEjSOsJolfalbRYtWWOaA3a5xQx_0u4ss5vE)), removing `?insertions` would at least better align with SPARQL Update.

---

As I see it, this is a [class 3](https://www.w3.org/policies/process/#class-3) change:

>makes conforming data, processors, or other conforming agents become non-conforming according to the new version
>makes non-conforming data, processors, or other agents become conforming

So, it is not a conformance issue for conforming client implementations that were not using blank nodes in INSERT. They can continue to work as they currently do but now have the option to include blank nodes in their INSERT queries.

As for previously non-conforming servers, i.e., the ones that were processing INSERT queries with blank nodes, with this change, they become conforming.

As for previously conforming servers, i.e., the ones that were not processing INSERT queries with blank nodes, with this change, they would become non-conforming. They have two general options: either implement to allow blank nodes in INSERT (successful response) or continue to run with reduced capability but make sure to reject payloads including INSERT and blank nodes with 405 or 415.

---

It is useful to be able to insert statements with blank nodes. Approving the PR based on the above. If there is information that's being overlooked as to why this change shouldn't be made, it'd be good to review that as well.

---

If not already, perhaps the N3 Patch tests can take the following into account from SPARQL Update's INSERT DATA ( https://www.w3.org/TR/2013/REC-sparql11-update-20130321/#insertData ).

>Blank nodes in QuadDatas are assumed to be disjoint from the blank nodes in the Graph Store, i.e., will be inserted with "fresh" blank nodes.

melvincarvalho:
During the SolidOS call, @bourgeoa mentioned that this might be seen as a bug fix rather than a spec change. Considering that this repo has 180 open issues, it's likely there are still edge-case bugs that will only surface during migrations and broader usage. I propose we allow some flexibility to address these issues on a case-by-case basis as we move forward.

RubenVerborgh (changes requested):
Removing this statement introduces ambiguity and incompatibility, as blank node scoping is not defined in N3, and existing implementations are known to differ. Additionally, it creates a multiplicity problem.

The blank node restriction was deliberately added tot the document after earlier experimentation was found to trigger such problems in existing implementations. Instead, patches should use variables, with the spec-mandated restriction that those should also occur in the conditions block, such that the spec-mandated procedure for dealing with multiplicities can be used.

bourgeoa:
> Instead, patches should use variables, with the spec-mandated restriction that those should also occur in the conditions block, such that the spec-mandated procedure for dealing with multiplicities can be used

That is already the case in RDFlib.js https://github.com/CommunitySolidServer/CommunitySolidServer/issues/1998#issuecomment-2665944336

lecoqlibre:
> Removing this statement introduces ambiguity and incompatibility, as blank node scoping is not defined in N3, and existing implementations are known to differ. Additionally, it creates a multiplicity problem.

@RubenVerborgh The proposal here still forbids to use blank nodes in the `?deletions` (`solid:deletes` ). It just allows to have them in the `?insertions` (`solid:inserts`):
- Current Solid spec: _The ?insertions and ?deletions formulae MUST NOT contain blank nodes_.
- Proposal:  _The ?deletions formulae MUST NOT contain blank nodes_.

Note that the current Solid spec also states: _The triples resulting from ?insertions are to be added to the RDF dataset, with each blank node from ?insertions resulting in a newly created blank node_. This suggests that blank nodes are allowed in `?insertions`, right? If we do not want to allow blank nodes in `?insertions` I think the second part of this statement should also be removed.

> The blank node restriction was deliberately added to the document after earlier experimentation was found to trigger such problems in existing implementations. Instead, patches should use variables, with the spec-mandated restriction that those should also occur in the conditions block, such that the spec-mandated procedure for dealing with multiplicities can be used.

I can understand that variables allow to clearly identify a blank node when deleting. But when inserting, the blank node does not exist yet in the patched document and thus it can't be identified in the `?conditions`.

RubenVerborgh:
@lecoqlibre There are still issues with blank node interpretation in N3 inside and outside of formulas.
We’d have to fix N3 before we can resolve this.

In any case, the current PR is missing a statement about uniqueness of any blank nodes that would occur in the insertions graph. And even if such a statement were to be added, the lack of consistent blank node semantics in N3 would prevent verification or enforcement thereof, so it would be void.

There are various cans of worms here; it's not an easy fix. N3Patch was always a stopgap; we knew this from its inception.

bourgeoa:
@RubenVerborgh Can you explain what is different when using non spec PATCH with SPARQL update which is used in NSS and CSS ? what are the limitations that allow `bnodes` in INSERT ?

NSS made the choice to map `n3-patch` content to a `SPARQL update PATCH` content. With this there is a unique entry point in Rdflib.js.
At the same time the Rdflib.js `updateManager` uses the same algorithm to create the PATCH WHERE clause from `inserts` and `deletes` clauses.
I was with the idea that `n3-patch` at first (before future spec development) would only do what the actual and may be badly named `SPARL update` was doing in actual implementations.

I think that at this point it is this minimal PATCH that we would like to see implemented in all SOLID servers.

lecoqlibre:
@RubenVerborgh I hoped there was an easy fix :)

So in the current Solid protocol we have no way to insert blank nodes with a PATCH request. We are then forced to do a PUT request. If a PATCH can't handle blank nodes it seems to loose its interest as blank nodes are a core feature of RDF. I think I already saw Solid apps that are only using PUT request even to patch documents. This workaround does the job but it's not satisfying and also leads to performance issues (transferring the whole document instead of just what needs to be changed).

> N3Patch was always a stopgap; we knew this from its inception.

If this issue was known from the beginning, did someone already work on trying to resolve the problem?

>  We’d have to fix N3 before we can resolve this.

Would you have a plan? Could you provide more details about "_issues with blank node interpretation in N3 inside and outside of formulas_"? Is "_the lack of consistent blank node semantics in N3_" the only problem?

FYI @bourgeoa, Tim. Berners-Lee indicated that the main problem of SPARQL-update was the impossibility to make a lock ([see this comment](https://github.com/solid/specification/issues/715#issuecomment-2690154960). He also added that N3 provides more features like to have more metadata about who made the change, etc.

RubenVerborgh:
> the actual point raised by @RubenVerborgh is with N3.js

No, it is with the (lack of) a Notation3 specification/standard implemented by all parties.

> Could you provide more details

Existing Notation3 implementations have diverging interpretations of how to handle blank nodes in Notation3 documents, when those blank nodes occur in different scopes/formulas (as would be the case with this proposal). Even worse, any constraint the Solid spec would impose on such blank nodes cannot be implemented consistently _because_ of those diverging interpretations. We'd essentially ask to operate on a subset of N3 that we cannot properly define without changing N3 and updating existing implementations.

The original phrasing sidestepped that problem by sticking to a commonly agreed subset of Notation3 supported by all known parsers, at the cost of limiting functionality (superficially/cosmetically, in the sense that `bnode:` IRIs or similar could be used).

Workarounds are possible; they always are. A proper fix is not within this group's mandate, as it pertains to Notation3.

bourgeoa:
@RubenVerborgh 
> In any case, the current PR is missing a statement about uniqueness of any blank nodes that would occur in the insertions graph. And even if such a statement were to be added, the lack of consistent blank node semantics in N3 would prevent verification or enforcement thereof, so it would be void.

Do you have a concrete example of what you mean by uniqueness.
Do you mean checking that the PATCH relates to only one document ?
or something else ?

lecoqlibre:
@RubenVerborgh I had in mind that we should allow this one:

1. Insert a new blank node:
```turtle
@prefix solid: <http://www.w3.org/ns/solid/terms#>.
@prefix ex: <http://www.example.org/terms#>.

_:insert a solid:InsertDeletePatch;
  solid:inserts { ex:product ex:hasPrice [ ex:value "5.5"; ex:currency "euro". ] . }.
  ```

**Comment by RubenVerborgh** [comment]
@lecoqlibre We totally should—but we can't (easily).

A different syntactical representation of the above Notation3 document is:

```turtle
_:p a solid:InsertDeletePatch;
  solid:inserts { ex:product ex:hasPrice _:v. _:v ex:value "5.5"; ex:currency "euro". }.
 ```
  
 And, according to some (!) parsers, another one is:
  
```turtle
_:p a solid:InsertDeletePatch;
  solid:inserts { ex:product ex:hasPrice _:p. _:p ex:value "5.5"; ex:currency "euro". }.
 ```
 
 And sure, we can start restricting the syntax of N3Patch documents.
 Meaning that Solid implementations also need a custom Notation3 parser specific to N3Patch.

We can do those things.

But none of them are a simple fix.

Better to build a proper patch format to replace N3Patch.

But again, not a simple fix.

**Comment by doerthe** [comment]
Dear all, 

I am very late in reading the  discussion but I am pleased to read that you discuss to use N3 for patches here.  I was co-chairing the N3 community group  and we released a first draft for the semantics where especially the meaning of blank nodes was important (all documents can be found here: https://w3c.github.io/N3/spec/). 
We would like to go for a working group if there is enough interest. Would you be interested in having N3 pushed further. Would you have further use cases?

I would also be open to give more feedback to the patches as they are once I better understand (I am a little bit late in that conversation :) ).

**Comment by doerthe** [comment]
@RubenVerborgh 

> And, according to some (!) parsers, another one is:
> 
> ```turtle
> _:p a solid:InsertDeletePatch;
>   solid:inserts { ex:product ex:hasPrice _:p. _:p ex:value "5.5"; ex:currency "euro". }.
> ```

I do not know any implementation which would rename a blank node `[]` to another existing balnk node `_:p`, where does that come from and what is the reasoning for that? It looks like that is contradicting RDF?

**Comment by doerthe** [comment]
Ah, then I misunderstood your comment here, sorry. Of course blank nodes are always complicated. I thought that it was simply about allowing blank nodes in insert-statements if you are sure that these blank nodes are "fresh".

File: ED/protocol.html [hunk 1]
```diff
@@ -1067,7 +1067,7 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
                       <li id="server-patch-n3-variables" rel="spec:requirement" resource="#server-patch-n3-variables"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain variables that do not occur in the <code>?conditions</code> formula.</span></li>
-                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
```
michielbdejong, line 1070: mind your Latin plurals there ;)

formulae -> formula

michielbdejong (approved):

ED/protocol.html
michielbdejong, line 1070:
```suggestion
                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?deletions</code> formula <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
```

jeswr:
> We would like to go for a working group if there is enough interest. Would you be interested in having N3 pushed further.

Hi @doerthe I think the short answer is - yes, there would be interest in giving N3 WG status. I would endorse / support such as WG; and encourage my institution to vote in favour of a charter.

I am interested is in seeing a stable WG specification so there is no ambiguity on how blank nodes are interpreted. I don't think there is a need from the Solid side for much further evolution of the specs.

ED/protocol.html
TallTed: Is this actually singular `formula` (being only `?deletions`, and not `?insertions and ?deletions`) or might there be multiple `?deletions formulae` (which is how I read it)?

michielbdejong: Ah right, in that case would still be plural, yes. But notice how, one line above it, `<code>?conditions</code> formula` is singular and see https://github.com/solid/specification/pull/711/files/bf731cbd8afb19cf216224a9972e37715edc0fd8#diff-d28e1706bf82e98cbda64de615436823ef8b0b61f5c05b2327b35bb1d0863558R1057-R1061 for context of how the `?inserts`, `?deletes` and `?conditions` formulae are introduced, I do feel it should be changed to singular 'formula' here.

michielbdejong:
In our [5 March 2025 CG call](https://github.com/solid/specification/blob/main/meetings/2025-03-05.md#patch) we discussed that we'll merge this PR and then once that's done, extract N3-PATCH into a separate document.

I think @jeswr said that in a few weeks from now (once he has a bit more time to dedicate to this) he will add a note to this PR about the planned extraction of N3-PATCH into a separate doc, and then merge it, right?

RubenVerborgh:
FYI Some potential issues with the notes:

> NSS always allowed it

Not correct; I wrote the NSS implementation and creating the spec test to reflect what I implemented. The rdflib–NSS connection relied on (a non-standard interpretation of) SPARQL UPDATE.

> If NSS already does it, even if CSS doesn't do it, we can do it in Pivot. 

So the if condition fails there. Pivot (or any other implementation) can't implement “it”, in the sense that “it” is a client–server contract. Pivot could create a custom Notation3 implementation, but clients wouldn't necessarily follow it. That's the real problem, not the implementation.

Pivot could implement SPARQL UPDATE though.

> Why don't we separate it from the protocol spec?

Good idea, but just a reminder that N3Patch was always a stopgap. It was never designed for the long term.

ED/protocol.html
TallTed: But, also in line 1069, `?insertions and deletions formulae` is plural, and I think that's the one that parallels the `formula/e` we're discussing in line 1070, as these are both part of the subject of these sentences.

The one you highlight in line 1069 is part of the object of that sentence. Looking carefully at *that* sentence, I think that the concluding `formula` *should* actually be written as "optionally singular or plural", i.e., `formula(e)` or `formula/e`.

jeswr, line 1070: **For the server-patch-n3-blank-nodes-syntax addition**

I've been through all these comments and other channels discussing this issue. I understand the following syntax restriction, identified as an option in https://github.com/solid/specification/pull/711#issuecomment-2697380995 to:
 - Not require further implementation work from parsers; assuming that server behavior is *undefined* when this condition is violated, rather than requiring servers to error.
 - Not likely to be a condition that existing applications violate. Even if the e.g. N3.js serializer were to be used; I don't expect any client applications re-use the same blank node id for blank nodes within and outside formulae in the quads they pass to the parser.
 
My understanding is that allowing the:
  - Server behavior to be undefined, thus not requiring the parser to identify duplicate blank node ids across scopes, and
   - Assumption that no applications are passing cross-scope ids to serialisers
   
  Mitigates any work for implementors of parsers.

**For the server-patch-n3-blank-nodes-multiplicities addition**

This does create work for implementors to:
1. Check if a blank node already exists in a document
2. If so mint a new identifier for it

This is a server, not a parser, level change - and I understand it to be straightforward to implement. I am willing to directly support implementation if that is the driving concern for proceeding.

```suggestion
                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">The same blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> appear in more than one scope.</span></li>
                      <li id="server-patch-n3-blank-nodes-multiplicity" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-multiplicity"><span property="spec:statement">All blank nodes explicitly added in the <code>?insert</code> statement are to be treated as distinct from existing blank nodes in the document.
```

RubenVerborgh: Security issue: in case of append-only permissions, conditionally minting new blank node identifiers allows, under certain circumstances, guessing whether the document is empty and/or whether certain blank nodes are already present. Minting should likely be unconditional.

The proposed text does not stipulate which party enforces the `MUST NOT` conditions; the undefined behavior suggest in the comment is not reflected in the text.

The overarching issue of whether this is a small change within CG scope still stands.

jeswr: > Security issue

Since the minting occurs on the server side I don't understand how this would reveal any information to the client of whether the document is empty and/or whether certain blank nodes are already present. Could you please explain further?

> The proposed text does not stipulate which party enforces the MUST NOT conditions; the undefined behavior suggest in the comment is not reflected in the text.

Suggested rephrasing. "Server behavior is undefined when blank nodes appear in more than one scope. Therefore clients MUST NOT submit documents which submit the same blank node identifier across multiple scopes."

> The overarching issue of whether this is a small change within CG scope still stands.

This PR was the driver for [this signal](https://lists.w3.org/Archives/Public/public-solid/2025Feb/0040.html) so my understanding is that the WG is on board should the CG wish to consider this in scope.

jeswr: Alternatively for the rephrasing:

"Servers SHOULD error when blank nodes appear in more than one scope. Therefore clients MUST NOT submit documents which submit the same blank node identifier across multiple scopes."

jeswr: Commited in https://github.com/solid/specification/pull/711/commits/70122a6ede9504057dcf863b74cd4625511732e7 to allow in-line commends on the changes

ED/protocol.html
```diff
@@ -1067,7 +1067,9 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
+                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">The same blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> appear in more than one scope.</span></li>
```
csarven, line 1071: This might more directly convey intentional application rather than just occurrence:

```suggestion
                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">A blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> be reused in more than one scope.</span></li>
```

That said, the context of this requirement may be unclear to the reader, more specifically the "scope". It might be more clear if it mentions deletes, inserts, the "formula(e)", "patch resource", the "patch document", or something else.

ED/protocol.html
```diff
@@ -1067,7 +1067,9 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
+                      <li id="server-patch-n3-blank-nodes-multiplicity" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-multiplicity"><span property="spec:statement">All blank nodes explicitly added in the <code>?insert</code> statement are to be treated as distinct from existing blank nodes in the document.
```
csarven, line 1072:
```suggestion
                      <li id="server-patch-n3-blank-nodes-multiplicity" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-multiplicity"><span property="spec:statement">All blank nodes in the <code>?insertions</code> statement are to be treated as distinct from existing blank nodes in the target document.
```

I think you meant `?insertions`.

The "document" at the end can be more clear if what you've intended is "patch document" or "target document" (or target resource). Seems more like the latter.

That said, is this actually repeating https://solidproject.org/ED/protocol#server-patch-n3-semantics :
>The triples resulting from ?insertions are to be added to the RDF dataset, with each blank node from ?insertions resulting in a newly created blank node.

csarven (changes requested):

ED/protocol.html
csarven: >This does create work for implementors to:

Is that really new work? My understanding was that implementations are expected to ignore the bnode identifier, i.e., they materialise a new identifier when inserting. Isn't that essentially expressed in processing https://solidproject.org/ED/protocol#server-patch-n3-semantics :

>The triples resulting from ?insertions are to be added to the RDF dataset, with each blank node from ?insertions resulting in a newly created blank node.

Or are you referring to something else?

---

N3 Patch is defined within the context of the Solid Protocol (with its own data model, processing instructions, behaviours) and is still within the scope of the Solid CG charter. Until N3 Patch is officially adopted or transferred to a more suitable W3C CG or W3C Group, and [generalised](https://github.com/solid/specification/issues/715#issuecomment-2703361188), or there are actual signs of coordinated work (with Solid CG and another group), or there is a better alternative proposed, the current work can continue to be incubated in the Solid CG.

ED/protocol.html
jeswr: Good call - I propose the following clarification 

```suggestion
                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">A blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> be reused in more than one scope - where the four possible scopes are the patch document scope and the insert, delete and condition formulae.</span></li>
```

ED/protocol.html
jeswr: Good point, it is repeating - let's remove it

```suggestion
```

ED/protocol.html
```diff
@@ -1067,7 +1067,8 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
-                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">The same blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> appear in more than one scope.</span></li>
```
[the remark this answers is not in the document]
jeswr: We need to also account for the document-level scope to prevent this case:

```ttl
_:b1 a solid:InsertDeletePatch;
 solid:inserts { _:b1 ex:givenName "Alex". }.
```

```suggestion
                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">The same blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> be used in more than one of the following: insert formula, condition formula, outside formulae.</span></li>
```

ED/protocol.html
jeswr: > Is this really new work

In practice yes, [CSS does have this logic](https://github.com/CommunitySolidServer/CommunitySolidServer/blob/262cdee72c5af492294159e8a4da142cb0793e5e/src/storage/patch/N3Patcher.ts#L85-L87) - since that statement is only relevant if there are blank nodes in the insert statement in the first place. Regardless, it is a 1-liner change.

I'll remove my duplicate text re-stating https://solidproject.org/ED/protocol#server-patch-n3-semantics

ED/protocol.html
jeswr: Closing as this has been picked up in https://github.com/solid/specification/pull/711#discussion_r2019324447

ED/protocol.html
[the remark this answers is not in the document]
jeswr: Alternate option so that it is not implied that servers should be detecting identifier re-use and erroring in that case.

```suggestion
                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">Clients <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> submit patch documents where the same blank node identifier in more than one of the following: insert formula, condition formula, outside formulae. Servers <span rel="spec:requirementLevel" resource="spec:SHOULD">SHOULD</span> treat blank nodes in these different locations as distinct</span></li>
```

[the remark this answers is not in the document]
RubenVerborgh: > no need to mention the "delete" formulae

Multiplicities.

The original N3Patch format was a deliberate simplification, designed to avoid a significant number of complexities and edge cases. There's little text in the original, but this does not generalize to subtle extensions. When avoiding blank nodes, all multiplicity problems are ruled out in step 3 of the algorithm.

And while all is fixable, I hope it becomes increasingly clear that none of what is suggested in this PR constitutes a bug fix or small change. Almost anything more powerful than the N3Patch, adds complexity that the original deliberately avoided (because it was meant to be replaced).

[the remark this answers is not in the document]
jeswr: The reason delete formula do not need to be mentioned because they are already forbidden from containing blank nodes.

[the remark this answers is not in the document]
jeswr: This solves a current problem for implementers using N3Patch.

It is unclear what extensions this is preventing in contrast to not making the changes, nor why those extensions are necessary at present.

I have outlined the solution in the case of the CSS above which is a 2 line change, in the case that it ignores the SHOULD statement. Keeping the implementation effort straightforward and small is the reason I propose making it a SHOULD rather than a MUST.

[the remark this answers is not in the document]
timbl: I want to confirm this fix
•⁠  ⁠is a bug fix,
•⁠  ⁠is not complex - blank nodes are used all over these systems and work well,
•⁠  ⁠is necessary for current applications

This has been requested by numerous community members,  and this needs to be merged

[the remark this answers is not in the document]
RubenVerborgh: It can't be a bug fix if it was designed exactly with this knowledge in mind. And it was.

I agree with necessity. I disagree with complexity; we're opening a can of worms by presuming this is a fix. We're introducing undefined behavior, not just on the level of blank nodes.

But I don't want to be the blocker on this, so I will update my review.

ED/protocol.html
jeswr: Closing as this is discussed in https://github.com/solid/specification/pull/711/files#r1981579330

jeswr: Committing @michielbdejong's `formula` suggestion as this aligns with the reference to conditions `formula` in the above line as the optionally singular formula(e) doesn't appear to be elsewhere in the document.

I suggest proposing a set of batch changes if there is a strong opinion that this should be updated

RubenVerborgh (commented):
(Supersedes my previous review and any of my comments.)

This review is on behalf of an implementation (the Community Solid Server) of this specification.

While I still disagree with the proposed change, I am removing any objections (in as far as they have any weight on the process).

CSS will, reluctantly, implement the updated specification, as it concerns a superset of existing behavior, and thus does not break clients that follow the existing specifications.

However, I would like to add the following comments to the record:

- As the original implementer of N3Patch, and author of the descriptive text, I can assure you that the lack of the described functionality was a deliberate design decision and not an oversight. At the moment of design and implementation, we were aware of this limitation, and the fact that some cases would require the insertion of blank nodes. N3Patch was always intended as a short-term stopgap for a design problem back then, namely that NSS relied on SPARQL UPDATE semantics that contradicted to SPARQL UPDATE 1.1 specification.

- N3Patch was a simplification that avoided a couple of complexities, which allowed us to keep the specification text short and simple. However, re-introducing those complexities, is unlikely to be possible while keeping the simple text.

- The previous point is exemplified by the fact that certain behaviors are now undefined (and not necessarily described as such in the text).

- Whether or not the change is limited to a small number of lines of code is, in this review, not part of the argument.

- The main argument is that:
  - This is _not_ a bugfix in the sense that the original behavior was intentional. Arguments that rely on it being a previously unknown or unintentional bug, do not match the timeline.
  - It not being a bugfix makes it a change, which, in our opinion, puts this work outside of the agreed-upon delineation of CG and WG.
  - For reasons explained above, the change is more substantial than the amount of text indicates, and leads to undefined behavior.
  - N3Patch shouldn't be the way forward in any case; it was a stopgap. A sustainable solution requires a more detailed requirements analysis, which was not performed back in the day. However, adding such a solution to the spec falls outside the CG scope (but it could be interesting for the CG to investigate and advise the WG.)

I still think better ways exist, notably supporting blank node IRIs, which the server could transform to blank nodes when needed.

Separately, we should also critically investigate existing libraries and tools for relying on behavior outside of the specification. The CSS has always been very literal in its interpretation, precisely to help surface such bugs. In that sense, it's a pity that this non-standard usage of the Solid Protocol only started to be discussed when the time pressure of solidcommunity.net was already there, as it forces us to find fast solutions rather than sustainable ones. The fact that this feature first appeared as a CSS pull request, rather than a discussion about the spec (which CSS implements), is also something we should strive to avoid.

TallTed:
Thank you for the [background details](https://github.com/solid/specification/pull/711#pullrequestreview-2729082209), @RubenVerborgh. It might be worth putting [this comment](https://github.com/solid/specification/pull/711#pullrequestreview-2729082209) (or one like it, or links from there to [it here](https://github.com/solid/specification/pull/711#pullrequestreview-2729082209)) in appropriate places on the CSS repo and/or more places (wiki, issue for which a solution will never be implemented) on this repo — something a bit less volatile and other–problem–specific than a comment on a PR. History is important, and we have all been in places where delivering something functional "enough" on a specific date was deemed more important than delivering an indisputably better and more functional thing on an indeterminate future date. Knowing the basis of the decisions that were made then, especially that some other decision was *only* ruled out because of time–to–implement and/or that a chosen implementation was intended to be a stopgap, even though it was implemented 10 (or whatever) years ago now....

ED/protocol.html
```diff
@@ -1067,7 +1067,8 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
+                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?deletions</code> formula <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">Clients <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> submit patch documents where the same blank node identifier in more than one of the following: insert formula, condition formula, outside formulae. Servers <span rel="spec:requirementLevel" resource="spec:SHOULD">SHOULD</span> treat blank nodes in these different locations as distinct if a client submits a non-conformant document.</span></li>
```
csarven, line 1071: Bear with me just a little longer on the editorial.

Background (feel free to ignore this paragraph but it is fundamental to the language used in the spec towards conformance and tests): I suggest not to start mentioning the "client" here, although it is generally, but not strictly, the client that produces a patch document. That's in addition to Solid Protocol not being particularly testable (yet) for clients. Most of the conformance expectation is put on the server. My concern is about the current conformance model and the things that can be tested. The current [classes of products](https://solidproject.org/ED/protocol#classes-of-products) are server and client and what constitutes [interoperable](https://solidproject.org/ED/protocol#interoperability) is their interaction. The actual granularity of the classes of products goes much further but we haven't yet defined them in the specification. See for example https://github.com/solid/specification/issues/480 . I'd love us to tackle that properly - .e.g, the N3 Patch document is more like the product [content](http://www.w3.org/ns/spec#Content) (the N3 Patch definition). For the time being, I suggest to not mess with this too much.

---

If a non-conforming document is applied, I'd argue that it is safer and more reliable to reject the request with `422` (as mentioned in https://solidproject.org/ED/protocol#server-patch-n3-invalid ) rather than processing further. So, the constraint only needs to say what is expected, and anything outside of that should fail, quickly and be loud.

"outside formula" is not clear because it is not particularly defined. With insert and condition formulae, there are at at least some specific references. I think the suggestion we arrived earlier and with a minor adjustment will suffice (to take document-level into account, "patch document", which is already mentioned in the text).

```suggestion
                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">The same blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> be used in the inserts and conditions formulae within the same patch resource, or elsewhere in the patch document.</span></li>                  
```

Separately, if it helps, we can update the [example](https://solidproject.org/ED/protocol#n3-patch-example) to include bnode.

csarven (changes requested):

ED/protocol.html
jeswr: Change looks good. I have committed it.

An updated or additional example with bnodes sounds like a good idea.

TallTed:
> [@csarven] [I suggest](https://github.com/solid/specification/pull/711#discussion_r2022423445) not to start mentioning the "client" here, although it is generally, but not strictly, the client that produces a patch document.

Yes, indeed. Words like "producer", "emitter", "provider", "consumer", "sender", "receiver" (and so many more!) are often better even where there's a clear difference between "client" and "server".

ED/protocol.html
```diff
@@ -1067,7 +1067,8 @@ <h4 property="schema:name">Modifying Resources Using N3 Patches</h4>
...
-                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?insertions</code> and <code>?deletions</code> formulae <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-blank-nodes" rel="spec:requirement" resource="#server-patch-n3-blank-nodes"><span property="spec:statement">The <code>?deletions</code> formula <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> contain blank nodes.</span></li>
+                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">The same blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> be used in the <code>?insertions</code> and <code>?conditions</code> formulae within the same patch resource, or elsewhere in the patch document.</span></li>
```
TallTed, line 1071: A more wholesale rephrasing might produce clearer text. This is an attempt to more completely describe the intended prohibition.
```suggestion
                      <li id="server-patch-n3-blank-nodes-syntax" rel="spec:requirement" resource="#server-patch-n3-blank-nodes-syntax"><span property="spec:statement">The same blank node identifier <span rel="spec:requirementLevel" resource="spec:MUSTNOT">MUST NOT</span> be used in both the <code>?insertions</code> and the <code>?conditions</code> formulae within the same patch resource, nor in either of these and elsewhere in the patch document.</span></li>
```

csarven: I took the liberty to update your suggestion to include "formulae" after "these" to maintain the subject. Adding suggestion.

csarven (approved):
As it stands, this PR has been reasonably reviewed, and feedback has been integrated. I'm not aware of any objections, particularly from [Solid CG participants](https://www.w3.org/groups/cg/solid/participants/)) or implementers. That said, since we've covered quite a bit of (back)ground in this PR, I think it would be worthwhile to signal a "last call" for review where necessary, e.g., Solid CG weekly calls, chats, and the mailing list. It would also be helpful to get additional feedback from client/server implementers on whether they commit to implementing this change or other concrete technical concerns, i.e., new information that's not already covered in this PR.

Please leave a comment in this PR linking to minutes or elsewhere on the decision, and then I can follow up with updating #changelog and other information before merging the PR into ED.

Once again, if and when there is official transfer of this work/findings to another W3C CG or W3C Group, incubation in the Solid CG can continue and reflected in the [ED](https://solidproject.org/ED/protocol).

I will request a snapshot of this PR's URL at https://web.archive.org/ and https://archive.is/ , FWIW.

---

As an implementer of patch document in [dokieli](https://dokie.li/) ([source code](https://git.dokie.li/)), the changes in this PR does not pose a change to the current implementation given that dokieli:

* does not use bnodes in the insertions formula;
* while its `patchResourceGraph` can potentially generate a patch document including the conditions formula, currently no feature uses the conditions formula. Hence, the changes related to bnode do not affect anything related to cases involving the conditions formula.
* dokieli currently only uses bnode with a UUID based identifier for the patch resource subject (but will soon change to use a URI with a fragment instead as there is no need for a bnode).

(Related implementation experience was shared in https://github.com/solid/specification/pull/652#issuecomment-2358473543 )

michielbdejong (changes requested):
Sorry for the confusion about whether this PR is paused or not, and also sorry for changing my stance on this PR twice. I initially [said we should leave this to LWS WG](https://github.com/solid/specification/blob/c3a63aefa6a54edde88f164963296f10e671be33/meetings/2025-03-05.md?plain=1#L67), then [Jesse convinced me](https://github.com/solid/specification/blob/c3a63aefa6a54edde88f164963296f10e671be33/meetings/2025-03-05.md?plain=1#L73), and then it sparked some deep conversation which made us take a step back and reflect on our community's dynamics.

On 24 March I had conversations with both @jeswr (as OP and also as ODI representative) and the CG co-chairs and we decided to take a break from this work. I had communicated this in the Solid OS chatroom where it came up, and I'm linking to that here now: https://matrix.to/#/!wAfwwonbRnRLYejFOR:gitter.im/$tNYNDeAilg5vmOhMwheiCgGoq-Gfn6aoMmMftzPHaC4?via=gitter.im&via=matrix.org&via=chat.semantic.works

So speaking from the CG side, this PR is paused now, until [the ~~21 May~~ **9 April** CG call](https://hackmd.io/c02SjMnwSJCpUjMt-pjCLg?view#Peaceful-Progress)

csarven:
@michielbdejong , can you please link to the CG decision - along with technical information - that justifies your blocking of this PR?

>then it sparked some deep conversation which made us take a step back and reflect on our community's dynamics.

Citation needed. Please also share your reflections.

>On 24 March I had conversations with both @jeswr (as OP and also as ODI representative) and the CG co-chairs and we decided to take a break from this work.

That's not CG consensus. As far as I can tell, Jesse has continued engaging with the PR to improve the text, so they are definitely not "taking a break". When you use "we", please recognise the difference between that "we" and the CG as a whole. That's why I'm asking for a CG decision that actually justifies things. And, no, the chairs doesn't just decide to take a break on things, and if anything, they try to ensure the process and consensus is accurately reflected and documented.

As I've said in my last review, if the CG needs more time to review this PR (to gather new information), it can do so. That however does not justify your non-technical blocking. Once again, please refer to https://github.com/w3c-cg/solid/blob/main/CONTRIBUTING.md#processing-pull-requests on the expected processing time, and that's of course being generous and mindful at this point about the fact that the last change to the text was made today. So, that's "Within 10 days or 2 meetings" if anyone would like to request that. It can be of course be earlier if there is sufficient feedback gathered at the discretion of the group. So, perhaps reach out to implementers and get them put things on record in this PR. As the editor, I believe the current text adequately expresses what's intended. And, TimBL, as co-editor, has also expressed interest in merging. So, we're just being mindful about the review period at this point.

michielbdejong:
Thanks for voicing your concerns about CG process Sarven, you definitely have a point. The reason for the pause was that the community we call "Solid" as a whole (not just the CG) is now failing to discuss a tiny topic like this in a peaceful way. Not all the conversations around it took place in CG calls, and not all of it was minuted, but we'll dedicate some time to it in [today's call](https://hackmd.io/QyCnLyCERKm6J--GZv-3Pw?view#PR-711), and we'll also discuss the bigger picture [next week](https://hackmd.io/c02SjMnwSJCpUjMt-pjCLg?view#Peaceful-Progress).

jeswr:
This issue was discussed in the [todays CG call](https://hackmd.io/QyCnLyCERKm6J--GZv-3Pw?view#PR-711), key points disscussed:
 - 10 days are needed from the last substantive change for further input to be given. This seems to be 8 days from now - and is after the proposed pause until the April 9th CG call, meaning there is time to align before this is merged,
 - this repo is governed by the W3C (CG), not the ODI

michielbdejong (approved):
Thanks for your patience, I'm glad we got a chance to talk about what happened here. Let's [move forward as discussed yesterday](https://github.com/solid/specification/pull/726/files#diff-34120ec9ab3ff0150002a4ea38d39acf30a3670061c13912e37f49b7ef87ba37R81-R84)

jeswr:
Note: I made a process mistake only editors are supposed to merge this, so will not merge PRs in this repo in the future. I have checked with @csarven who has indicated that in this instance the PR can be left merged - and will make the changelog updates (which should have been made in this PR) in a separate commit.

csarven:
Thanks @jeswr . Updated with related in https://github.com/solid/specification/commit/9c0467ebf9314b27da9d51b94e206c37b8ae615e .

Just to be clear, changelog and other related info didn't necessarily need to be part of this PR. That is not a major issue. In a nutshell, the editors ensure group's decision is best represented (which is why I [called for a final decision](https://github.com/solid/specification/pull/711#pullrequestreview-2735459686) to be noted).  That said, since this PR introduced two *[substantive changes](https://www.w3.org/policies/process/#substantive-change)* ([changelog](https://solidproject.org/ED/protocol#changelog)), it is generally best for someone else (i.e., the editors) to merge the PR besides the proposer/authors. It is only in some cases (such as the *[editorial changes](https://www.w3.org/policies/process/#editorial-change)* and the TR/ document) that the editors can merge at their own discretion (even if they've authored the PR) but still with some expectations on what's reasonable to PR: Solid Technical Reports Contributing Guide, [Processing pull requests](https://github.com/w3c-cg/solid/blob/main/CONTRIBUTING.md#processing-pull-requests).

PR #716, opened 2025-03-04
Remove confusing mention of blank nodes in PATCH inserts

### LINKED ISSUE #704: N3 Patch contradiction about blank nodes: can they be part of `?insertions`?

michielbdejong:
This PR is an attempt to resolve #704, but in the opposite way of how #711 tries to fix it.


michielbdejong:
Retracting this after discussion at today's CG call.

PR #717, opened 2025-03-04
Fix incorrect pim:Storage link in section 4.1

lecoqlibre:
https://solidproject.org/TR/protocol#client-rdf-storage was referring to http://www.w3.org/ns/pim/space#storage instead of http://www.w3.org/ns/pim/space#Storage (s -> S).


csarven (commented):
I don't think this change is correct. Lowercase "s" is intended because we want to use the property `http://www.w3.org/ns/pim/space#storage` not the class `http://www.w3.org/ns/pim/space#Storage`. The requirement is saying "look for the property <lowercase s> and the object of that triple is of type <uppercase S>". Does that address your concern?

lecoqlibre:
@csarven Ah yes you're right, sorry, my mistake.

PR #721, opened 2025-03-19
Create 2025-03-19.md

michielbdejong:


meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+## Topics
+
+### Moving AddressBook from vcard to foaf
+
+* MdJ: See https://github.com/solid/contacts/issues/8Md
+* MdJ: ...
+* MdJ: VCard is controlled by IETF. If we change W3C ontology it will result in a fork.
+  ... https://lists.w3.org/Archives/Public/semantic-web/2025Mar/0041.html
+  ... This topic pops up in discussions since 1999
+* JW: I'm looking at VCard as RFC which was published in 2011. We could extend vocabulary which is a common practice. We can have `solid-contacts` namespace and define additional terms there.
```
TallTed, line 47:
```suggestion
* JW: I'm looking at the VCard RFC which was published in 2011. We could extend the vocabulary which is a common practice. We can have `solid-contacts` namespace and define additional terms there.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* MdJ: This is close to Dan's proposal of moving AddressBook to FOAF. Address books are not specific to Solid. FOAF has similar terms to VCard. 
+* JW: FOAF already has alternatives to Vcard...
+  ... I see Tim's disagreement to use FOAF
+* MdJ: Instead of migrating it to FOAF we
```
TallTed, line 51:
```suggestion
* MdJ: Instead of migrating it to FOAF we ???
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* eP: What is a problem with reusing SIOC AddressBook?
+* MdJ: We have code that uses a term. What went wrong is squating vcard ontolgy. The mechanism that address book uses is different from what SIOC is using. We just want to publis ontology which is used by existing implementations.
```
TallTed, line 53:
```suggestion
* MdJ: We have code that uses a term. What went wrong is squatting on the vcard ontology. The mechanism that address book uses is different from what SIOC is using. We just want to publish an ontology which is used by existing implementations.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* ... databases have migrations, we shouldn't instead of just rename it have to use a different mechanism.
```
TallTed, line 54: I can't decipher this sentence/fragment
```suggestion
* ... databases have migrations, we shouldn't instead of just rename it have to use a different mechanism.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* eP: There is no existing stable shape for address books. One implementation can't expect that everyone else will adopt their choice. We need a curation process which we partially discussed in shaperepos meeting. 
+* HZ: Why do we need to migrate? We should allow on the fly conversions in the infrastructure. This will allow using new standards that may be developed in the future.
+* MdJ: What makes it special is that terms used were never published and can't be published for political reason. If we could publish we wouldn't need to migrate. The main problem is the squating.
```
TallTed, line 57:
```suggestion
* MdJ: What makes it special is that terms used were never published and can't be published for political reason. If we could publish, we wouldn't need to migrate. The main problem is the squatting.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* eP:...
+* JW: We have been discussin ongology alignment and using it in Solid for years. We don't have infrastructure for that so we need an alignment to have interop.
```
TallTed, line 59:
```suggestion
* JW: We have been discussing ontology alignment and use in Solid for years. We don't have infrastructure for that, so we need an alignment to have interop.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* ...: As for getting everyone together, we need more lightweight way of finding consensus. If we need a series of calls this will not neccessarly lead to consensus. I'm thinking about more lightweight alternative. I see 2 or 3 options. We can create new ontoloty that extends Vcard. Another is adding it to FOAF and migrating to FOAF. The last one would be creating existing contacts ontology. We should make a straw poll to see if there is interest in using one of those approaches. Hearing Tim's comments I would prefer keeping VCard and extending it.
```
TallTed, line 60:
```suggestion
* ...: As for getting everyone together, we need more lightweight ways of finding consensus. A series of calls will not necessarily lead to consensus. I'm thinking about more lightweight alternatives. I see two or three options. We can create a new ontology that extends Vcard. Another is adding it to FOAF and migrating to FOAF. The last one would be extending existing contacts ontology. We should make a straw poll to see if there is interest in using one of those approaches. Hearing Tim's comments, I would prefer to extend and keep using VCard.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* MdJ: see https://lists.w3.org/Archives/Public/semantic-web/2025Mar/0031.html
+* ...: If we don't change the name and just change the prefix it would be easy to change.
+* JW: My suggestion is not listed in your email. I propose keepting vcard and just creating `vcard-extended` where we can define `AddressBook`.
+* MdJ: Are you saying that this is non-breaking change, so we are going to use `http://www.w3.org/2006/vcard/ns#` namespace?
```
TallTed, line 64:
```suggestion
* ...: If we don't change the name and just change the prefix, it would be easy to change.
* JW: My suggestion is not listed in your email. I propose keeping Vcard and just creating `vcard-extended` where we can define `AddressBook`.
* MdJ: Are you saying that this is non-breaking change, so we are going to use `http://www.w3.org/2006/vcard/ns#` namespace?
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* MdJ: This sounds similar to publising a new note, which seems problemtatic. We will need to use a different namespace.
+* JW: We are not copying any old VCard temrs, we just minting new URI for extended context where we define one new term.
+* MdJ: There actually 6 squatted terms,
```
TallTed, line 67:
```suggestion
* MdJ: This sounds similar to publising a new note, which seems problemtatic. We will need to use a different namespace.
* JW: We are not copying any existing VCard terms; we are just minting a new URI for extended context where we define one new term.
* MdJ: There actually 6 squatted terms:
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+  vcard:nameEmailIndex
+  vcard:inAddressBook
+  vcard:includesGroup
+  vcard:WebID
+  ```  
+* MdJ: Still the namespace would be different than currently used in data.
+* JW: Where is the AddressBook term already used?
+* MdJ: In SolidOS and in Contacts DataModules
+* JW: I'm not sure if this is an issue that we currently squat the namespace.
+* MdJ: Calling it Solid extansion to vcard makes sense to me.
```
TallTed, line 80:
```suggestion
* MdJ: Calling it Solid extension to vcard makes sense to me.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* JW: We shouldn't let this kinds of discussions get in the way of making apps. What we can do, if you want to interoperate with SolidOS you need to use the squatted terms now. Later we can introduce the extension and keep any new terms there.
```
TallTed, line 81:
```suggestion
* JW: We shouldn't let this kind of discussion get in the way of making apps. What we can do — if you want to interoperate with SolidOS, you need to use the squatted terms, for now. Later, we can introduce the extension and keep any new terms there.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* eP: Who thinks that WebID Profiles and Contact cards should be assigned?
+* MdJ: ... Solid Data modules allows using both vcard and foaf terms. For instance `foaf:name` is used by CSSe: 
+  ... https://github.com/SolidOS/css-mashlib/blob/a5b0fa02beb78efae79259672f85f2a0971f2824/templates/pod/wac/profile/card%24.ttl.hbs#L11
+* eP: Is SolidOS storing each vcard in a separate resource?
+* eP: Maybe we shouldn't be telling people to make their apps interoperable with SolidOS?
+* MdJ: If we say don't make your app interoperable with existing solid apps we are not getting anywhere.
+* HZ: I see things opposite to how you phrased it, but I see your point of view as valid. Data is data, the meaning is expressed in vocabularies. We have ways to resolve mismatch in terms used in the client or infrastructure.
```
TallTed, line 88:
```suggestion
* MdJ: If we say don't make your app interoperable with existing solid apps, we are not getting anywhere.
* HZ: I see things opposite to how you phrased it, but I see your point of view as valid. Data is data, the meaning is expressed in vocabularies. We have ways to resolve mismatches in terms used in the client or infrastructure.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* MdJ: It doesn't address current squatting.
+* HZ: I see confision between identity and addresses.
+* TT: By squatting you are building a URI in someone's elses world. In semantic web, you should be able to follow your nose and get useful infromation. Squatting breaks this promise.
```
TallTed, line 91:
```suggestion
* TT: By squatting, you are building a URI in someone else's world. In semantic web, you should be able to follow your nose and get useful information. Squatting breaks this promise.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* HZ: Jesse proposed very good solution. To create namespace for extensions.
+* TT: This term has been put in someone's namespace already. We maybe need to say, "you broke it, you fix it". We don't want to standardize or bless things that do something the wrong way.
+* JW: my proposal is: We define the term Addressbook in a namespace we own, but we also keep using the vcard ontology for all the terms that already exist there. We call it something like the 'solid extension to vcard'
```
TallTed, line 94:
```suggestion
* TT: This term has been put in someone else's namespace already. We maybe need to say, "you broke it, you fix it". We don't want to standardize or bless things that do something the wrong way.
* JW: My proposal is: We define the term Addressbook in a namespace we own, but we also keep using the vcard ontology for all the terms that already exist there. We call it something like the 'solid extension to vcard'.
```

meetings/2025-03-19.md
```diff
@@ -0,0 +1,109 @@
...
+* TT: I'm ok with a straw poll, we may still need to explore it further.
```
TallTed, line 95:
```suggestion
* TT: I'm ok with a straw poll, we may need to explore it further before reaching consensus or taking a vote on it.
```

TallTed (changes requested):
Minor cleanup.

PR #725, opened 2025-04-06
Fix present

CxRes:
Fixes present entries for a lot of participants that I was reasonably certain of. 

@mrkvon @thhck You might want to check and PR yours.

@TallTed Your entries were especially difficult. While I have cleaned it to two types, I am not sure which one you prefer.


thhck:
I would not be in favor of removing the double space at the end of the lines, since this is the way in traditional markdown to indicate newline. 

Maybe github flavored markdown don't needs the double space, but if we want to reuse those markdown files outside of github it would be nice to keep the double spaces

TallTed:
> I would not be in favor of removing the double space at the end of the lines, since this is the way in traditional markdown to indicate newline.

I would suggest changing the invisible double-space-EOL to visible `<br>`EOL.

CxRes:
@csarven Most of the fixes that I have made were done with the help of the Solid Ecosystem Monitor. Already this took quite a lot of time.

CxRes:
@thhck Thanks for pointing out double spaces being removed, my code editor is set up to automatically remove spaces at the end of lines.  But I agree with @TallTed that adding `<br>` is better.

PR #760, opened 2026-01-28
Create 2026-01-28.md

elf-pavlik:
at the end we have test of auto transcript by https://meetily.ai/

PR #761, opened 2026-01-30
Remove automated transcript from meeting minutes

uvdsl:
@elf-pavlik,  I propose to remove the transcript from these meeting minutes (which were added via https://github.com/solid/specification/pull/760). 
I tried to get an understanding of what was discussed in the meeting from these notes but I found them to be not useful.
They do not add value to the minutes. On the contrary, they
- are not formatted according to the meeting minutes template (a formality issue)
- are hard to read for humans when rendered (an accessibility / usability issue)
- do not indicate who is talking (an accountability issue)
- have not been at least proof read (an accuracy / correctness issue)


elf-pavlik (approved):

elf-pavlik:
I don't think just having them there is a problem but I don't want to dedicate additional time to discussing it so let's just remove it.

Amnat-auyana1998 (approved):

PR #771, opened 2026-03-29
Require public WebID profile document

langsamu:
Current section 9.1 [mentions](https://solidproject.org/TR/protocol#webid) WebID as "underpinning component" and "used as the primary identifier" in Solid.

This change adds a **requirement for [WebID Profile Documents](https://www.w3.org/2005/Incubator/webid/spec/identity/#:~:text=WebID%20Profile%20or-,Profile%20Document,-A%20WebID%20Profile) to be dereferencable 'publicly'**.

---

WebID 1.0 (section 5.4 Privacy) [says](https://www.w3.org/2005/Incubator/webid/spec/identity/#privacy:~:text=A%20WebID%20Profile%20is%20a%20public%20document)
> A WebID Profile is a public document [...]

but does so in a non-normative section.

This change aims to improve interoperability by adding a normative requirement which corresponds to both
- the intent of the underlying specification (WebID 1.0),
- overwhelming practice in implementations of the subject specification (Solid Protocol) and an
- implicit requirement in a normative dependency (Solid-OIDC, see also solid/solid-oidc#80).


elf-pavlik:
This relates to our conversation from last week CG call which we will continue next Wednesday. Which class of product this requirement is placed on? Currently there is no requirement for hosting WebID Documents in Solid Storage. Besides minutes from last week CG call there is also this discussion:

* https://github.com/solid/webid-profile/discussions/122

I agree that this should be required somewhere, my main point is that it isn't clerar how to forumate it in terms of classes of products.

elf-pavlik:
I added it to the agenda of our weekly CG call
https://hackmd.io/@solid/cg-weekly/#WebID-or-WebID-Profile

PR #772, opened 2026-03-30
Require `pim:storage` in WebID profile

langsamu:
Current section 9.1 [mentions](https://solidproject.org/TR/protocol#webid) WebID as "underpinning component" and "used as the primary identifier" in Solid.

This change adds a **requirement for [WebID Profiles](https://www.w3.org/2005/Incubator/webid/spec/identity/#:~:text=WebID%20Profile%20or-,Profile%20Document,-A%20WebID%20Profile) to be dereferencable 'publicly'**.

---

The Protocol currently [enables](https://solidproject.org/TR/protocol#client-rdf-storage) discovery of the root container from a Solid resource

However there is no defined mechanism for discovering a Solid storage from a WebID.

This change aims to improve interoperability by adding a normative requirement which corresponds to overwhelming practice in implementations of WebID in the ecosystem.


uvdsl (changes requested):
[[WebID](https://www.w3.org/2005/Incubator/webid/spec/identity/)] has been profiled for the Solid ecosystem in the [[Solid WebID Profile](https://solid.github.io/webid-profile/)]

The suggested addition to the Solid Protocol is already covered by that Solid WebID profile, see https://solid.github.io/webid-profile/#webid-profile-data-model .

Also note, that an agent identified by a WebID does not necessarily have a storage, and is not required to obtain a storage at any point. Whether that is useful depends on the use case: There might be a (software) agent which is identified by a WebID but does not have/own storages itself but writes only to external ones.

In addition, requiring the inclusion of the storage on a WebID top-level breaks existing approaches to hide the storage in [extended profiles](https://solid.github.io/webid-profile/#extending-profile) to only let authorized agents (e.g., clients) discover this protected storage. 

I suggest to point to the [[Solid WebID Profile](https://solid.github.io/webid-profile/)] instead of breaking with the orthogonality of specifications employed by the current Solid Protocol.

PR #776, opened 2026-04-05
feat: initialise Solid26 implementors guide

jeswr:
This is the beginning of the implementors guide discussed in #773 - currently it just fixes the specs and versions to be included.

Additional specs such as #774 (which I acknowledge I still owe a response to) may be added to this guide if developed and CG endorsed in time.

A preview link can be found [here](https://htmlpreview.github.io/?https://github.com/solid/specification/blob/solid26/solid26.html).

**EDIT** since there is a lot of active editing on this PR -- I am marking comments as resolved as I implement changes to ease navigation (cc @csarven @elf-pavlik - I hope this is ok, as far as I understand anyone with read access can still expand and read the content as they desire).

PR #780, opened 2026-04-16
SR-187: WAC & ACP

langsamu:
Adding a section about WAC & ACP based on https://github.com/solid/specification/pull/776#issuecomment-4252700765

[Preview](https://htmlpreview.github.io/?https://github.com/solid/specification/blob/solid26-ac/solid26.html)

### DIFF
--- a/solid26.html
+++ b/solid26.html
@@ -301,6 +301,15 @@ content:counter(appendix, upper-alpha) "." counter(sub-appendix) "\00a0";
                   <p><a class="tocxref" href="#implementation-guidance"><bdi class="secno">3.</bdi> <span>Implementation Guidance</span></a></p>
                   <ol>
                     <li class="tocline"><a class="tocxref" href="#webid"><bdi class="secno">3.1</bdi> <span>WebID</span></a></li>
+                    <li class="tocline">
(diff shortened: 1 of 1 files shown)


csarven (changes requested):
The CG did not call for something like this let alone for solid26.

Based on the data published by the CG: https://github.com/w3c-cg/solid/blob/main/implementations/wac-acp.2026-04-01.csv , I think it would be more sensible / useful to help the outlier implementations align with the rest of the ecosystem. So, perhaps the suggestions in this document should be passed on to implementations that would like to conform to WAC if they are arriving from ACP.

elf-pavlik:
I think this should be a separate document which could be referenced from State of Solid 26 report as work in progress.

uvdsl:
I am not sure whether the target audience of this content is the same or sufficiently overlaps with the target audience of the Solid26 Implementation Guide. My gut feeling is that transformations from ACP to WAC and vice versa are an implementation detail that is not universally applicable across all clients... Clients might choose to handle their implementations differently and distinctly / in separate style, especially given the advisement of Solid26 to consider relying on dedicated authorization applications.

PR #781, opened 2026-04-19
solid26: WebID Guidance

jeswr:
This PR adds the WebID 1.0 and the Solid WebID Profile to Solid26.

Additionally it provides:
 - Guidance on what can be concluded from all input documents to Solid26 as far as WebID's are concerned
 - Clear algorithms on how to construct a Solid WebID profile. I have intentionally been _more_ prescriptive than the existing specification here so that implementors have clear guidance. I would very much appreciate a review from the authors of the Solid WebID profile document, and I am happy to contribute this to the Solid WebID profile spec.
 - A guide for "common clients" on how to authenticate a user, and construct their user profile
 
 A preview link can be found [here](https://htmlpreview.github.io/?https://github.com/solid/specification/blob/feat/solid26-webid/solid26.html).
 
 cc @uvdsl @csarven @jeff-zucker @VirginiaBalseiro @timea-solid


jeff-zucker:
Great!  Thanks.  Minor quible: 404 on the link to "here".

jeswr:
Link fixed @jeff-zucker - thanks for catching!

solid26.html
```diff
@@ -453,143 +466,90 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
 
                   <p>For the purposes of this section, the following terms are used:</p>
                   <dl>
                     <dt><dfn id="dfn-webid">WebID</dfn></dt>
                     <dd>An IRI that denotes an agent [<cite><a class="bibref" href="#ref-webid">WEBID</a></cite>], [<cite><a class="bibref" href="#ref-solid-protocol">SOLID-PROTOCOL</a></cite>].</dd>
+                    <dt><dfn id="dfn-webid-profile-document">WebID Profile Document</dfn></dt>
+                    <dd>The Web resource obtained by dereferencing a <a href="#dfn-webid">WebID</a>. Generally public.</dd>
```
jeff-zucker: My understanding is that if the WebID Doc is not public, it can't be used to authenticate. So we should warn users that, while they can make it private, doing so will make it unusable.

solid26.html
```diff
@@ -453,143 +466,90 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
-                        <h5>Presenting WebIDs in user interfaces</h5>
-                        <p>WebID Profiles may carry human-readable labels (such as <code>foaf:name</code> or <code>rdfs:label</code>) that clients can use for display, but Profiles are not required to include such labels. User interfaces should therefore be designed to display a WebID URI as a fallback identifier and should not assume that a short, human-friendly name is always available for either the agent or any associated storage.</p>
-                      </section>
+                      <p>Per [<cite><a class="bibref" href="#ref-webid-profile">Solid WebID Profile</a></cite> § <a href="https://solid.github.io/webid-profile/#discovery">2 Discovery</a>], the full profile of an agent is the union of statements found in:</p>
+                      <ul>
+                        <li>the <a href="#dfn-webid-profile-document">WebID Profile Document</a>;</li>
+                        <li>any <a href="#dfn-extended-profile-document">Extended Profile Documents</a> linked from it (e.g. via <code>rdfs:seeAlso</code>, <code>foaf:isPrimaryTopicOf</code>, or <code>owl:sameAs</code>);</li>
```
jeff-zucker: [EDIT: TLDR; for this thread - the assembling the profile advice to "Repeat step 2 transitively for Extended Profile Documents linked from the Preferences Document."  should be changed to refer to the WebID Document, not the Preferences Document.  This change supports the ability of users to (at their own discretion) add links to extended documents in extended documents.  It does not specify anything about writing to an extended document, only that if a client encounters a link to extended docB inside extended docA, they should follow the link unless they've already loaded docB.  Not all pods  have Preference files and allowing links to extended docs in it but not in public-facing extended docs makes no sense since the Preference file is more vulnerable than the public-facing files.]

For a normative view, that is probably sufficient, but for a practical guide to implementers, there are several things we should mention in the interest of preventing excessive loading.  1) What happens with seeAlso references inside seeAlso files? Implementers should be warned to prevent multiple loading and loops if, e.g., seeAlsoA both references and is referenced by seeAlso B. Also, to what depth does one descend?  Personally, I think users should be warned against long chains of seeAlsos containing seeAlsos 2) I also think that users should be warned not to depend on information in sameAs links - that a client may want to read the specific profile they requested, not all the user's profiles.  If someone gets to my software-dev profile, they may or may not want to include my fire-dancing profile.

solid26.html
jeswr: Thanks Jeff!

This document already advises / assumes to make it public.

My understanding is there is nothing stopping an authorization server from having a setup where it can make authenticated requests. Hence I am not sure we can accurately make that claim. 

I believe @uvdsl had something implemented along these lines as part of MANDAT.

What could be safe to claim is:
 - It is just not common practise 
 - it work not well for the Social Web suite of use-cases for Solid as most Solid (authorisation) servers will not be configured to make authenticated requests.

jeff-zucker: Ah, I wasn't aware there were ways around it.  In that case I think as-is is fine.

jeswr: https://github.com/solid/specification/pull/781/changes/39dfce9071d163afe55f813edf15a88ad6797302

solid26.html
```diff
@@ -433,15 +466,158 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
... summarises predicate usage from a wide sample of live Solid profiles. The list below aligns with the companion <a href="https://github.com/solid/webid-profile/blob/main/shapes/profile-shapes.ttl"><code>profile-shapes.ttl</code></a> in the <a href="https://github.com/solid/webid-profile/tree/main/shapes"><code>solid/webid-profile</code> shapes</a> &mdash; which targets <code>foaf:Person</code>, <code>vcard:Individual</code>, and <code>schema:Person</code> &mdash; supplemented with widely-used fallbacks populated by existing profile editors (<a href="https://github.com/SolidOS/solidos">SolidOS</a>/<a href="https://github.com/SolidOS/mashlib">mashlib</a>, <a href="https://github.com/inrupt/pod-browser">Inrupt PodBrowser</a>, <a href="https://github.com/pod-os/PodOS">PodOS</a>, <a href="https://gitlab.com/vincenttunru/penny">Penny</a>). Vocabularies referenced: FOAF [<cite><a class="bibref" href="#ref-foaf">FOAF</a></cite>], vCard [<cite><a class="bibref" href="#ref-vcard">VCARD-RDF</a></cite>], Schema.org [<cite><a class="bibref" href="#ref-schema">SCHEMA-ORG</a></cite>], <a href="http://www.w3.org/ns/org#">Org</a>, and <a href="http://www.w3.org/ns/solid/terms#">Solid Terms</a>.</p>
```
csarven: The implementation reports normally mention software, so it may not be particularly meaningful to list them in this kind of a document.

Otherwise, please add <a href="https://dokie.li/">dokieli</a> (<a href="https://git.dokie.li/">source</a>) to the list. Dokieli makes use of a wide range of information about agents, their additional/extended profiles, assets, preferences, policies, as well as their social connections. It also makes use of variety of equivalent properties.

Also agree with earlier comments that referencing abandoned software, like PodBrowser, is neither useful or attractive.

solid26.html
jeswr: > What happens with seeAlso references inside seeAlso files?

My interpretation of https://solid.github.io/webid-profile/#discovery is that there are two locations you look for seeAlso references:
 - The WebID Profile Document
 - The preference document
 
and that one should not recursively look up extended profiles from extended profiles, nor type indexes from typeindexes.

If this is the case, I can clarify the text here - "transitively" - is a bad word for me to be using in this context. If this is not the correct interpretation / not what is done in practise - I am happy to update here. We should probably also clarify that in https://solid.github.io/webid-profile/

> 2) I also think that users should be warned not to depend on information in sameAs links - that a client may want to read the specific profile they requested, not all the user's profiles. If someone gets to my software-dev profile, they may or may not want to include my fire-dancing profile.

Interesting ... I had always thought about profiles a bit like how google chrome operates at the moment. Each profile corresponds to a different WebID (equivalently email when one is using chrome), and that is how you make the distinction.

Are there live examples of these different profiles within one Pod? If so, what are the context hints available; do we need to know the origin document / named graphs to know which statement comes from which profile? Are there other modelling hints?

solid26.html
```diff
@@ -433,15 +466,158 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                <p>Common predicates for rendering a profile (predicates appearing in <code>profile-shapes.ttl</code> are marked &dagger;):</p>
+                <ul>
+                  <li><strong>Name:</strong> <code>vcard:fn</code>&dagger;, with fallbacks to <code>foaf:name</code>, <code>schema:name</code>, or <code>rdfs:label</code>.</li>
+                  <li><strong>Short name / nickname:</strong> <code>foaf:nick</code>&dagger;, with fallback to <code>vcard:nickname</code>.</li>
+                  <li><strong>Avatar / photo:</strong> <code>vcard:hasPhoto</code>&dagger;, with fallbacks to <code>foaf:img</code> or <code>schema:image</code>.</li>
```
csarven: I'll use this as an example but it holds true for this whole list. Stating that vcard:hasPhoto has higher priority than foaf:img is not accurate. It may be the case for ODI or SolidOS (or wherever it is pulled from) but that's not universally agreed upon by all implementations. In fact, clients are more robust when they can take into several of them into account to cater for different circumstances. For instance, dokieli in fact makes use of the following properties in this order:

```
      ns.as.image,
      ns.as.icon,
      ns.foaf.img,
      ns.schema.image,
      ns.vcard.photo,
      ns.vcard.hasPhoto,
      ns.sioc.avatar,
      ns.foaf.depiction
```

I suspect that others implementations may have a different order given their application environment.

So, what I suggest here is a recommendation that makes it clear that there consumers and writers of WebID may encounter or need to create different properties/classes depending on their area of application. This is typically accompanied with Protocols and Data Models that are used alongside Solid Protocol and Solid WebID Profile. And lastly, applications may consume from a range of terms but they may write in a particular one.

It is not so clear cut at as e.g., use hasPhoto if not image.

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                        <li>
+                          The WebID Profile Document is the resource obtained by dereferencing the WebID HTTP URI: for a WebID with a fragment identifier (e.g. <code>#me</code>), the URI without the fragment denotes the Profile Document; for a WebID without a fragment, an HTTP <code>GET</code> on the WebID MUST return a <code>303 See Other</code> response whose <code>Location</code> header URI denotes the Profile Document [<cite><a class="bibref" href="#ref-webid">WebID 1.0</a></cite> § <a href="https://www.w3.org/2005/Incubator/webid/spec/identity/#terminology">2 Terminology</a>; <cite><a class="bibref" href="#ref-webid">WebID 1.0</a></cite> § <a href="https://www.w3.org/2005/Incubator/webid/spec/identity/#the-webid-http-uri">3 The WebID HTTP URI</a>].
+                        </li>
+                        <li>
+                          A WebID Document MAY be hosted anywhere on the Web; it MAY, but need not, reside in a Solid storage [<cite><a class="bibref" href="#ref-webid">WebID 1.0</a></cite> § <a href="https://www.w3.org/2005/Incubator/webid/spec/identity/#publishing-the-webid-profile-document">5 Publishing the WebID Profile Document</a>].
```
csarven: Guide the implementer here. Tell them that a WebID in a Solid storage guarantees certain affordances for reading and writing with the support of Solid Protocol.

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                        </li>
+                        <li>
+                          The WebID Profile Document MUST have a <code>text/turtle</code> representation [<cite><a class="bibref" href="#ref-webid">WebID 1.0</a></cite> § <a href="https://www.w3.org/2005/Incubator/webid/spec/identity/#publishing-the-webid-profile-document">5 Publishing the WebID Profile Document</a>].
```
csarven: This is either redundant or underspecified. It is redundant because it talks about a base WebID Profile Document. It is underspecified because it completely overlooks what Solid WebID Profile says: https://solid.github.io/webid-profile/#reader-application-webid-profile , including JSON-LD. Again, related to the server that's behind it, and that Solid server provides some expectations.

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                        </li>
+                        <li>
+                          The WebID Profile MUST contain one or more statements of the form <code>?webid solid:oidcIssuer ?iss</code>, where <code>?webid</code> is the WebID and each <code>?iss</code> is the IRI of an OpenID Provider trusted to issue ID Tokens for that WebID [<cite><a class="bibref" href="#ref-solid-oidc">Solid-OIDC</a></cite> § <a href="https://solidproject.org/TR/2022/oidc-20220328#oidc-issuer-discovery">6.1 OIDC Issuer Discovery</a>].
+                        </li>
+                        <li>
+                          An agent may have zero or more Solid storages, and any number of these MAY be discoverable from the WebID Profile via statements of the form <code>?webid pim:storage ?storage</code>, where <code>?webid</code> is the WebID and each <code>?storage</code> is the IRI of a storage owned by that agent [<cite><a class="bibref" href="#ref-solid-protocol">Solid Protocol</a></cite> § <a href="https://solidproject.org/TR/2024/protocol-20240512#storage">4 Storage</a>].
+                        </li>
+                      </ol>
```
csarven, line 531: This should also list `ldp:inbox` given that:

* Solid WebID Profile refers to it: https://solid.github.io/webid-profile/#webid-profile-data-model and it is on equal requirement as some other properties like storage.
* Solid Protocol refers to LDN for servers and clients: https://solidproject.org/TR/protocol#linked-data-notifications .
* Data at https://jeff-zucker.github.io/solid-load-profile/ indicates there is substantial amount of deployments out there using it (context: e.g., foaf:name 76, solid:oidcIssuer 71, **ldp:inbox: 66**, space:storage 64, solid:publicTypeIndex: 64, foaf:knows 53). 
* An agent's inbox contributes towards the "social" aspect of Solid.
* ActivityPub's as:inbox aliases ldp:inbox, so there is a wider interop opportunity when they are used interchangeably depending on the context.
* ...

solid26.html
jeff-zucker: > My interpretation of https://solid.github.io/webid-profile/#discovery is that there are two locations you look for seeAlso references:

There is one problem with your interpretation (and perhaps with our original doc): if the WebID Document is hosted outside of Solid, one needs one of those predicates to get from there to extended profile documents.  If any Solid app is to be able to write seeAlso documents (seems like an important function for Solid Apps to be able to do) , in this case, they can't put a pointer to the seeAlso in the WebID doc because it is not on a Solid Resource server.  Therefore the only place they can put a pointer to their new seeAlso is in a seeAlso.

solid26.html
```diff
@@ -423,6 +433,29 @@ <h3>Web Access Control</h3>
...
+                  </div>
+                </div>
+              </section>
+
+              <section id="webid-profile" inlist="" rel="schema:hasPart" resource="#webid-profile">
+                <h3>Solid WebID Profile</h3>
+                <div datatype="rdf:HTML" property="schema:description">
+                  <p>The <a href="https://solid.github.io/webid-profile/">Solid WebID Profile</a> (CG Draft Report, v1.0.0, 12 May 2024) is included with the following comments:</p>
+                  <ul>
+                    <li>Unlike the other pinned specifications, the Solid WebID Profile is <em>not</em> normatively referenced by the Solid Protocol. It is included here as the only specification that describes the structure of a Solid Profile and the predicates clients use to discover its parts.</li>
```
csarven: I don't find this framing helpful. The purpose of solid26 is not about what's referenced from the "pinned specifications". The frame is about what's useful for implementers from the https://solidproject.org/TR/ so they can build useful tools. I suggest expressing this paragraph differently. It should be about why do we need a WebID Profile document? What is it useful for? What are the use cases? It should capture the essence of use cases: https://solid.github.io/webid-profile/#use-cases

solid26.html
jeff-zucker: To accommodate this need for a place to write pointers to seeAlsos, my load-profile app descends exactly two levels of seeAlsos - from the WebID doc to a seeAlso and from there to any seeAlso it lists, but not seeAlso's listed in the second set of seeAlsos.

solid26.html
```diff
@@ -423,6 +433,29 @@ <h3>Web Access Control</h3>
...
               </section>
 
+              <section id="webid-1" inlist="" rel="schema:hasPart" resource="#webid-1">
+                <h3>WebID 1.0</h3>
+                <div datatype="rdf:HTML" property="schema:description">
+                  <p><a href="https://www.w3.org/2005/Incubator/webid/spec/identity/">WebID 1.0</a> (W3C Editor's Draft, 5 March 2014) is included with the following comments:</p>
+                  <div class="note">
+                    <h4><span>Note</span></h4>
+                    <p>Throughout this document, references to "the WebID specification" or [<cite><a class="bibref" href="#ref-webid">WEBID</a></cite>] refer to WebID 1.0 as pinned here. WebID 1.0 has the status of a W3C Editor's Draft from the (now closed) W3C WebID Incubator Group; it has not progressed to a W3C Recommendation. The Solid Protocol and Solid-OIDC normatively reference WebID 1.0 for the definition of the term <em>WebID</em>.</p>
+                  </div>
```
csarven: Is this information useful or need to be mentioned for implementers?

solid26.html
```diff
@@ -455,6 +634,27 @@ <h2>References</h2>
...
+
+                <dt id="ref-vcard">[VCARD-RDF]</dt>
+                <dd><cite><a href="https://www.w3.org/TR/vcard-rdf/">vCard Ontology - for describing People and Organizations</a></cite>. Renato Iannella; James McKinney. W3C. 22 May 2014. W3C Working Group Note. URL: <a href="https://www.w3.org/TR/vcard-rdf/">https://www.w3.org/TR/vcard-rdf/</a></dd>
+
+                <dt id="ref-schema">[SCHEMA-ORG]</dt>
+                <dd><cite><a href="https://schema.org/">Schema.org</a></cite>. W3C Schema.org Community Group. URL: <a href="https://schema.org/">https://schema.org/</a></dd>
+
+                <dt id="ref-bio">[BIO]</dt>
+                <dd><cite><a href="https://vocab.org/bio/">BIO: A vocabulary for biographical information</a></cite>. Ian Davis; David Galbraith. URL: <a href="https://vocab.org/bio/">https://vocab.org/bio/</a></dd>
+
```
csarven: I don't quite follow why this is here.

solid26.html
```diff
@@ -423,6 +433,29 @@ <h3>Web Access Control</h3>
...
+                    <li>Most clients only need <a href="https://solid.github.io/webid-profile/#discovery">§ 2 Discovery</a> and <a href="https://solid.github.io/webid-profile/#reading-profile">§ 3.1 Reading Profile</a>; the <a href="https://solid.github.io/webid-profile/#dfn-writer-application">Writer Application</a> sections (<a href="https://solid.github.io/webid-profile/#updating-profile">§ 3.2</a>, <a href="https://solid.github.io/webid-profile/#extending-profile">§ 3.3</a>) primarily apply to profile editors and onboarding flows.</li>
+                    <li>The companion <a href="https://github.com/solid/webid-profile/tree/main/shapes">shapes</a> in <code>solid/webid-profile</code> &mdash; <a href="https://github.com/solid/webid-profile/blob/main/shapes/profile-discovery-shape.ttl"><code>profile-discovery-shape.ttl</code></a> (discovery predicates) and <a href="https://github.com/solid/webid-profile/blob/main/shapes/profile-shapes.ttl"><code>profile-shapes.ttl</code></a> (profile data) &mdash; can be used by implementers to validate WebID Profiles, though they are not part of the pinned specification.</li>
```
csarven: This whole utterance of "pinned specification" is undermining the purpose of things. There is no value to the reader other than potentially causing confusion. They need to know that either this document is sure / confident about what should be implemented / encouraged and what is not encouraged. Not: "oohh, you know, if you really want to use it, go for it, I guess, but just sayin' that it is not part of the pinned specifications club that we pulled out of thin air..."

csarven (commented):
Thanks for kicking this off. Though I don't understand why this is not carried out in the main solid26 PR. I've only looked through the WebID related parts in this PR, and don't necessary dis/agree or endorse whatever else is included in this document.

Indicating section numbers of referenced material is not particularly reliable generally. For snapshots, the numbers may be reliable but it still reads a bit odd e.g., "§ 3.2". Could leave them out IMO or they should always be used consistently in the whole document.

Left inline comments.

solid26.html
jeff-zucker: > Are there live examples of these different profiles within one Pod?

Not that I know of and I believe there are so many unknowns about what that would cause that we should either warn against it or perhaps better just not mention it.

I was talking about sameAs pointers to profiles on other pods or in non-pod location like Wikipedia.  To be frank, these are not widely used - Tim and Kingsley have many and almost no one else has any.

elf-pavlik:
If Solid26 guide includes Solid WebID Profile and CG Type Indexes (LWS WG has more privacy aware version), I think it should also mention known security & privacy issues with both of those proposals. I can PR them once we agree where exactly they should go.

jeswr:
> Thanks for kicking this off. 

Thanks for the review(s)! 

> Though I don't understand why this is not carried out in the main solid26 PR. I've only looked through the WebID related parts in this PR, and don't necessary dis/agree or endorse whatever else is included in this document.

There is quite a lot of discussion on the other PR. I opened this up separately so that I can easily see and action comments specifically on the WebID without it getting lost amongst all the existing comments on the main PR.

Once I have incorporated most of the feedback from the Solid WebID Profile authors I plan to merge this into the `solid26` branch.

> Indicating section numbers of referenced material is not particularly reliable generally. For snapshots, the numbers may be reliable but it still reads a bit odd e.g., "§ 3.2". Could leave them out IMO or they should always be used consistently in the whole document.

Thanks for the steer. Happy to drop section numbers!

solid26.html
```diff
@@ -224,7 +224,7 @@
     <main>
       <article about="" typeof="schema:Article">
         <div class="head">
-          <h1 property="schema:name">Solid26: Implementation Guide</h1>
+          <h1 property="schema:name">Solid26 Implementors Guide</h1>
```
uvdsl: Could we leave such general changes to the main PR, just for ease of merging?
These do not seem particular to the WebID section :)

```suggestion
          <h1 property="schema:name">Solid26: Implementation Guide</h1>
```

solid26.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
   <head>
     <meta charset="utf-8" />
-    <title>Solid26: Implementation Guide</title>
+    <title>Solid26 Implementors Guide</title>
```
uvdsl: Could we leave such general changes to the main PR, just for ease of merging?
These do not seem particular to the WebID section :)

```suggestion
    <title>Solid26: Implementation Guide</title>
```

solid26.html
```diff
@@ -259,24 +259,24 @@ <h1 property="schema:name">Solid26: Implementation Guide</h1>
             </dl>
             <dl id="document-type">
               <dt>Document Type</dt>
-              <dd>Implementation Guide</dd>
+              <dd>Implementors Guide</dd>
```
uvdsl: Could we leave such general changes to the main PR, just for ease of merging?
These do not seem particular to the WebID section :)
```suggestion
              <dd>Implementation Guide</dd>
```

solid26.html
```diff
@@ -259,24 +259,24 @@ <h1 property="schema:name">Solid26: Implementation Guide</h1>
...
             </dl>
           </details>
-          <p class="copyright" rel="dcterms:rights" resource="#document-rights"><span property="dcterms:description"><a href="https://www.w3.org/policies/#copyright">Copyright</a> © 2026 the Contributors to Solid26: Implementation Guide, under the <a about="" href="https://www.w3.org/community/about/agreements/cla/" rel="schema:license">W3C Community Contributor License Agreement (CLA)</a>. A human-readable <a href="https://www.w3.org/community/about/agreements/cla-deed/">summary</a> is available.</span></p>
+          <p class="copyright" rel="dcterms:rights" resource="#document-rights"><span property="dcterms:description"><a href="https://www.w3.org/policies/#copyright">Copyright</a> © 2026 the Contributors to the Solid26 Implementors Guide, under the <a about="" href="https://www.w3.org/community/about/agreements/cla/" rel="schema:license">W3C Community Contributor License Agreement (CLA)</a>. A human-readable <a href="https://www.w3.org/community/about/agreements/cla-deed/">summary</a> is available.</span></p>
```
uvdsl: Could we leave such general changes to the main PR, just for ease of merging?
These do not seem particular to the WebID section :)
```suggestion
          <p class="copyright" rel="dcterms:rights" resource="#document-rights"><span property="dcterms:description"><a href="https://www.w3.org/policies/#copyright">Copyright</a> © 2026 the Contributors to Solid26: Implementation Guide, under the <a about="" href="https://www.w3.org/community/about/agreements/cla/" rel="schema:license">W3C Community Contributor License Agreement (CLA)</a>. A human-readable <a href="https://www.w3.org/community/about/agreements/cla-deed/">summary</a> is available.</span></p>
```

solid26.html
```diff
@@ -259,24 +259,24 @@ <h1 property="schema:name">Solid26: Implementation Guide</h1>
...
           <hr title="Separator for header" />
         </div>
         <div datatype="rdf:HTML" id="content" property="schema:description">
           <section id="abstract">
             <h2>Abstract</h2>
             <div datatype="rdf:HTML" property="schema:abstract">
-              <p>This document is an implementation guide for Solid26 — a snapshot of the most mature and widely implemented Solid specification versions, collected to help developers and organisations identify a common baseline. It provides practical guidance for implementing the Solid specifications included in this collection.</p>
+              <p>This document is the Solid26 Implementors Guide — a snapshot of the most mature and widely implemented Solid specification versions, collected to help developers and organisations identify a common baseline. It provides practical guidance for implementing the Solid specifications included in this collection.</p>
```
uvdsl: Could we leave such general changes to the main PR, just for ease of merging?
These do not seem particular to the WebID section :)
```suggestion
              <p>This document is an implementation guide for Solid26 — a snapshot of the most mature and widely implemented Solid specification versions, collected to help developers and organisations identify a common baseline. It provides practical guidance for implementing the Solid specifications included in this collection.</p>
```

solid26.html
```diff
@@ -259,24 +259,24 @@ <h1 property="schema:name">Solid26: Implementation Guide</h1>
...
             <h2 property="schema:name">Status of This Document</h2>
             <div datatype="rdf:HTML" property="schema:description">
               <p>This document was published by the <a href="https://www.w3.org/groups/cg/solid/">W3C Solid Community Group</a>. It is not a W3C Standard nor is it on the W3C Standards Track. Please note that under the <a href="https://www.w3.org/community/about/agreements/cla/">W3C Community Contributor License Agreement (CLA)</a> there is a limited opt-out and other conditions apply. Learn more about <a href="https://www.w3.org/community/">W3C Community and Business Groups</a>.</p>
-              <p>This document is intended to provide implementation guidance for the Solid specifications collected in the Solid26 release. The content of this guide is informative and non-normative. It may be updated, replaced, or obsoleted by other documents at any time.</p>
+              <p>This document is intended to provide implementation guidance for the Solid specifications collected in the Solid26 Implementors Guide. The content of this guide is informative and non-normative. It may be updated, replaced, or obsoleted by other documents at any time.</p>
```
uvdsl: Good catch! Still, recommending to solve that in the main PR.
```suggestion
              <p>This document is intended to provide implementation guidance for the Solid specifications collected in the Solid26 release. The content of this guide is informative and non-normative. It may be updated, replaced, or obsoleted by other documents at any time.</p>
```

solid26.html
```diff
@@ -295,14 +295,24 @@ <h2>Table of Contents</h2>
                     <li class="tocline"><a class="tocxref" href="#solid-protocol"><bdi class="secno">2.1</bdi> <span>Solid Protocol</span></a></li>
                     <li class="tocline"><a class="tocxref" href="#solid-oidc"><bdi class="secno">2.2</bdi> <span>Solid-OIDC</span></a></li>
                     <li class="tocline"><a class="tocxref" href="#web-access-control"><bdi class="secno">2.3</bdi> <span>Web Access Control</span></a></li>
+                    <li class="tocline"><a class="tocxref" href="#webid-1"><bdi class="secno">2.4</bdi> <span>WebID 1.0</span></a></li>
+                    <li class="tocline"><a class="tocxref" href="#webid-profile"><bdi class="secno">2.5</bdi> <span>Solid WebID Profile</span></a></li>
```
uvdsl, line 299: Discussion starter: Do we need to have those separate or should we have a joint section on "WebID and Solid WebID Profile"? That could read more concise as a section?

solid26.html
```diff
@@ -295,14 +295,24 @@ <h2>Table of Contents</h2>
...
                 </li>
                 <li class="tocline">
                   <p><a class="tocxref" href="#implementation-guidance"><bdi class="secno">3.</bdi> <span>Implementation Guidance</span></a></p>
                   <ol>
-                    <li class="tocline"><a class="tocxref" href="#webid"><bdi class="secno">3.1</bdi> <span>WebID</span></a></li>
+                    <li class="tocline">
+                      <p><a class="tocxref" href="#webid"><bdi class="secno">3.1</bdi> <span>WebID</span></a></p>
+                      <ol>
+                        <li class="tocline"><a class="tocxref" href="#webid-normative"><bdi class="secno">3.1.1</bdi> <span>Requirements derivable from Solid26 Implementors Guide specifications</span></a></li>
+                        <li class="tocline"><a class="tocxref" href="#webid-non-normative"><bdi class="secno">3.1.2</bdi> <span>Common assumptions not required by Solid26 Implementors Guide specifications</span></a></li>
```
uvdsl: To keep it consistent with main PR right now...
```suggestion
                        <li class="tocline"><a class="tocxref" href="#webid-normative"><bdi class="secno">3.1.1</bdi> <span>Requirements derivable from Solid26 Implementation Guide specifications</span></a></li>
                        <li class="tocline"><a class="tocxref" href="#webid-non-normative"><bdi class="secno">3.1.2</bdi> <span>Common assumptions not required by Solid26 Implementation Guide specifications</span></a></li>
```

solid26.html
```diff
@@ -314,8 +324,8 @@ <h2 property="schema:name">Introduction</h2>
...
                 The <a href="https://solidproject.org/TR/protocol">Solid Protocol</a> bundles the <a href="https://solidproject.org/TR/protocol#abstract">specifications that, together, provide applications with secure and permissioned access to externally stored data in an interoperable way</a>. 
-                Solid26 points implementers to fixed versions of the <a href="https://solidproject.org/TR/protocol">Solid Protocol</a> and its sub-specifications, a stable snapshot of the specifications to build against today. 
-                Solid26 selects for the most widely implemented specification versions at the time of this documents publication.
+                The Solid26 Implementors Guide points implementers to fixed versions of the <a href="https://solidproject.org/TR/protocol">Solid Protocol</a> and its sub-specifications, a stable snapshot of the specifications to build against today. 
+                The Solid26 Implementors Guide selects for the most widely implemented specification versions at the time of this documents publication.
```
uvdsl: Change for main PR...
```suggestion
                Solid26 points implementers to fixed versions of the <a href="https://solidproject.org/TR/protocol">Solid Protocol</a> and its sub-specifications, a stable snapshot of the specifications to build against today. 
                Solid26 selects for the most widely implemented specification versions at the time of this documents publication.
```

solid26.html
```diff
@@ -324,7 +334,7 @@ <h2 property="schema:name">Introduction</h2>
           <section id="specifications" inlist="" rel="schema:hasPart" resource="#specifications">
             <h2 property="schema:name">Specifications</h2>
             <div datatype="rdf:HTML" property="schema:description">
-              <p>This Solid26 Implementation Guide recommends the following specification snapshots.</p>
+              <p>The Solid26 Implementors Guide recommends the following specification snapshots.</p>
```
uvdsl: Change for main PR...
```suggestion
              <p>This Solid26 Implementation Guide recommends the following specification snapshots.</p>
```

solid26.html
```diff
@@ -399,7 +409,7 @@ <h3>Solid-OIDC</h3>
                   <p><a href="https://solidproject.org/TR/2022/oidc-20220328">Solid-OIDC</a> (v0.1.0, 2022-03-28) is included with the following comments:</p>
                     <ul>
                       <li>
-                        <p>Despite Solid26 including Solid-OIDC v0.1.0, it is not widely implemented. In particular, the Solid-OIDC recommended flow of User-Managed Access (UMA) is not supported by any open source Solid Server implementation. Current implementations rely on the access token issued by the OpenID Provider of the user to authenticate the user at a Solid Storage.</p>
+                        <p>Despite the Solid26 Implementors Guide including Solid-OIDC v0.1.0, it is not widely implemented. In particular, the Solid-OIDC recommended flow of User-Managed Access (UMA) is not supported by any open source Solid Server implementation. Current implementations rely on the access token issued by the OpenID Provider of the user to authenticate the user at a Solid Storage.</p>
```
uvdsl: Change for main PR...
```suggestion
                        <p>Despite Solid26 including Solid-OIDC v0.1.0, it is not widely implemented. In particular, the Solid-OIDC recommended flow of User-Managed Access (UMA) is not supported by any open source Solid Server implementation. Current implementations rely on the access token issued by the OpenID Provider of the user to authenticate the user at a Solid Storage.</p>
```

solid26.html
uvdsl: I propose to remove the note. Otherwise we have to have such notes for all the other sections and specifications (including the Solid Protocol itself) as well, and I don't think that is helpful or useful for our cause here.
```suggestion
```

solid26.html
```diff
@@ -423,6 +433,29 @@ <h3>Web Access Control</h3>
                   </div>
               </section>
 
+              <section id="webid-1" inlist="" rel="schema:hasPart" resource="#webid-1">
+                <h3>WebID 1.0</h3>
+                <div datatype="rdf:HTML" property="schema:description">
+                  <p><a href="https://www.w3.org/2005/Incubator/webid/spec/identity/">WebID 1.0</a> (W3C Editor's Draft, 5 March 2014) is included with the following comments:</p>
```
uvdsl:
```suggestion
                  <p><a href="https://www.w3.org/2005/Incubator/webid/spec/identity/">WebID 1.0</a> (W3C Editor's Draft, 5 March 2014) is included.</p>
```

solid26.html
uvdsl: Not sure what to make of that statement... @jeswr, can you make more explicit what you want to imply here? To my understanding the Solid WebID Profile spec reflects what is currently implemented in the Solid ecosystem in general, and to my understanding Solid26 aims to recommend specifications to help implementers achieve broad interoperability. In addition, there are additional comments and considerations outlined below... Just not sure what this bullet point aims to say really :)

solid26.html
uvdsl: I don't see how the shapes must be mentioned at all? these are in the repository of the repo, and (I think) not even mentioned in/by the specification.  I suggest to remove this bullet point.
```suggestion
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
               <section id="webid" inlist="" rel="schema:hasPart" resource="#webid">
                 <h3>WebID</h3>
                 <div datatype="rdf:HTML" property="schema:description">
-                  <div class="note">
-                    <h4><span>Note</span></h4>
-                    <p>Work in progress: the content from the <a href="https://docs.google.com/document/d/1da2J-NsU3K-4kWEFOvhzIdrvy_KftewXdlxfu401kY0/edit?tab=t.j8ef1ds1wza#heading=h.r94tmffvqx0e">WebID guidance document</a> is to be integrated into this section.</p>
-                  </div>
+
+                  <p>This section gives implementation guidance for WebID in the context of the Solid26 Implementors Guide. It is organized into requirements derivable from the pinned specifications (<a href="#webid-normative">§ 3.1.1</a>) and common assumptions not required by them (<a href="#webid-non-normative">§ 3.1.2</a>) (see <a href="#specifications">§ 2 Specifications</a>), and a recommended profile assembly algorithm (<a href="#webid-algorithms">§ 3.1.3</a>). Guidance for client implementors is given in <a href="#client-guidance">§ 4 Guidance for clients</a>.</p>
```
uvdsl:
```suggestion
                  <p>This section gives implementation guidance for WebID in the context of the Solid26 Implementation Guide. It is organized into requirements derivable from the respective specifications (<a href="#webid-normative">§ 3.1.1</a>) and common assumptions not required by them (<a href="#webid-non-normative">§ 3.1.2</a>) (see <a href="#specifications">§ 2 Specifications</a>), and a recommended profile assembly algorithm (<a href="#webid-algorithms">§ 3.1.3</a>). Guidance for client implementers is given in <a href="#client-guidance">§ 4 Guidance for clients</a>.</p>
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+
+                  <p>For the purposes of this section, the following terms are used:</p>
+                  <dl>
+                    <dt><dfn id="dfn-webid">WebID</dfn></dt>
+                    <dd>An IRI that denotes an agent [<cite><a class="bibref" href="#ref-webid">WEBID</a></cite>], [<cite><a class="bibref" href="#ref-solid-protocol">SOLID-PROTOCOL</a></cite>].</dd>
```
uvdsl: Editorial.
```suggestion
                    <dd>An HTTP URI that denotes an agent [<cite><a class="bibref" href="#ref-webid">WEBID</a></cite>], [<cite><a class="bibref" href="#ref-solid-protocol">SOLID-PROTOCOL</a></cite>].</dd>
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                    <dt><dfn id="dfn-webid-profile-document">WebID Profile Document</dfn></dt>
+                    <dd>The Web resource obtained by dereferencing a <a href="#dfn-webid">WebID</a>. Generally public.</dd>
```
uvdsl: > Web resource

Is that term established somewhere in the Solid Protocol or other specifications? If not, I suggest just re-use information resource [WEBARCH]...

```suggestion
                    <dd>The information resource obtained by dereferencing a <a href="#dfn-webid">WebID</a>. Generally public.</dd>
```

But maybe just say RDF document here, as defined by [WEBID]? There is a paragraph about RDF and UNION of RDF graphs below...

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                    <dt><dfn id="dfn-extended-profile-document">Extended Profile Document</dfn></dt>
+                    <dd>A further Web resource linked from the <a href="#dfn-webid-profile-document">WebID Profile Document</a> that contains additional statements about the <a href="#dfn-webid">WebID</a>. Generally permissioned, but available to trusted participants (e.g. friends).</dd>
```
uvdsl:
```suggestion
                    <dd>A further information resource linked from the <a href="#dfn-webid-profile-document">WebID Profile Document</a> that contains additional statements about the <a href="#dfn-webid">WebID</a>. Generally permissioned and available only to trusted participants (e.g. friends).</dd>
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                    <dt><dfn id="dfn-preferences-document">Preferences Document</dfn></dt>
+                    <dd>A private Web resource, linked from the agent's <a href="#dfn-webid-profile">WebID Profile</a>, that holds settings and pointers to private data. Generally accessible only to the user (or a delegated agent).</dd>
```
uvdsl:
```suggestion
                    <dd>A private information resource, linked from the agent's <a href="#dfn-webid-profile">WebID Profile</a>, that holds settings and pointers to private data. Generally accessible only to the user (or a delegated agent).</dd>
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                    <dt><dfn id="dfn-webid-profile">WebID Profile</dfn></dt>
+                    <dd>The RDF graph [<cite><a class="bibref" href="#ref-rdf11-concepts">RDF11-CONCEPTS</a></cite>] formed by the union of the <a href="#dfn-webid-profile-document">WebID Profile Document</a> with any <a href="#dfn-extended-profile-document">Extended Profile Documents</a>, assembled per <a href="#webid-non-normative">§ 3.1.2 Common assumptions not required by Solid26 Implementors Guide specifications</a>.</dd>
```
uvdsl: Consistency with main PR again... :)
```suggestion
                    <dd>The RDF graph [<cite><a class="bibref" href="#ref-rdf11-concepts">RDF11-CONCEPTS</a></cite>] formed by the union of the <a href="#dfn-webid-profile-document">WebID Profile Document</a> with any <a href="#dfn-extended-profile-document">Extended Profile Documents</a>, assembled per <a href="#webid-non-normative">§ 3.1.2 Common assumptions not required by Solid26 Implementation Guide specifications</a>.</dd>
```

solid26.html
uvdsl: As implementer of https://github.com/uvdsl/solid-authorization-app, which consumes the WebID profile of a user, I second that `ldp:inbox` should be included here.

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                        </li>
+                        <li>
+                          An agent may have zero or more Solid storages, and any number of these MAY be discoverable from the WebID Profile via statements of the form <code>?webid pim:storage ?storage</code>, where <code>?webid</code> is the WebID and each <code>?storage</code> is the IRI of a storage owned by that agent [<cite><a class="bibref" href="#ref-solid-protocol">Solid Protocol</a></cite> § <a href="https://solidproject.org/TR/2024/protocol-20240512#storage">4 Storage</a>].
+                        </li>
+                      </ol>
+                    </div>
+                  </section>
+
+                  <section id="webid-non-normative" inlist="" rel="schema:hasPart" resource="#webid-non-normative">
+                    <h4>Common assumptions not required by Solid26 Implementors Guide specifications</h4>
```
uvdsl:
```suggestion
                    <h4>Common assumptions not required by Solid26 Implementation Guide specifications</h4>
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                    <div datatype="rdf:HTML" property="schema:description">
+                      <ol>
+                        <li>
+                          The WebID Document is a public resource (i.e. an unauthenticated HTTP <code>GET</code> on the WebID URI returns the Profile) [<cite><a class="bibref" href="#ref-webid">WebID 1.0</a></cite> § <a href="https://www.w3.org/2005/Incubator/webid/spec/identity/#processing-the-webid-profile">6 Processing the WebID Profile</a>].
+                          <ul>
+                            <li>The pinned specifications do not preclude a server from retrieving a non-public WebID Document via authenticated requests, but this is not well-supported in Social Web settings, where a WebID is dereferenced by many other agents' servers that cannot be assumed to make authenticated requests to WebIDs.</li>
```
uvdsl: This is not a limitation of the use cases from or environment of the open Web. It is merely an implementation limitation to facilitate use cases that desire access control on WebID profile documents.
```suggestion
                            <li>The pinned specifications do not preclude a server, acting as a client, from retrieving a non-public WebID Document via authenticated requests. However, this is generally not well-supported by current server implementations.</li>
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                          </ul>
+                        </li>
+                        <li>
+                          The <code>pim:storage</code> statement in the WebID Profile is the only sound mechanism for discovering an agent's storages. The <code>storageDescription</code> Link header is not to be used for this purpose: a WebID Document hosted in a storage does not imply the WebID owns that storage [<cite><a class="bibref" href="#ref-solid-protocol">Solid Protocol</a></cite> § <a href="https://solidproject.org/TR/2024/protocol-20240512#webid-profile-privacy">Privacy Considerations</a>].
```
uvdsl, line 546:
```suggestion
                          The <code>pim:storage</code> statement in the WebID Profile is the only correct mechanism for discovering an agent's storages. The <code>storageDescription</code> Link header is not to be used for this purpose: a WebID Document hosted in a storage does not imply the WebID owns that storage [<cite><a class="bibref" href="#ref-solid-protocol">Solid Protocol</a></cite> § <a href="https://solidproject.org/TR/2024/protocol-20240512#webid-profile-privacy">Privacy Considerations</a>].
```

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                      <ul>
+                        <li>the <a href="#dfn-webid-profile-document">WebID Profile Document</a>;</li>
+                        <li>any <a href="#dfn-extended-profile-document">Extended Profile Documents</a> linked from it (e.g. via <code>rdfs:seeAlso</code>, <code>foaf:isPrimaryTopicOf</code>, or <code>owl:sameAs</code>);</li>
+                        <li>the agent's <a href="#dfn-preferences-document">Preferences Document</a> (linked via <code>pim:preferencesFile</code>), and any further Extended Profile Documents linked from it;</li>
+                        <li>the agent's public and private Type Index documents (linked via <code>solid:publicTypeIndex</code> and <code>solid:privateTypeIndex</code>).</li>
+                      </ul>
+
+                      <p>To assemble it:</p>
+                      <ol>
+                        <li>Unauthenticated <code>GET</code> the WebID Document. If it is not publicly readable, surface a clear error: the WebID cannot be used for unauthenticated discovery.</li>
```
uvdsl: Visibility of the profile document has been discussed above already. Either a client is able to retrieve it or not, regardless of whether it is public or requires authentication.
```suggestion
                        <li> <code>GET</code> the WebID Document. If it is not  readable, surface a clear error: the WebID cannot be used for discovery.</li>
```

uvdsl (changes requested):
This is part 1 of my review on this.

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                      </ol>
+
+                      <div class="note">
+                        <h4><span>Note</span></h4>
+                        <p>The assembled WebID Profile graph can optionally be validated against <a href="https://github.com/solid/webid-profile/blob/main/shapes/profile-discovery-shape.ttl"><code>profile-discovery-shape.ttl</code></a> from the <a href="https://github.com/solid/webid-profile/tree/main/shapes"><code>solid/webid-profile</code> shapes</a>. The shape covers <code>pim:preferencesFile</code> (exactly 1), <code>pim:storage</code> (0+), <code>rdfs:seeAlso</code> (0+), <code>solid:oidcIssuer</code> (0+), <code>solid:publicTypeIndex</code>/<code>solid:privateTypeIndex</code> (at most 1 each), and <code>foaf:primaryTopicOf</code> (at most 1). Two caveats apply: the shape uses the non-standard <code>foaf:primaryTopicOf</code> rather than the FOAF-defined <code>foaf:isPrimaryTopicOf</code> used elsewhere in this guide; and Solid-OIDC requires at least one <code>solid:oidcIssuer</code>, while the shape permits zero.</p>
```
csarven: Add "`<code>ldp:inbox</code>` (at most 1)" given that:

https://github.com/solid/webid-profile/blob/b012305f5a0c9e10ffd5b8953b766dfbf3425781/shapes/profile-discovery-shape.ttl#L57-L64

>    # INBOX
>    #
>    [
>      sh:path ldp:inbox ;
>      sh:nodeKind sh:IRI ;
>      sh:maxCount 1 ;
>      sh:description "The profile may have 0 or 1 inbox."
>    ] ,

as well as this whole explanation: https://github.com/solid/specification/pull/781/changes#r3107227912

solid26.html
```diff
@@ -433,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                        <li>Repeat step 2 transitively for Extended Profile Documents linked from the Preferences Document.</li>
+                        <li>Discard any retrieved document that is not self-describing for the WebID (no triple with the WebID as subject, nor a <code>foaf:primaryTopic</code> triple naming it), to prevent attribution of arbitrary statements to the WebID.</li>
+                        <li>Union the remaining documents with the WebID Document, dropping <code>solid:oidcIssuer</code> from any source other than the WebID Document: it is authoritative only when sourced from the WebID Document, mirroring the write protection in <a href="https://solid.github.io/webid-profile/#updating-profile">Solid WebID Profile § 3.2</a>. Other discovery predicates (e.g. <code>pim:storage</code>) may legitimately appear in any of these documents.</li>
```
csarven: Add ", `<code>ldp:inbox</code>`" into the parenthesis example.

uvdsl:
@jeswr, it seems that https://github.com/solid/specification/pull/781/commits/a67f2f7e6865a452c3a92a8a61004b8aa830a09d has overwritten the entire document instead of marking just the changes. This has resulted in all the comments, suggestions, and feedback being labeled as outdated and don't show up in the files changed view anymore :/ Could you take a look at this to fix it? (e.g., maybe externally save status, revert and re-commit such that the changes are tracked in-place?)

solid26.html
jeff-zucker: This PR was merged without resolving this comment.   The preference file is not a MUST and may be missing.  If the WebID doc is not on  a Solid Resource server, this would mean that no Solid app can ever create a link to an Extended Profile resource.  Even if there is a preferences file, it would mean that no Solid app can create a public facing Extended Profile resource.  

I can check, but I am not aware of any currently existing links from Preference files to other Extended Profile resources. 

I can see the need to prevent apps writing to the WebID document to prevent issuer hijacking, but I do not see a rationale for preventing them from writing anywhere in the profile.  Even if that is good thing, we are not ready for it. Perhaps in the future when there are dedicated profile apps and rules on servers to enforce their use, we could restrict writing everywhere in the profile graph.  But if this rule goes into place now, a number of Extended Profile resources  that are currently reachable by Solid apps will become unreachable and the logic of most current profile-accessing libraries would need to be rewritten.

uvdsl:
That is some weird GitHub behavior... the local diff displays nicely... let me try to fix this...

EDIT: I think this should have done the trick.

solid26.html
uvdsl: @jeff-zucker, I believe the "merge commit" you are referring to was actually a merge from the upstream/main Solid26 branch into the branch of this PR. So, this PR is not yet merged, and I believe there are many comments that need to be resolved before it can be merged :slightly_smiling_face: 

On the matter:
If a WebID profile is not served by a Solid Resource Server, then indeed not all Clients might be able to update it. But that might even be intentional by design. After all, it is up to the URI owner on what content to put where and how to make it accessible or modifiable. 

On the other hand, I totally see your concerns and I believe they are worth noting down in this guide:
If a WebID profile does not point clients to extended profile documents that they are able and allowed to modify, then some Client might not be able to function in their whole capacity or even function at all.

jeff-zucker: I am NOT suggesting ability to write to the WebID Doc.  I am not even suggesting that the Guide say anything abut writing to extended  profile docs.  What concerns me is the statement that clients should recursively descend through extended links only from the preferences file rather than from the WebID document itself.

>  After all, it is up to the URI owner on what content to put where and how to make it accessible or modifiable.

Nothing about my suggestion changes that.  I only want to be able to depend that *if* I want to place a link to an extended document in an extended document, I can be assured that conforming clients will follow it.

Aside : the logic of allowing writing to the private preferences file but disallow it in a public extended document makes little sense to me.  The preferences file is a much more dangerous target.  Additionally, any app that has access to write to documents restricted to the URI owner can do at least as much damage roaming around a pod than it can just focusing on extended profile documents.

jeff-zucker: An example of the usefulness of links to extended documents in extended documents.  Suppose I have a seeAlso link in my WebID doc to friendsAndFamily.ttl which is accessible to friends and family. And suppose that within that doc there is a seeAlso link to familyOnly.ttl, accessible only to family.  That pattern supports a rich ability to structure privacy within a pod.  [EDIT: this allows class inheritance - familyOnly inherits the properties accessible to freindsAndFamily]

elf-pavlik: I think combination of
* https://github.com/w3c/lws-protocol/pull/115
* https://github.com/w3c/lws-protocol/pull/106

Can offer much cleaner discovery of protected data (eg. profiles) compared to trying one's luck over random `seeAlso` arcs.

jeff-zucker: A type-index is not an appropriate place to store information such as phone#a is available to friends and family but phone#b is available only  to family6.  Those are descriptions of me and belong in my profile.   LWS grants may be cleaner, but this Guide is about what is in wide use now.

jeff-zucker: > trying one's luck over random seeAlso arcs.

They are not random arcs, they are arcs created by the WebID owner for a purpose.

jeff-zucker: @uvdsl wrote

> If a WebID profile does not point clients to extended profile documents that they are able and allowed to modify, then some Client might not be able to function in their whole capacity or even function at all.

Although I still believe that the ability to modify is an important, I am not asking for that to be included in this iteration of the Guide.  My only ask is that the section on gathering the profile graph 

> Repeat step 2 transitively for Extended Profile Documents linked from the Preferences Document.

be amended to use "WebID Document" in place of "Preferences Document".  That will not say anything about who or what writes the links.

elf-pavlik: In user/agent centric data spaces, where the description of the user/agent in the center really stops? One can argue that my calendar, my geo-locations etc. are part of 'my profile'. Back in myspace days people were including all kind of wild stuff in their 'profile'.
I think the guide should clearly mark it as WIP and possibly point people to
* https://github.com/w3c-cg/solid/issues/63
* https://github.com/w3c-cg/solid/issues/61

And signal that breaking changes are very likely in the future.

solid26.html
```diff
@@ -455,15 +466,161 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                      </ol>
+
+                      <div class="note">
+                        <h4><span>Note</span></h4>
+                        <p>The assembled WebID Profile graph can optionally be validated against <a href="https://github.com/solid/webid-profile/blob/main/shapes/profile-discovery-shape.ttl"><code>profile-discovery-shape.ttl</code></a> from the <a href="https://github.com/solid/webid-profile/tree/main/shapes"><code>solid/webid-profile</code> shapes</a>. The shape covers <code>pim:preferencesFile</code> (exactly 1), <code>pim:storage</code> (0+), <code>rdfs:seeAlso</code> (0+), <code>solid:oidcIssuer</code> (0+), <code>solid:publicTypeIndex</code>/<code>solid:privateTypeIndex</code> (at most 1 each), and <code>foaf:primaryTopicOf</code> (at most 1). Two caveats apply: the shape uses the non-standard <code>foaf:primaryTopicOf</code> rather than the FOAF-defined <code>foaf:isPrimaryTopicOf</code> used elsewhere in this guide; and Solid-OIDC requires at least one <code>solid:oidcIssuer</code>, while the shape permits zero.</p>
```
jeff-zucker: I have created a pull to use the correct foaf predicate (https://github.com/solid/webid-profile/pull/125).
For issuer, it seems better to say that although you can have 0 issuers, in most cases that will prevent your WebID from being validated.

jeswr:
```suggestion
                        <p>The assembled WebID Profile graph can optionally be validated against <a href="https://github.com/solid/webid-profile/blob/main/shapes/profile-discovery-shape.ttl"><code>profile-discovery-shape.ttl</code></a> from the <a href="https://github.com/solid/webid-profile/tree/main/shapes"><code>solid/webid-profile</code> shapes</a>. The shape covers <code>pim:preferencesFile</code> (exactly 1), <code>pim:storage</code> (0+), <code>rdfs:seeAlso</code> (0+), <code>solid:oidcIssuer</code> (0+), <code>solid:publicTypeIndex</code>/<code>solid:privateTypeIndex</code> (at most 1 each), and <code>foaf:isPrimaryTopicOf</code> (at most 1). Note that Solid-OIDC requires at least one <code>solid:oidcIssuer</code> in order to attest ownership of the WebID, while the shape permits zero.</p>
```

solid26.html
```diff
@@ -295,12 +295,20 @@ <h2>Table of Contents</h2>
                     <li class="tocline"><a class="tocxref" href="#solid-protocol"><bdi class="secno">2.1</bdi> <span>Solid Protocol</span></a></li>
                     <li class="tocline"><a class="tocxref" href="#solid-oidc"><bdi class="secno">2.2</bdi> <span>Solid-OIDC</span></a></li>
                     <li class="tocline"><a class="tocxref" href="#web-access-control"><bdi class="secno">2.3</bdi> <span>Web Access Control</span></a></li>
+                    <li class="tocline"><a class="tocxref" href="#webid-1"><bdi class="secno">2.4</bdi> <span>WebID 1.0</span></a></li>
```
jeff-zucker, line 298: Add in WebID Profile draft.

solid26.html
```diff
@@ -453,143 +466,90 @@ <h2 property="schema:name">Implementation Guidance</h2>
...
+                      <ol>
+                        <li>Unauthenticated <code>GET</code> the WebID Document. If it is not publicly readable, surface a clear error: the WebID cannot be used for unauthenticated discovery.</li>
+                        <li>For each linked document (Extended Profile Documents via <code>rdfs:seeAlso</code>/<code>foaf:isPrimaryTopicOf</code>/<code>owl:sameAs</code>; the Preferences Document via <code>pim:preferencesFile</code>; Type Index documents via <code>solid:publicTypeIndex</code>/<code>solid:privateTypeIndex</code>), <code>GET</code> the document. Public documents may be fetched unauthenticated; on <code>401</code>/<code>403</code> with a logged-in user, retry authenticated. Preferences and the private Type Index will normally require authentication [<cite><a class="bibref" href="#ref-webid-profile">Solid WebID Profile</a></cite> § <a href="https://solid.github.io/webid-profile/#discovery">2 Discovery</a>, § <a href="https://solid.github.io/webid-profile/#private-preferences">4 Private Preferences</a>].</li>
+                        <li>Repeat step 2 transitively for Extended Profile Documents linked from the Preferences Document.</li>
```
jeff-zucker: "Preferences Document" should be replaced with "WebID Document".  I see no reason to forbid extended document links in any extended document.  This is not about writing one way or the other - users should be able to create pathways of links between extended profile documents.  That does not specify anything about who or which apps  can create the links.  For further discussion of this point see https://github.com/solid/specification/pull/781/changes/8d3dbe409db45c5b769aa9a86ac26d59e991aad9#r3118413017 and https://github.com/solid/specification/pull/781/changes/8d3dbe409db45c5b769aa9a86ac26d59e991aad9#r3118590667 .

elf-pavlik:
@jeswr you wanted my comments on Security and Privacy captured somewhere besides minutes, In short

### Security of `solid:oidcIssuer`

* Demo of very simple exploit: https://github.com/elf-pavlik/mallory#cuckoo
* Section suggesting that this may be mitigated but with 0 implementations https://solid.github.io/webid-profile/#protected-properties

### Privacy for Type Indexes

* WG darft includes https://pr-preview.s3.amazonaws.com/w3c/lws-protocol/pull/115.html#security-and-authorization
   CG draft discloses existance of types even if requesting credentials don't allow access to those resources.

jeff-zucker:
The security of the issuer triple is an important issue, but it is the responsibility of the server not the client.  There is nothing a client needs to or can do to change that.  If my app writes a seeAlso triple in the document, that does nothing at all to impact the security of the document.   So what exactly is the advice to clients here?  Don't use servers that expose the issuer?  

The  exposure of the type index is a red herring.  There is nothing in the draft that says an app MUST place pointers to its data in a type-index.   So a caveat in the text might be something like "If you do not want to expose the types used by your app, don't put them in the type index".    What would be the point of having them in the type index if no other app is supposed to see them?

elf-pavlik:
> So what exactly is the advice to clients here? Don't use servers that expose the issuer?
[...]
>  "If you do not want to expose the types used by your app, don't put them in the type index".

I don't see Soild 26 being published just for app developers, but even if someone develops an app that works with sensitive data, it could warn user about their WebID not being protected from identity high-jacking / impersonation and asking to use a different WebID or move their WebID to a secure provider.

And yes, if someone develops app where wide disclosure of specific types existence in user's storage is undesired. They shouldn't add those types until the issue is addressed, for example by work happening in LWS WG.

Those considerations are useful more broadly, especially for anyone who needs to meet certain security and privacy requirements. If we have specific well known issues, I would prefer to disclose them up front, rather than let someone invest weeks or months into some work just to realize that those issues exist and are a show stopper for them.

jeff-zucker:
Would this phrasing work ?

### Application implementers should                                                   
                                                                                  
  * be aware that the type indexes are for sharing type information between apps. Apps that do not want to share type information with other apps, should not use the type indexes.                                                                 
                                                                                  
### End-Users should                                                                  
                                                                                  
  *  be aware that Solid resource servers which do not protect solid:oidcIssuer triples are insecure.                                                              
                                                                                  
### Server implementers should                                                        
                                                                                  
  * provide protection for solid:oidcIssuer triples

elf-pavlik:
In general I wouldn't boil everything down to telling peope what they should and shouldn't do. Mostly inform them how things work, provide some known options and let them make decisions that meet their requirements. 

For CG Type Indexes it may be enough to explain that they don't respect authorization policies on indexed resources and disclose existence and locations of all indexed types. Mentioning LWS WG work could be an option here. 
https://pr-preview.s3.amazonaws.com/w3c/lws-protocol/pull/115.html#security-and-authorization
SAI discovery also respects authorizations on indexed resources but I don't see need to mention that exploration in Solid 26.

For `solid:oidcIssuer`, it may be enough to clarify that while Solid WebID Profile requires servers to protect those triples. No known servers implement it so it can't be relied upon without verifying specific implementation being used. As option hosting WebID Documents elsewhere (not in Soild storage) is something widely used and available now.
 
https://github.com/elf-pavlik/mallory#cuckoo demonstrates the problem and it could be summed up in few sentences without referencing that demo.

csarven:
There is a disconnect here that is being repeated. One of the key aspects of the Solid pitch/vision is that individuals will be able to control their own identities, shape their data, and use applications of their choosing. That is kind of the ground truth/expectation. If there is not a sensible way for individuals to express their profiles using Solid apps, then that creates an obvious tension against the pitch/vision.

The concern here does not just boil down to potential modification of the issuer statement but whether that issuer is in the right place to begin with. Nevertheless, that is the card we have been dealt. Solid WebID Profile is at the very least trying to incorporate the issuer as part of a profile description and with some measures for servers to protect it. When agents and applications with write access can modify them on servers that do not protect it, given some context, that does not entail that server protection is faulty, or the specification is useless, or that we should be dropping everything in favour of this other thing with zero track record. One of the simplest explanations may be that servers are not (yet) conforming to it (based on the understanding that the issuer triple needs to exist in a profile description). It could also mean there is a lack of interest, care, or clarity on the matter as to why. It would be useful to investigate that in and of itself. Jumping from the observation that there is a lack of implementation as to why it is not protected, in a similar way to protecting containment triples, to the specification being wrong about suggesting Solid profiles should not be modifiable by Solid servers is rather a strange expectation, let alone one that runs against the Solid pitch/vision. The scope of Solid WebID Profiles is literally the Solid environment. It is not talking about all other places in the universe in which WebIDs may exist or how they should be managed. That is obvious. Those things do exist. Solid WebID Profiles is focused on explaining how to describe and to some extent manage profiles in the Solid universe, not anywhere else.

The argument raised against this is also very similar to how some server implementations have not taken steps to put additional inbox management measures in place to prevent misuse, e.g., against agents and applications having append access. This is not because the notion and functionality of inbox is bad on its own, but simply because implementations need to take additional considerations into account for their environment. This was all well understood and discussed in several circles in the past decade, and if not, for several decades. So it is certainly not some groundbreaking revelation that when an agent and an application are given append access, they can append to an inbox. Those are literally conforming implementations, doing exactly what they are supposed to for interop. Everything else is completely orthogonal. It is also worth noting that normalising workarounds as interim advice in specification text, e.g., **suggesting individuals host their WebID elsewhere or avoid using certain features, actively contradicts the Solid value proposition rather than preserving it**.

On a related note, I would just add that if we are dropping down to this level of scrutiny, because some deem it to be a necessary criterion of sorts, then we should certainly apply that consistently across every specification mentioned in solid26, including those that do not have significant uptake, and whether each feature in their specification has a test and a publicly verifiable implementation implementing it. If not, there should be strong advisory against using those specifications.

elf-pavlik:
> If there is not a sensible way for individuals to express their profiles using Solid apps, then that creates an obvious tension against the pitch/vision.

With the notion of *extended profile* one can express whatever they like without compromising integrity of main WebID Document which is the only valid source for `solid:oidcProvider` I recall @mrkvon having some approach that fits current state of Solid.

I believe this will also fit beter the direction taken by LWS WG where https://www.w3.org/TR/cid/ will be the existing web standard used.

solid26.html
jeswr: https://github.com/solid/specification/pull/781/commits/dc7a48b6deb4b5e8e73879b3fc0aca7d50870b73

solid26.html
jeswr: https://github.com/solid/specification/pull/781/changes/2a984a4a11465ef2501619bed7f7acd053ebe7de

solid26.html
jeff-zucker: Either works for me.

solid26.html
jeff-zucker: I think they are useful.  We could add a mention to them in the draft if needed.

solid26.html
renyuneyun: This section is said to be a "Guidance for clients", so I feel it already implies that this is the advised order of retrieving/finding (and maybe putting) information.

solid26.html
jeswr: I would be thinking about the guidance as follows "It is desirable to provide a means for Solid clients to modify content of the Solid WebID profile. Therefore, when a WebID is itself not placed in Solid storage -- it is advised that the WebID links to an extended profile document which is located in a Solid storage that clients can edit."

solid26.html
jeswr: I agree that this statement doesn't help or add anything. I have remove it.

solid26.html
jeswr: @jeff-zucker yes I think it might be better to mention them upstream. I have opened https://github.com/solid/webid-profile/issues/126 and will commit @uvdsl's suggestion.

solid26.html
jeswr: Thanks @jeff-zucker, @uvdsl, @elf-pavlik. Addressed in two steps:

- `11316c9` — one bounded level of traversal from Extended Profile Documents linked directly from the WebID Document; discovery links in documents reached via the Preferences Document or Type Index documents are not followed. A note explains why one level of traversal is needed when the WebID Document is not hosted in Solid storage.
- `4946a84` — replaces the earlier document-level self-describing filter with the current Solid WebID Profile §Discovery phrasing: the assembled profile is 'all statements that have the WebID as subject or object'. The asymmetric primaryTopic-filter concern jeff raised on [2a984a4](https://github.com/solid/specification/commit/2a984a4a11465ef2501619bed7f7acd053ebe7de#r183211156) is also addressed by this commit.

solid26.html
jeswr: Addressed in `c8f9754` — dropped PodBrowser, added dokieli. Also refreshed the full predicate list in `2f2d6b6` from what dokieli and SolidOS profile-pane actually read.

solid26.html
jeswr: Thanks @csarven, @renyuneyun. Worked through this in three commits:

- `544b010` — ordering framed as advisory.
- `5acc54a` — further softened per discussion: the list is presented as 'fallback predicates applications may wish to consider using' rather than normative equivalence.
- `2f2d6b6` — refreshed the whole list from dokieli and SolidOS profile-pane. The Avatar entry now follows dokieli's actual order (`vcard:hasPhoto`, then `as:image`, `as:icon`, `foaf:img`, `schema:image`, `vcard:photo`, `sioc:avatar`, `foaf:depiction`), and Name flips to `foaf:name`-first with `as:name` and structured-name fallbacks. Social accounts and richer organisation/skills/languages entries added too.

solid26.html
jeswr: Addressed in `159e511` — added a sub-bullet to the normative hosting item: when the WebID Document is not in a Solid storage, it is recommended to link at least one Extended Profile Document that is, so clients have a writable attachment point for further profile statements.

solid26.html
jeswr: Addressed in `3f512f2` — the WebID Profile Document MUST be available in `text/turtle` and `application/ld+json`, citing both WebID 1.0 §Publishing the WebID Profile Document and Solid WebID Profile §Reading Profile.

solid26.html
jeswr: Addressed in `7743a51` — dropped the unused [BIO] reference.

solid26.html
jeswr: Keeping §2.4 WebID 1.0 and §2.5 Solid WebID Profile as separate subsections. §2 is the list of Solid26 Implementation Guide specifications with one spec per subsection; merging the two would break that pattern.

solid26.html
jeswr: Addressed in `e55840d` — WebID definition now reads 'An HTTP URI that denotes an agent'.

solid26.html
jeswr: Addressed in `e55840d` — dropped 'Web resource' (not an established term) from the four term definitions. WebID Profile Document now simply reads 'The document obtained by dereferencing a WebID.'

solid26.html
jeswr: Addressed in `e55840d` — Extended Profile Document is now 'A further document linked from the WebID Profile Document that contains additional statements about the WebID; retrieval may require authentication.'

solid26.html
jeswr: Addressed in `e55840d` — Preferences Document is now 'A document, linked from the agent's WebID Profile via `pim:preferencesFile`, that holds settings and pointers to private data; typically accessible only to the agent or a delegate.'

solid26.html
jeswr: Addressed in `86880ff` — step 1 now reads: 'GET the WebID Document. If it cannot be retrieved, surface a clear error.'

solid26.html
jeswr: The Solid WebID Profile TOC entry was added in `d10c40967e84a520e735eb0fcfc6170cae619b87` (now §2.5).

solid26.html
jeswr: Addressed in `11316c9` and `4946a84`. See also the longer discussion on the sibling thread above: the algorithm now does one bounded level of traversal from Extended Profile Documents linked directly off the WebID Document, and retention is governed by the statement filter 'WebID as subject or object'.

jeswr:
On the security/privacy concerns raised by @elf-pavlik:

- **`solid:oidcIssuer` protection.** Added a new §3.1.4 Security Considerations in `1a8508c` (later softened in `656d603`). It reads:
  > [Solid WebID Profile §Protected properties] requires servers hosting a WebID Document to protect `solid:oidcIssuer` triples from modification by non-owners. Not all Solid servers implement this protection; on a server that does not, any agent with write access to the document can modify the WebID's issuer.

- **Type-index disclosure.** The broader point about data-discovery mechanisms (including Type Indexes, SAI, the LWS PR, and others) is noted in `be009c2` at the end of §4:
  > Data discovery beyond the profile varies across the ecosystem. Mechanisms in active use include: Type Indexes; Solid Application Interoperability (SAI); SPARQL or Quad Pattern Fragments endpoints; fixed paths under the root of a storage; and DCAT catalogues. This guide does not prescribe a single mechanism.

Happy to refine the wording or expand the type-index privacy framing if this doesn't capture what you had in mind.

jeswr:
Thanks to everyone who contributed to the WebID guidance here — @uvdsl, @csarven, @jeff-zucker, @elf-pavlik, @renyuneyun, and the authors of the Solid WebID Profile — the review comments have meaningfully shaped the shape and tone of this section. I've just pushed a run of commits that work through the open threads (bounded ext-profile traversal, spec-aligned statement filter, refreshed predicate list from dokieli / SolidOS profile-pane, §3.1.4 Security Considerations for `solid:oidcIssuer`, data-discovery note in §4, section-number cleanup, and a language pass), and resolved each review thread with a reply linking the commit.

If any of the resolutions feel unsatisfactory or a discussion is worth continuing, please flag it on the main Solid26 PR #776 so it can be carried forward and resolved there.

PR #782, opened 2026-04-22
Create 2026-04-22.md

elf-pavlik:


meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+### Solid26
+
+#### Solid26 - Solid-OIDC
+https://github.com/solid/solid-oidc/pull/252
+
+* CB: Addresses https://github.com/solid/solid-oidc/issues/250 and captures the authentication flow that all current server and client implementations support (as of 2026-04-21).
+* ...: This specification version is a snapshot of Solid-OIDC, where a Client uses a global DPOP-bound access token issued by an OpenID Provider to authenticate at an Resource Server.
+* eP: was the intent for Client Credentials to be included in Solid26?
```
TallTed:
```suggestion
* eP: Was the intent for Client Credentials to be included in Solid26?
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* CB: https://github.com/solid/solid-oidc/pull/252#issuecomment-4290449076
+* CB: Current Solid-OIDC 0.1 recommends flow for example using UMA. It's fine but it is not widely implemented. I'm not aware of open source implementation. That prompted us to reflect what actually is implemented across servers. The idea was, with history knowledge of Pavlik, that there had been version that uses global access token and we can dig it up.
```
TallTed:
```suggestion
* CB: Current Solid-OIDC 0.1 recommends flow for example using UMA. It's fine but it is not widely implemented. I'm not aware of open source implementation. That prompted us to reflect what actually is implemented across servers. The idea was, based on Pavlik's historic knowledge, that there had been a version that uses a global access token and we can dig it up. 
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* eP: was the intent for Client Credentials to be included in Solid26?
+* CB: https://github.com/solid/solid-oidc/pull/252#issuecomment-4290449076
+* CB: Current Solid-OIDC 0.1 recommends flow for example using UMA. It's fine but it is not widely implemented. I'm not aware of open source implementation. That prompted us to reflect what actually is implemented across servers. The idea was, with history knowledge of Pavlik, that there had been version that uses global access token and we can dig it up. 
+* ...: I did that in this PR
+* eP: From my reviewing, it seems good.
+* AC: Thank you for going through this, and identifying the features that we want to retain and the features that might be retired. There are two feature changes in this PR, (a) one is removing as_uri header (which I would be in favor), and (b) it changes the effective token from an ID token to the access token (which seems an okay change, given how the community does it). The current requirement UMA is a SHOULD, which dropping it would not have an effect much. Any implementation that wants can implement it on-top.
```
TallTed:
```suggestion
* AC: Thank you for going through this, and identifying the features that we want to retain and the features that might be retired. There are two feature changes in this PR: (a) it removes `as_uri` header (of which I would be in favor), and (b) it changes the effective token from an ID token to an access token (which seems an okay change, given how the community does it). The current requirement of UMA is a SHOULD, so dropping it would not have much effect. Any implementation that wants can implement it on top.
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* ...: The change from ID token to access token is more controversial.
+* ...: What I would prefer is to put these changes to main and publish 0.2, such that we don't recommend an old version
+* eP: I think I agree with AC, here. There is also LWS authentication suites. By putting it 0.2, this would signal to people that this is what is currently there.
```
TallTed:
```suggestion
* ...: What I would prefer is to put these changes to main and publish 0.2, such that we don't recommend an old version.
* eP: I think I agree with AC, here. There are also LWS authentication suites. Publishing 0.2 would signal to people what is currently there.
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* CB: Thank you for all the feedback Aaron. Earlier I said that those changes are already what is out there. I'm confused on whether the change from ID Token to Access Token would be that controversial? Not sure here.
+* SC: I don't quite follow the suggestion to deprecate something or whether the current state of Solid-OIDC will be considered or unstable/stable w.r.t. to what the Solid Protocol is before suggesting to swap it for anything.
```
TallTed:
```suggestion
* CB: Thank you for all the feedback, Aaron. Earlier I said that those changes are what is already out there. I'm confused on whether the change from ID Token to Access Token would be that controversial? Not sure here.
* SC: I don't quite follow the suggestion to deprecate something or whether the current state of Solid-OIDC will be considered or unstable/stable w.r.t. to what the Solid Protocol is, before suggesting to swap it for anything.
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* ...: I think authentication mechanisms for Solid is a consideration on its own. I'd like to understand better the discussion around whether or how Solid-OIDC is good enough to a point.
+* AC: The change from ID to access token is conceptually significant, personally I would support that, and the community might not see this as controversial.
```
TallTed:
```suggestion
* ...: I think authentication mechanisms for Solid are a consideration on their own. I'd like to better understand the discussion around whether or how Solid-OIDC is good enough to a point.
* AC: The change from ID to access token is conceptually significant. Personally, I would support that, and the community might not see this as controversial.
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* ...: SC, you are right: Authentication you might use in the browser might be very different, but authentication on the CLI might better work another way.
+* ...: Solid-OIDC works well in the browser but maybe not in a different environment. So maybe, we should have affordances to work with different authentication mechanisms.
```
TallTed:
```suggestion
* ...: SC, you are right: Authentication you might use in the browser might be very different, but authentication on the CLI might work better another way.
* ...: Solid-OIDC works well in the browser, but maybe not in a different environment. So maybe, we should have affordances to work with different authentication mechanisms.
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* eP: https://elf-pavlik.github.io/lws-auth/view/oidc/?dynamic=sequence AC, might LWS authentication suites superseed Solid OIDC. Question to AC and JW, can LWS provide something that superseedes Solid-OIDC or are they redundant.
+* AC: There are two perspective:
+* ...: a server might implement multiple authentication suites. LWS relies on OAuth. In Solid-OIDC, there are some places where OIDC might not forbid it but it is also not defining what we specify and use. So this goes out of what is already defined. LWS is just using vanilla OAuth. It does not forbid using different authentication suites.
```
TallTed: Hard to tell what the second perspective is.
```suggestion
* eP: https://elf-pavlik.github.io/lws-auth/view/oidc/?dynamic=sequence AC, might LWS authentication suites supersede Solid OIDC? Question to AC and JW: can LWS provide something that supersedes Solid-OIDC, or are they redundant?
* AC: There are two perspectives:
* ...: A server might implement multiple authentication suites. LWS relies on OAuth. * * ...: In Solid-OIDC, there are some places where OIDC might not forbid it but it is also not defining what we specify and use. So this goes out of what is already defined. LWS is just using vanilla OAuth. It does not forbid using different authentication suites.
```

meetings/2026-04-22.md
```diff
@@ -0,0 +1,192 @@
...
+* JW: There might be a subset of the authentication suites defined in LWS, that the CG might be intereseted in.
+* eP: From my perspective, I miss a kind of a roadmap what might be available to people in 2 years.
+* SC: I think there is a bunch of stuff that needs to happen before the CG can consider to adopt something. LWS auth work is not incubated. And no horizontal reviews yet. This includes scrutinizing what would be entailed by adopting a particular specification. Is it a mere swap of the auth spec or other specs need to change? So what kind of baggage there may be? Implications of a change should be clear to the ecosystem / communtiy before jumping into concrete ideas to adopt something. LWS is not the only thing out there as far as spec goes that might be interesting to Solid. I would caution to re-visit what Solid needs to fulfill the vision.
```
TallTed:
```suggestion
* JW: There might be a subset of the authentication suites defined in LWS, that the CG might be interested in.
* eP: From my perspective, I miss a kind of roadmap that might be available to people in 2 years.
* SC: I think there is a bunch of stuff that needs to happen before the CG can consider to adopt something. LWS auth work is not incubated. And no horizontal reviews yet. This includes scrutinizing what would be entailed by adopting a particular specification. Is it a mere swap of the auth spec, or would other specs need to change? What kind of baggage might there be? Implications of a change should be clear to the ecosystem/community before jumping into concrete ideas to adopt something. LWS is not the only spec out there that might be interesting to Solid. I would caution to re-visit what Solid needs to fulfill the vision.
```

PR #783, opened 2026-04-24
solid26: draft WAC/ACP wording for CG discussion

jeswr:
A preview link can be found [here](https://htmlpreview.github.io/?https://github.com/solid/specification/blob/feat/solid26-wac-acp-draft/solid26.html).

## Purpose

Action item from the 2026-04-22 CG call: draft a neutral wording change for the WAC/ACP framing in solid26. It is not a proposal for merge on review — it is text to amend in-place until the CG is ready to take a position.

Raised as a separate PR so the access-control wording can be worked on without blocking the rest of solid26 (#776).

## Seeking consensus

@elf-pavlik, @uvdsl, @thhck  — as CG chairs, how would you like to proceed with getting group consensus on this text? @uvdsl, you're also a co-editor of solid26 — are you ok with this approach wearing both hats?

My suggestion: a 👍 / 👎 reaction vote on this PR<s>, open until **Wednesday 2026-04-29**</s>, with line-level edit suggestions welcome in the meantime. Happy to follow a different process if you prefer.

## Discussion context

- CG call 2026-04-15: [meeting notes](https://github.com/solid/specification/blob/main/meetings/2026-04-15.md)
- CG call 2026-04-22: [meeting notes](https://github.com/solid/specification/blob/main/meetings/2026-04-22.md) (action item at the end: draft initial text)
- Main solid26 PR: #776

PR #785, opened 2026-04-26
solid26: placeholder for access-control language pending CG consensus

jeswr:
A preview link can be found [here](https://htmlpreview.github.io/?https://github.com/solid/specification/blob/feat/solid26-access-control-placeholder/solid26.html).

## Purpose

Replaces the WAC/ACP content in §2 with placeholder text marked "subject to discussion before inclusion" so the wider Solid26 launch can proceed without holding on the unresolved access-control-language wording. The proposed access-control-language text itself is being worked on in #783 and will land here once consensus is reached.

## What's in this PR

- **Top-of-document notice** — states the whole document is a draft, with §Web Access Control and §Access Control Policy at primary risk.
- **§2 specifications table** — WAC version cell becomes "Subject to discussion before inclusion"; new ACP row added with the same.
- **TOC** — §2.4 Access Control Policy added; WebID 1.0 and Solid WebID Profile renumbered to §2.5 / §2.6.
- **§2.1 Solid Protocol commentary** — the WAC-encouragement bullet and the March 2026 survey note are replaced with a placeholder bullet pointing at §2.3 / §2.4. The "Some Clients…" architectural-separation bullet is kept.
- **§2.3 Web Access Control** and **§2.4 Access Control Policy (new)** — both contain `class="issue"` placeholder boxes pointing at #783.

## Review

@uvdsl as co-editor most directly affected by the placeholder approach — does this shape work for you, or want to adjust before we land it?

## Test plan

- [x] Render `solid26.html` and confirm the four `class="issue"` boxes (top-of-doc banner, §2.1 bullet, §2.3, §2.4) render visually distinct.
- [x] §2 specifications table shows "Subject to discussion before inclusion" for WAC and ACP rows.
- [x] TOC shows §2.4 Access Control Policy and renumbered §2.5 / §2.6.

### DIFF
--- a/solid26.html
+++ b/solid26.html
@@ -367,7 +367,7 @@ content:counter(appendix, upper-alpha) "." counter(sub-appendix) "\00a0";
                     <td><a href="#web-access-control">Link</a></td>
                   </tr>
                   <tr>
-                    <td>Access Control Policy</td>
+                    <td><a href="https://solidproject.org/TR/2022/acp-20220518">Access Control Policy</a></td>
                     <td><em>Subject to discussion before inclusion</em></td>
                     <td><a href="#access-control-policy">Link</a></td>
                   </tr>
--- a/solid26.html
+++ b/solid26.html
@@ -385,7 +385,10 @@ content:counter(appendix, upper-alpha) "." counter(sub-appendix) "\00a0";
                       <p>If a server does not support HTTP <code>PATCH</code>, a client can read-modify-write via <code>GET</code> then <code>PUT</code>. To avoid the lost-update problem [<a href="https://www.w3.org/1999/04/Editing/">W3C &mdash; Editing the Web</a>], issue the <code>PUT</code> conditionally with <code>If-Match</code> carrying the <code>ETag</code> returned from the <code>GET</code>; the server responds with <code>412 Precondition Failed</code> if the resource has changed in the interim. If only <code>Last-Modified</code> is exposed, <code>If-Unmodified-Since</code> serves the same role, though its one-second resolution is coarser than <code>ETag</code>-based validation. See <a href="https://httpwg.org/specs/rfc9110.html#response.validator">RFC 9110 &sect; Validator Fields</a> for the full mechanics.</p>
                     </li>
                     <li>
-                      <p>The Solid Protocol's access-control requirements (Web Access Control and Access Control Policy) are <em>subject to discussion before inclusion</em>. See <a href="#web-access-control">§ Web Access Control</a> and <a href="#access-control-policy">§ Access Control Policy</a>.</p>
+                      <div class="warning">
+                        <h4><span>Subject to discussion before inclusion</span></h4>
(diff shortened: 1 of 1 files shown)


uvdsl (changes requested):
Either the document will be published reflecting CG consensus or it will not be published. Adding such a note of missing consensus (in-fact even agreement), while declaring that it is published by the W3C Solid CG, is a contradiction.

This document is not yet published and is proposed as a PR where editing is actively taking place. It is a work in progress, that is quite apparent.

If you want to publish this document, then it is you (or an organization) who publishes the document. That published document MUST NOT declare that it is published or even endorsed by the CG because it is not. The proposed change is unfit to for that purpose. 

From the description of this PR, I don't quite grasp the goal here? Could you clarify? 

---

I note that this PR tries to add in ACP as a recommended access control language here again, which is unnecessary (as a different PR tries to do that already) and not reflecting agreement of the group (https://github.com/solid/specification/pull/783#issuecomment-4325498262).

jeswr:
I was of the understanding that the CG was _not_ at consensus on the text around access control -- for the reasons that @elf-pavlik just identified on the mailing list (see below).

The goal of this PR is to have a draft copy of the document that we are all comfortable pointing people to for the duration of this week; by omitting those topics still under discussion. I will make some suggested changes removing text that implies this document is published.

This PR does not even need to be merged. I just would like to have a document where the preview link (https://htmlpreview.github.io/?https://github.com/solid/specification/blob/feat/solid26-access-control-placeholder/solid26.html) contains text we can agree on.

> I note that this PR tries to add in ACP as a recommended access control language here again, which is unnecessary (as a different PR tries to do that already) and not reflecting agreement of the group (https://github.com/solid/specification/pull/783#issuecomment-4325498262).

It adds it with a note that the section could be removed from the spec; since there is not currently consensus on whether or not it would be in.

```
Hi Christoph,

I'm writing from mobile mostly to ask for a quick clarification on CG process.

https://www.w3.org/community/solid/charter/#decision-policy

"To afford asynchronous decisions and organizational deliberation, any resolution (including publication decisions) taken in a face-to-face meeting or teleconference will be considered provisional. A call for consensus (CfC) will be issued for all resolutions (for example, via email, GitHub issue, or web-based survey), with a response period of five to ten working days, depending on the chair’s evaluation of the group consensus on the issue. If no objections are raised by the end of the response period, the resolution will be considered to have consensus as a resolution of the Community Group."

My understanding is that we were using GitHub PRs for that, even without explicitly calling it out, and objections were raised by multiple people.

I would also note https://github.com/solid/specification/blob/main/meetings/2026-04-01.md#wac--acp

"SC: Maybe, give the group the opportunity to voice (dis-)comfort with this result and messaging?"

Certain discomfort seems to have been voiced.

Even without reading the CG charter. My common sense tells me that CG doesn't have consensus on this issue. We clearly don't agree and the whole ongoing conversation can be seen as evidence.

We already were planning to communicate Solid 26 as something that is still being finalized. This issue seems like one of the points that in the end we may note as something that we didn't reach consensus on.

Given ongoing Solid Week I would suggest that we only try to agree on how this work in progress can be communicated.

Best,
elf Pavlik
```

TallTed:
@jeswr — Please use the quote `>` marker (unfortunately, only available per-line) instead of code fences (even though the ` ``` ` are easier for big blocks) when quoting, email or otherwise. Infinite horizontal overflow is not easy to read, especially when needing to slide back to the start of the next line.

PR #789, opened 2026-05-07
post sosy CG 6 may 2026

thhck:


meetings/2026-05-06.md
```diff
@@ -0,0 +1,132 @@
+# W3C Solid Community Group: Weekly
+
+* Date: 2026-05-06T14:00:00Z at using the SolidOS stack, but I have heard no arguments as to what issues they think that raises.  No, it is not fully compliant with RDF-JS, but there are easy ways around the incompatibilities and many features available in the SolidOS stack not available in other approaches.
```
csarven: This comment "at using..." seems a bit out of place. It should probably move to where it is supposed to be, unless it is some comment from another meeting that was left over for some reason, and in which case it should be removed.

thhck: good catch, probably a copy past

elf-pavlik (approved):

PR #790, opened 2026-05-13
Create 2026-05-13.md

thhck:


meetings/2026-05-13.md
```diff
@@ -0,0 +1,129 @@
...
+- eP: you implemented WAC in java, I'll be curious how the WAC plays with encryption. specially with LDP containers.
+- RZ: ... How does ACL synchronize ? 
+- Niko: NG not implemented WAC for now. Did it for ActivityPod, but for now this would break e2ee so not implemented. If we start with private/public key for user and start with CRDT updates, good start, but this is costly because need crypt operation on each update. Later we can talk about optimization.
```
uvdsl, line 104: why does e2ee break authorization?? That does not sound right to me. e2ee is an additional layer to authorization in my mind...

Do you specifically refer to NextGraph? Could we clarify  that here before merging, please?

elf-pavlik: I wouldn't say it breaks but effective access is based on who can decrypt the information. Having access to encrypted blobs doesn't count IMO.

uvdsl: And yet, having access to encrypted blobs or not makes a difference from a security perspective.

I believe you have advocated for proper threat modelling in the past ;)

thhck: From my understanding it's not about WAC or ACP or else, it is that Solid is just not compatible with e2ee for now. The sentence should be understood as "if I must implement WAC in NG, then I would have to remove e2ee feature to make it work"
@uvdsl would that be better if I reformulate it that way ?

elf-pavlik: > And yet, having access to encrypted blobs or not makes a difference from a security perspective.

Preferably it shouldn't disclose anything, but let's not dive into on this thread

> I believe you have advocated for proper threat modelling in the past ;)

Yes! We need to start with a diagram https://www.w3.org/TR/threat-modeling-guide/

elf-pavlik (approved):

PR #798, opened 2026-07-01
first draft for an intent-handover spec

ylebre:
This PR is a first draft for an intent-handover specification. The aim is to allow initiating applications a flow to signal a 'login as' intent as well as an 'open this resource' intent to the receiving application.

### DIFF
--- a/2026/intent-handover.html
+++ b/2026/intent-handover.html
@@ -1,0 +1,85 @@
+<h1 id="intent-handover-for-solid-applications">Intent Handover for Solid Applications</h1>
+<h2 id="abstract">Abstract</h2>
+<p>This specification defines a mechanism that enables one Solid application to convey user intent to another Solid application. A receiving application MAY use any, all, or none of the conveyed information.</p>
+<h2 id="status-of-this-document">Status of This Document</h2>
+<p>This is a draft proposal intended for discussion within the Solid community. It does not yet define a standard and may change based on implementation experience and community feedback.</p>
+<h2 id="introduction">1. Introduction</h2>
+<p>Solid applications can redirect users to other Solid applications to continue a workflow. During such transitions, the initiating application may already possess information that could simplify the user’s experience, such as the user’s WebID, Identity Provider (IDP), storage location, or a resource that should be opened.</p>
+<p>This specification defines a standardized mechanism for communicating such information between Solid applications.</p>
+<p>The mechanism is intentionally non-prescriptive. The receiving application retains complete control over how, or whether, the conveyed information is used.</p>
+<h2 id="conformance">2. Conformance</h2>
+<p>The key words <strong>MUST</strong>, <strong>MUST NOT</strong>, <strong>SHOULD</strong>, <strong>SHOULD NOT</strong>, and <strong>MAY</strong> in this document are to be interpreted as described in RFC 2119 and RFC 8174 when, and only when, they appear in all capitals, as shown here.RFC 2119 and RFC 8174.</p>
+<h2 id="design-principles">3. Design Principles</h2>
+<p>The mechanism defined by this specification follows these principles:</p>
+<ul>
+<li>The initiating application MAY provide one or more pieces of contextual information.</li>
+<li>The receiving application MAY use the provided information.</li>
+<li>The receiving application MAY ignore any or all provided information.</li>
+<li>The receiving application remains responsible for validating any information before acting upon it.</li>
+<li>The mechanism does not require any particular authentication or authorization model.</li>
+</ul>
+<h2 id="intent-parameters">4. Intent Parameters</h2>
+<p>This specification defines intent parameters as HTTP URL query parameters.</p>
+<p>An initiating application MAY include any combination of the following parameters. When multiple intent parameters are present, the receiving application MAY use any subset of them. Query parameter values MUST be URL-encoded according to RFC 3986. Examples in this document omit URL encoding for readability.</p>
+<table>
+<thead>
+<tr class="header">
+<th>Parameter</th>
+<th>Value</th>
+</tr>
+</thead>
+<tbody>
+<tr class="odd">
+<td><code>solidIntent</code></td>
+<td>loginAs, openResource</td>
+</tr>
+<tr class="even">
+<td><code>solidIntentWebId</code></td>
+<td>WebID URI</td>
+</tr>
+<tr class="odd">
+<td><code>solidIntentIdp</code></td>
+<td>Identity Provider URI</td>
+</tr>
+<tr class="even">
+<td><code>solidIntentStorage</code></td>
+<td>Storage URI</td>
+</tr>
+<tr class="odd">
+<td><code>solidIntentResource</code></td>
+<td>Resource URI</td>
+</tr>
+</tbody>
+</table>
+<h3 id="intent">4.1 Intent</h3>
+<p>This parameter signals the intent of the user to the receiving app. Currently two possible values are specified.</p>
+<h3 id="loginas">4.1.1 loginAs</h3>
+<p>When this intent is signalled, the receiving application MAY initiate authentication using the supplied values from either <code>solidIntentWebId</code> or <code>solidIntentIdp</code></p>
+<h3 id="openresource">4.1.2 openResource</h3>
+<p>When this intent is signalled, the receiving application MAY use the provided values from <code>solidIntentStorage</code> and/or <code>solidIntentResource</code> to start the application with a specific resource. The initiating application can implement ‘open with…’ using this mechanic.</p>
+<p>Example:</p>
+<pre><code>https://target.app/?intent=loginAs&amp;solidIntentWebId=https://example.com/profile/card#me
+https://target.app/?intent[]=loginAs&amp;intent[]=openResource&amp;solidIntentWebId=https://example.com/profile/card#me&amp;solidIntentResource=https://storage.example.com/resource.ttl</code></pre>
+<h3 id="webid">4.2 WebID</h3>
+<p>A WebID identifying the user.</p>
+<p>Example:</p>
+<pre><code>https://target.app/?solidIntentWebId=https://example.com/profile/card#me</code></pre>
+<p>A receiving application MAY use this value to initiate authentication for the supplied WebID.</p>
+<h3 id="identity-provider-idp">4.3 Identity Provider (IDP)</h3>
+<p>The URI of an Identity Provider.</p>
+<p>Example:</p>
+<pre><code>https://target.app/?solidIntentIdp=https://idp.example.com/</code></pre>
+<p>A receiving application MAY use this URI as a suggested Identity Provider for authentication.</p>
+<h3 id="storage-uri">4.4 Storage URI</h3>
+<p>The URI of a Solid storage location.</p>
+<p>Example:</p>
+<pre><code>https://target.app/?solidIntentStorage=https://storage.example.com/</code></pre>
+<p>A receiving application MAY use this value as the preferred storage location. For users that have multiple storage locations in their profile information, this parameter conveys which storage location the initiating application intends the receiving application to use.</p>
+<h3 id="resource-uri">4.5 Resource URI</h3>
+<p>The URI of a specific resource.</p>
+<p>Example:</p>
+<pre><code>https://target.app/?solidIntentResource=https://storage.example.com/documents/report.ttl</code></pre>
+<p>A receiving application MAY attempt to open, inspect, or otherwise use the referenced resource as appropriate for the application’s functionality.</p>
+<h2 id="extensibility">5. Extensibility</h2>
+<p>Future versions of this specification MAY define additional intent parameters.</p>
+<p>Applications SHOULD ignore parameters they do not recognize unless another specification defines their semantics.</p>
--- a/2026/intent-handover.html
+++ b/2026/intent-handover.html
@@ -19,67 +19,91 @@
 <li>The mechanism does not require any particular authentication or authorization model.</li>
 </ul>
 <h2 id="intent-parameters">4. Intent Parameters</h2>
-<p>This specification defines intent parameters as HTTP URL query parameters.</p>
+<p>This specification defines intent parameters as either HTTP URL fragment parameters or HTTP URL query parameters. The usage depends on the capabilities of the receiving application.</p>
+<p>Receiving applications that run purely in the browser SHOULD use fragment parameters to prevent transmission of information to the server.</p>
+<p>Receiving applications that run server side and thus require the intent to be transmitted to the server, MUST use query parameters.</p>
 <p>An initiating application MAY include any combination of the following parameters. When multiple intent parameters are present, the receiving application MAY use any subset of them. Query parameter values MUST be URL-encoded according to RFC 3986. Examples in this document omit URL encoding for readability.</p>
 <table>
 <thead>
 <tr class="header">
 <th>Parameter</th>
 <th>Value</th>
+<th>Used by intent</th>
 </tr>
 </thead>
 <tbody>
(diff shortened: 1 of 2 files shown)


jeswr:
When this is read without the context of the demo given in the Solid CG, it can be hard to imagine the expected flow. I would recommend describing the flow of the demo - in addition to having the flow diagram described in the meeting, so that the intended behaviour can be understood.

uvdsl:
This PR was discussed on the [2026-07-01 meeting](https://github.com/solid/specification/pull/800), including a quick rundown of the demo.

elf-pavlik:
Two recommendations for sequence diagrams
* Mermaid as in https://github.com/solid/solid-oidc/blob/main/sequence.mmd - very simple and github can render it everywhere
* LikeC4 as in https://elf-pavlik.github.io/lws-auth/view/oidc/?dynamic=sequence
    This one can be navigated step by step and show rich Markdown notes, eventually Solid-OIDC primer could have an optional representation like that. I'll be working on reusable LWS model in LikeC4 to also reuse it for https://github.com/w3c/lws-protocol/issues/177

Potherca:
> in addition to having the flow diagram

Something like this?

Potherca:
Also, this discussion already talks about the scenario for opening resources: https://github.com/pod-os/PodOS/issues/47

Potherca:
Also also, as the intent is an instruction for the client-side code and not the server, I think that using a hash `#` would be more fitting than using query params (as those are meant as instructions for the server).

This would also prevent the WebID (or other values) from being visible server-side (in logs, etc.) which is better from a privacy perspective.

2026/intent-handover.html
```diff
@@ -0,0 +1,109 @@
...
+<ul>
+<li>The initiating application MAY provide one or more pieces of contextual information.</li>
+<li>The receiving application MAY use the provided information.</li>
+<li>The receiving application MAY ignore any or all provided information.</li>
+<li>The receiving application remains responsible for validating any information before acting upon it.</li>
+<li>The mechanism does not require any particular authentication or authorization model.</li>
+</ul>
+<h2 id="intent-parameters">4. Intent Parameters</h2>
+<p>This specification defines intent parameters as either HTTP URL fragment parameters or HTTP URL query parameters. The usage depends on the capabilities of the receiving application.</p>
+<p>Receiving applications that run purely in the browser SHOULD use fragment parameters to prevent transmission of information to the server.</p>
```
elf-pavlik, line 23: Does the initiating application construct url with parameters? If it does how would it decide to use query or fragment? The whole idea of having parameters in a fragment doesn't sound familiar to me. Do you have examples of it being used elsewhere?

elf-pavlik:
Thank you for the diagram @Potherca I think I understand the intent (no pun intended!).

My main issue is that in my idea of a default state of things, applications don't know anything about any other applications that the user is using - privacy concern. Even if they know they shouldn't be disclosing any information to other applications, specifically information which isn't publicly available via user's WebId.

I recall some prior conversations about 'open with' style interactions. The only link I can find now https://github.com/solid/catalog/issues/24

While we can't assume that any application used by a user knows anything about any other application used by that user. At least SAI and MANDAT have notion of Authorization App/Agent, which is special case app/service with hight trust from the user. In https://sai.js.org I actually combine it with the IdP itself, but they could be separate. Whatever flow we design should accommodate for possibility of user having such special app/service, in SAI it is advertised in WebID, similar to OIDC Issuer. And flow may need to go through that special party.

Can we pick 2-3 specific, real world use cases and walk through them 

1. without this proposal
2. with this proposal
  a) as of now
  b) extended to a special trusted app/party incorporated in this flow to avoid privacy leaks between applications

I'm happy to work on the same use cases with approach 2b. 

@jg10-mastodon-social @srosset81

jg10-mastodon-social:
Here's what I had been working on:
https://jg10.solidcommunity.net/open-with/databrowser-open-with.html

We indeed agreed that apps should not need to have knowledge of what other apps a user is using.

Given the state of things, I moved to working on a less normative guide, still a WIP:

https://jg10.solidcommunity.net/open-with/guide.html

We hadn't covered auto-login/webId specification.

On Solid Focus (which now supports auto-login), we had talked about [linking to other resources](https://github.com/NoelDeMartin/solid-focus/issues/16#issuecomment-2996057993) and [deep linking to a task](https://github.com/NoelDeMartin/solid-focus/issues/29)

2026/intent-handover.md
```diff
@@ -0,0 +1,163 @@
...
+
+* The initiating application MAY provide one or more pieces of contextual information.
+* The receiving application MAY use the provided information.
+* The receiving application MAY ignore any or all provided information.
+* The receiving application remains responsible for validating any information before acting upon it.
+* The mechanism does not require any particular authentication or authorization model.
+
+## 4. Intent Parameters
+
+This specification defines intent parameters as either HTTP URL fragment parameters or HTTP URL query parameters. The usage depends on the capabilities of the receiving application.
```
jg10-mastodon-social, line 35: question: How would an initiating application discover the capabilities of the receiving application?

suggestion: I would prefer to use only query parameters

2026/intent-handover.md
```diff
@@ -0,0 +1,163 @@
...
+Receiving applications that run purely in the browser SHOULD use fragment parameters to prevent transmission of information to the server.
+
+Receiving applications that run server side and thus require the intent to be transmitted to the server, MUST use query parameters.
+
+An initiating application MAY include any combination of the following parameters. When multiple intent parameters are present, the receiving application MAY use any subset of them. Query parameter values MUST be URL-encoded according to RFC 3986. Examples in this document omit URL encoding for readability.
+
+
+| Parameter              | Value                          | Used by intent |
+| ---------------------- | ------------------------------ | -------------- |
+| `intent`               | login, open                    |                |
```
jg10-mastodon-social, line 46: suggestion: open would presumably imply login, and the presence of a webId or resource would presumably imply login and/or open with respectively. The intent parameter could therefore be removed/optional unless there are specific future plans for it?

2026/intent-handover.md
```diff
@@ -0,0 +1,163 @@
...
+| `solidWebId`           | A Solid compatible WebID URI   | login          |
+| `solidIdp`             | A Solid Identity Provider URI  | login          |
+| `solidStorage`         | A Solid Storage URI            | open           |
```
jg10-mastodon-social, line 49: suggestion: this would presumably be a special case of solidResource, and if needed the fact that it is a storage could be confirmed using https://solidproject.org/TR/protocol#storage-resource

2026/intent-handover.html
Potherca: Originally, the intent was to only support the fragment `#`, as that avoid more complexity (like capability discovery). 

Concerns were raised that this would not work with server-side apps, which I disagree with.

From my perspective, intent **only** makes sense for a UI. The UI for a server-side app could easily convert the `#` to a `?` and do a URL reload to trigger the server, without the need for supporting both fragment and query parameters.

Regarding

> Do you have examples of it being used elsewhere?

Besides generic patterns of single-page applications (SPA) for things like `#q=foo&type=bar&filter=baz`, there are various apps and protocol that use fragments with the explicit goal of avoiding parameters from reaching the server. Either because they are secret or because the server would not know what to do with those params.

<details><summary>For instance... (click to toggle details)</summary>

- Excalidraw uses fragment parameters for share URLs:  `https://excalidraw.com/#json=SvgTAQvEpfAAwYdFqFhy1,kT7-izZXZFSgAIxQ9TctGw`

- Kibana (Elasticsearch dashboard) uses fragment parameters for [app state](https://www.elastic.co/docs/extend/kibana/key-concepts/platform-architecture/routing-navigation-and-url#specifying-state):  `https://your-kibana-domain:5601/app/discover#/?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_a=(columns:!(message),filters:!())`

- M$ Teams uses fragment parameters for streaming initialization context:  `https://teams.microsoft.com/_#/?ctx=chat&tenantId=xxxx-xxxx-xxxx`

- OAuth 2.0 Implicit Grant Flow (now deprecated) uses fragments for it's [Access Token Response](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2) [RFC 7111](https://datatracker.ietf.org/doc/html/rfc7111)

- [RFC-5147](https://datatracker.ietf.org/doc/html/rfc5147) uses fragments to refer to parts of a text file by character or line position (or range):  `http://example.com/text.txt#line=10,20`

- [RFC-7111](https://datatracker.ietf.org/doc/html/rfc7111) does the same but with by row, column, or cell, for CSV files.

- vdo.ninja  (previously bdo.ninja) uses fragment parameters [to hide Stream IDs and Room Names](https://docs.vdo.ninja/help/privacy-and-security-details/fragment-url-parameters#hiding-stream-ids-and-room-names) from the server:  `https://vdo.ninja/?push=cam1#p=secret&token=abc123&key=xyz`

- YouTube video player uses fragment parameters for player instructions:  `https://www.youtube.com/watch?v=dQw4w9WgXcQ#t=90&autoplay=1&cc_load_policy=1`

- etc.

</details>

2026/intent-handover.md
Potherca: I would prefer using only fragment parameters.

As the intent is an instruction for the client-side code and not the server, I think that using a hash `#` would be more fitting than using query params (as those are meant as instructions for the server).

This would also prevent the WebID (or other values) from being visible server-side (in logs, etc.) which I think is better from a privacy perspective.

jg10-mastodon-social: The reason I would prefer to use query parameters is that its a standard way of passing values for parameters, whereas fragments identify a specific part of the resource.

How would you handle opening a url that already has a fragment?

uvdsl: > As the intent is an instruction for the client-side code and not the server,

I do see the argument of an SPA not accidentally leaking information on the host server logs.
If however the client has a BFF, I think the information are transmitted to the server-side component anyway for authentication?   (but that is on purpose then, so probably fine anyway)
And if the app is fully server rendered, I am unsure whether this would work using only fragments?

That asked, there is indeed prior art to having parameters in the fragment, see W3C REC of [Media Fragments URI 1.0 (basic)](https://www.w3.org/TR/media-frags/#media-fragment-syntax). Maybe worth a look regarding their approach and reasoning.

csarven (commented):
Pardon me as I couldn't review this earlier.

Signalling who to log in as or what to open when handing a user over to another application is a real need. I want to support them, with a suggestion and a few concerns.

Suggestion: consider converging with Application Capability, proposed as a work item in #806, draft: https://dokieli.github.io/application-capability/ . It covers this draft's use cases: `login` and `open` are defined [invocation variables](https://dokieli.github.io/application-capability/#invocation-variables), carried in a URI Template that the receiving application publishes in a discoverable description. Different kinds of capabilities and invocations can be created (beyond what's in the spec, see the [Invocation Extension](https://dokieli.github.io/application-capability/#invocation-extension) section). Discovery also fixes a gap here: an initiating application cannot know which parameters or which form (query or fragment) a receiver supports, so pairings end up hardcoded. If you have needs it does not yet cover, I'd rather we capture them there together than maintain two overlapping mechanisms. As explained elsewhere Application Capability goes much further than URI Templates for intending/invoking/triggering/hinting... and we need to look at this as part of a bigger picture.

In RFC 3986, the query is part of what identifies a resource, and per RFC 9110 it is sent to the server. So `?intent=login&solidWebId=...` discloses the user's identity and target resources to the receiving host, its logs, and intermediaries. The MUST for server-side applications normalises this, and there are no security or privacy considerations discussing it. See [URI Query](https://dokieli.github.io/application-capability/#uri-query), [URI Fragment](https://dokieli.github.io/application-capability/#uri-fragment), and the [Threat Model](https://dokieli.github.io/application-capability/#threat-model) for how AC approaches this.

Regarding Solid-specific naming, the `solid*` prefixes bake one ecosystem into the wire format, while section 5 itself says the mechanism is not unique to Solid. These are all just IRIs; generic names travel further.

`solidIdp` and `solidStorage` are discoverable from the WebID profile (`solid:oidcIssuer`, `pim:storage`). Passing them duplicates authoritative data and enlarges what is disclosed per invocation. The agent's URI is enough; the rest follows by discovery.

None of this is a "no" to the use cases. There is implementation experience covering them on the AC side, and I would welcome working on this together.

2026/intent-handover.md
```diff
@@ -0,0 +1,163 @@
...
+
+The mechanism defined by this specification follows these principles:
+
+* The initiating application MAY provide one or more pieces of contextual information.
+* The receiving application MAY use the provided information.
+* The receiving application MAY ignore any or all provided information.
+* The receiving application remains responsible for validating any information before acting upon it.
+* The mechanism does not require any particular authentication or authorization model.
+
+## 4. Intent Parameters
```
uvdsl, line 33: Do you see intent parameters in this section that are not covered in https://dokieli.github.io/application-capability/#invocation-variables? I believe them to be covered?

Do you see https://dokieli.github.io/application-capability/#uri-template-invocation as an equivalent way to construct the URI with the intent parameters? Or do you see a blocker there?

uvdsl (commented):
I am wondering whether you think that https://dokieli.github.io/application-capability/ might be a suitable generalization of your proposal?

Potherca:
We'll have to wait for his response, but I think making this a lightweight (login only? spec is closer to @ylebre's original idea.

@uvdsl I was thinking in the same direction. Being compatible with the proposed Application Capability spec seems A Good Idea :tm:

2026/intent-handover.md
uvdsl: @ylebre, a friendly ping :slightly_smiling_face:

2026/intent-handover.html
ylebre: The initiating application would need prior knowledge of the target application to know which format the parameters should go in. I have not really thought about discoverability - it would make sense for target applications to somehow advertise that they support this and what format they require.
A simple example popped up with the implementation that was done in Media Kraken, which uses the ?loginWith parameter. I added the query parameters option because I don't really want to tell other developers how they should implement things on their side.

2026/intent-handover.md
ylebre: Fair point - I tend to try writing things in such a way that they are explicit and do not rely on implied instructions.
When looking at the other parameters, one could imply that if I am passing a solidWebId and a solidResource, me intent would be to login and open the resource. I just prefer to be explicit about this, hence the extra parameter.

2026/intent-handover.md
ylebre: I think the preferred way of passing information should be the fragment, as that does not send the information to the host server logs. On the other hand, if we mandate fragment as the 'only way' for this, we block a bunch of implementation options. This is why both options are supported in the proposal.

2026/intent-handover.md
ylebre: I added the solidStorage for a file-browser use case. This allows the file browser app to know which storage I would like as the root folder, and what resource I would like it to start with in the file browser. I think adding this variant allows the initiating app to convey the intent more clearly.

2026/intent-handover.md
ylebre: I think the application capability covers all the use cases I have in the proposal, as well as discovery and a couple of others. I would have no issue converging on the application capability instead of this proposal - all I am after is that we have some way to cover the use cases I have in mind in a way that is easy for implementers to support.

I am trying to keep this proposal light weight and open for extension so that if other 'intents' came up, those could be worked on as extensions to this proposal in their own right rather than making them part of this proposal. My hope is that it will be easier to find alignment with other implementers with a smaller proposal.

